PR #8, opened 2020-08-26
Phase 1 basic API

parc-jason:
# Description
Add in API to support form, form submission, and RBAC.
No data seed scripts.
No security on endpoints (except /rbac/current), will deal with security later when API is complete.
Using Views to build out the relations between a user and a form and vice versa.

## Types of changes

## Checklist


app/src/forms/form/service.js
```diff
@@ -0,0 +1,161 @@
...
+  createSubmission: async (formVersionId, data) => {
+    let trx;
+    try {
+      const formVersion = await service.readVersion(formVersionId);
+      trx = await transaction.start(FormSubmission.knex());
+
+      const obj = Object.assign({}, data);
+      obj.id = uuidv4();
+      obj.formVersionId = formVersion.id;
+      obj.confirmationId = obj.id.substring(0, 8).toUpperCase();
```
jujaga, line 123: Wondering if we should still be retaining a submission/confirmation id of just 8 characters or if we should be using the whole GUID instead. The confirmationid shorthand is slightly less useful if we don't have a way to reverse lookup and find your original submission otherwise.

app/src/forms/rbac/service.js
```diff
@@ -0,0 +1,440 @@
...
+    } else {
+      // what if name or email changed?
+      user = await service.updateUser(user.id,{
+        keycloakId: access_token.content.sub.toString(),
+        fullName: access_token.content.name.toString(),
+        username: access_token.content.preferred_username.toString(),
+        firstName: access_token.content.given_name.toString(),
+        lastName: access_token.content.family_name.toString(),
+        email: access_token.content.email.toString()
+      });
```
jujaga, line 277: Since the passed in object for keycloakId, fullName... email is used in both branches of the conditional, that should be specified as a const outside of it to remove redundancy.
```js
const userData = { keycloakId: ... };
if (!user) {
    user = await service.createUser(userData);
} else {
    user = await service.updateUser(user.id, userData);
}
```

app/src/forms/rbac/service.js
```diff
@@ -0,0 +1,440 @@
...
+      keycloakId: item.keycloakId,
+      username: item.username,
+      fullName: item.fullName,
+      email: item.email,
+      firstName: item.firstName,
+      lastName: item.lastName,
+      roles: [],
+      permissions: []
+    };
+  }
```
jujaga, line 102: These helper transform functions appear duplicated at the top of the file as well. As these are generally more "schema" transform functions, it may not be a bad idea to create a separate transform.js supplementary file, or even just a top level const transformer that can be module.exported where its sole purpose is to just do an object to object mapping. Also makes it easier to unit test this as a sole unit later instead of having it be defined and embedded down a level, where we'd only be able to test it implicitly.

app/src/forms/rbac/service.js
```diff
@@ -0,0 +1,440 @@
...
+      .modify('filterActive', params.active)
+      .modify('filterArray', 'roles', params.permissions)
+      .modify('orderDefault');
+
+    return toUserForms(permissions, roles);
+  },
+
+  setFormUsers: async (formId, userId, data) => {
+    // check this in middleware? 422 in valid params
+    if (!formId || 0 === formId.length) {
```
jujaga, line 364: Style: Normally in JS you check for equality by having the thing you are checking for be the first statement, and then the constant or condition to match as the second statement. Obviously not a hard/fast style guide and it's still obvious what's being done, but nearly all examples have it in reverse. (https://github.com/airbnb/javascript#control-statements)

app/src/forms/rbac/service.js
parc-jason: done

app/src/forms/form/service.js
parc-jason: yes, confirmation id is still valuable.  rendered on screen, it's easier to digest, plus we format it and save it so the the UI doesn't have to do any mangling.

PR #11, opened 2020-08-31
Add Identity Providers

parc-jason:
Enhance the user to form access querying.
No public flag, use Identity Providers for access.  Users can now create a form that is "private" but anyone with an IDIR will have access without being granted explicit roles.
Remove the need for additional server-side filtering on roles and permissions.

Forms no longer have a public flag, but the will have a new relation named identityProviders.  These will need to be set on create and update.  Listing for allowed IDPs is GET /rbac/idps...

We have removed the draft flag from versions, that is added to the backlog if we want to re-introduce the concept later.

For user forms queries (ex, /rbac/current, /rbac/forms, rbac/users), we can now filter by "idps".
Keep in mind, /current will pass in the current user's idp as found in their token, so we really should only pass in ?idps=public (their idp will be added).

**Important**: Will need to rollback all previous migrations or start with a clean database.

# Description

## Checklist


app/src/forms/common/models/models.js
```diff
@@ -335,16 +341,44 @@ class UserFormRoles extends Model {
...
+
+          if (_permissions.length) {
+            clauses.push(`(array_length("permissions", 1) > 0 and (${inArrayClause('permissions', _permissions)}))`);
+          }
+
+          if (clauses.length) {
+            query.whereRaw(`(${clauses.join(' or ')})`);
+          }
+        }
+      },
```
jujaga, line 381: Do we need to consider some kind of return value when neither idps, roles nor permissions are specified? I think the return value of this function would be undefined as it is atm.

app/src/forms/form/service.js
```diff
@@ -17,10 +18,16 @@ const service = {
     try {
       trx = await transaction.start(Form.knex());
 
-      const obj = Object.assign({}, data);
+      const obj = {};
       obj.id = uuidv4();
+      obj.name = data.name;
+      obj.description = data.description;
+      obj.shortName = data.shortName ? data.shortName : obj.id.substring(0, 8).toUpperCase();
```
jujaga, line 25: Will need to discuss with team - but should decide if we still need a "shortname GUID" at all or not (depends on how we're thinking about handling form/submission version identification either by "autogen names, or something else".

In comfort, we did have short GUID confirmations, but there was never a direct mechanism to do lookup based on only that value - it was always only a derived value from the id key.

app/src/forms/form/service.js
```diff
@@ -36,7 +43,12 @@ const service = {
       const obj = await service.readForm(formId);
       trx = await transaction.start(Form.knex());
 
-      await Form.query(trx).patchAndFetchById(formId, {name: data.name, description: data.description, public: data.public, active: data.active, labels: data.labels});
+      await Form.query(trx).patchAndFetchById(formId, {name: data.name, description: data.description, shortName: data.shortName, active: data.active, labels: data.labels});
```
jujaga, line 46: Style: some of these statements are long enough we should consider them being multi-line instead.

app/src/forms/common/models/models.js
parc-jason: There is no return in these methods.  They build out the database query. If there are no criteria, then we don't alter the query (add where clause etc).

app/src/forms/form/service.js
parc-jason: Still in rough, will wait until API is settled and we start adding validation and security.  Then we will clean up the code.

app/src/forms/form/service.js
parc-jason: I don't think it's quite as complicated as you are making it out to be.  But I am removing shortName for now and adding a backlog ticket so we don't waste time discussing this right now.

PR #12, opened 2020-09-02
Add minimally functional form editor

jujaga:
# Description

This PR adds formio editor component support. FormDesigner component and Root route takes in formId and formVersionId as props. This has been mounted in the vue router as a parameterized child. At this time, the component can only save to an existing version (U in CRUD).

## Types of changes

## Checklist

## Further comments

Assuming you are using the temporary seed script... route has following path: `/editor/:formId/versions/:formVersionId`
"Start a new form": http://localhost:8081/app/editor (no path params, submit button does not work in this case)
"Update an existing form": http://localhost:8081/app/editor/b1cff682-1412-442a-ac6a-cd70baa35022/versions/544cb861-a435-4f34-8994-49a0723de0b7 (has path params for both formId and formVersionId).

TODO: figure out the C in CRUD flow.

### DIFF
--- a/app/frontend/public/index.html
+++ b/app/frontend/public/index.html
@@ -1,20 +1,28 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
-    <meta charset="utf-8">
-    <meta http-equiv="X-UA-Compatible" content="IE=edge">
-    <meta name="viewport" content="width=device-width,initial-scale=1.0">
-    <link rel="icon" href="<%= BASE_URL %>favicon.ico">
-    <title><%= process.env.VUE_APP_TITLE %></title>
-    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:100,300,400,500,700,900">
-    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Material+Icons">
-    <link rel="stylesheet" href="https://unpkg.com/@mdi/font@latest/css/materialdesignicons.min.css">
-  </head>
-  <body>
-    <noscript>
-      <strong>We're sorry but <%= process.env.VUE_APP_TITLE %> doesn't work properly without JavaScript enabled. Please enable it to continue.</strong>
-    </noscript>
-    <div id="app"></div>
-    <!-- built files will be auto injected -->
-  </body>
+
+<head>
+  <meta charset="utf-8">
+  <meta http-equiv="X-UA-Compatible" content="IE=edge">
+  <meta name="viewport" content="width=device-width,initial-scale=1.0">
+  <link rel="icon" href="<%= BASE_URL %>favicon.ico">
+  <title><%= process.env.VUE_APP_TITLE %></title>
+  <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:100,300,400,500,700,900">
+  <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Material+Icons">
+  <link rel="stylesheet" href="https://unpkg.com/@mdi/font@latest/css/materialdesignicons.min.css">
+  <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" />
+  <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.1.3/css/bootstrap.min.css" />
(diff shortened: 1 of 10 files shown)


app/frontend/public/index.html
```diff
@@ -1,20 +1,28 @@
...
+  <meta http-equiv="X-UA-Compatible" content="IE=edge">
+  <meta name="viewport" content="width=device-width,initial-scale=1.0">
+  <link rel="icon" href="<%= BASE_URL %>favicon.ico">
+  <title><%= process.env.VUE_APP_TITLE %></title>
+  <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:100,300,400,500,700,900">
+  <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Material+Icons">
+  <link rel="stylesheet" href="https://unpkg.com/@mdi/font@latest/css/materialdesignicons.min.css">
+  <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" />
+  <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.1.3/css/bootstrap.min.css" />
+  <link rel="stylesheet" href="https://unpkg.com/formiojs@latest/dist/formio.full.min.css" />
```
parc-jason, line 15: do we want latest?  isn't there a potential to screw us up if they modify and breaks something?

app/frontend/src/components/designer/FormDesigner.vue
```diff
@@ -0,0 +1,105 @@
...
+  }),
+  created() {
+    this.getFormSchema();
+  },
+  methods: {
+    async getFormSchema() {
+      this.formSchema = {};
+      try {
+        if (this.formId) {
+          const form = await formService.readForm(this.formId);
```
parc-jason, line 54: you could just read the form, and look in its versions[] for the version you want.

app/frontend/src/components/designer/FormDesigner.vue
```diff
@@ -0,0 +1,105 @@
...
+          if (this.formVersionId) {
+            const response = await formService.readVersion(this.formId, this.formVersionId);
+            this.formSchema = response.data.schema;
+          }
+        }
+      } catch (error) {
+        console.error(`Error loading form schema: ${error}`); // eslint-disable-line no-console
+      }
+    },
+    async submitFormSchema() {
```
parc-jason, line 67: Just need to be very clear that everyone realizes at this time, the form must exist and have an existing version for this component to work.  This only updates the specified version.

parc-jason (commented):
Going to let @loneil decide to pull this in.
Just need to be aware what this PR can do and what the limitations are...

app/frontend/public/index.html
loneil: Try it out for now? But yeah on our stable release we'd want a static one. Might not even want to use a CDN anyways so our system isn't dependent on the hosted files...

app/frontend/src/components/designer/FormDesigner.vue
loneil: Yeah lets address Create New Form after

loneil:
Lets pull it in, will look at the concepts around versioning after integrating a few things how about?

app/frontend/public/index.html
jujaga: Yea - we can always lock it down once we get a better handle on things. I suspect once we get into custom component builds, we may need to internally build the formio distribution anyways so this may potentially disappear.

app/frontend/src/components/designer/FormDesigner.vue
jujaga: This component will be able to handle both scenarios, but yea - right now only Update (not Create) has been implemented. Will need to figure out the create flow pretty soon.

app/frontend/src/components/designer/FormDesigner.vue
jujaga: I did see that in the form level object response. The reason I didn't put in any custom logic to parse/list available versions is because I was assuming there would either be a management page or some other entrypoint to this component that would answer the question for it already. The current incarnation of this component is relatively idempotent - if it knows the `formId` and `formVersionId`. Not sure on how we handle the "create a totally new form" just yet - for later discussion.

loneil (approved):

PR #15, opened 2020-09-03
Feature/current user enhancements

parc-jason:
# Description
Add current user middleware to routes.  This will work for public users too.
Add currentUser to request, and all form access.
Will leverage the user/forms/roles/permissions on subsequent API calls to throw 401 if not allowed to perform an action.
Update the access views.  Give submitter permissions for public and idp only access (for permissions checks), but do not give any roles.
Add filters to /rbac/current to limit what form access to return (public, idp, team).
public=true means you only have the form is public, and you have no explicit role assignments
idp=true means you have access because the form allows users with your idp (from jwt token) and you have no explicit role assignments.
team=true means you have explicitly been given a role on the form.

These filters do not apply to a public user, they will always return public accessible forms.

/rbac/current will return a single user object with a forms array property.

## Types of changes

## Checklist


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-wxpbtr-tools.pathfinder.gov.bc.ca/dashboard?id=common-hosted-form-service-pr-15

loneil (approved):

PR #19, opened 2020-09-11
Create custom formio component.

parc-jason:
Add a "project" for creating custom formio components for bcgov. (under /components, will move to own repo later)
Add Org Book lookup/select component.
Add the component to frontend.

Org. Book component will be available under the Advanced listing of components.
Defaults:
The label is Registered Business Name.
The submission field name is registeredBusinessName.
The value returned in only the string; it is not the object from Org Book API.
The field is NOT required.

A weirdness/glitch is once an item is selected, opening the drop-down list shows two of the same item.  I cannot figure it out, and it appears that this is an issue in the underlying, underlying choices.js library.  Leaving for now so we can move on, the task has been demonstrated on how to create a custom component and if we can create one for Org. Book.  We can refine later (most likely by building from scratch not inheriting the formio select box).  

# Description

## Types of changes

## Checklist


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-wxpbtr-tools.pathfinder.gov.bc.ca/dashboard?id=common-hosted-form-service-pr-19

TimCsaky (approved):
great work!

jujaga (approved):
Looks like the thing works when drag and dropped in ui. Some JS files have strange indentation (4 spaces or tabs) - presuming this is due to TS auto-generated code. Looks good otherwise. 👍

loneil (approved):

PR #27, opened 2020-09-21
Bugfix comp is null bug when using top-level Layout options

jujaga:
# Description

Form.io was not happy in the onChangeMethod when we were storing and managing the formSchema as a Vue data object. This PR resolves this by making formSchema be a plain old javascript variable instead.

## Types of changes

## Checklist

### DIFF
--- a/app/frontend/src/components/designer/FormDesigner.vue
+++ b/app/frontend/src/components/designer/FormDesigner.vue
@@ -74,7 +74,7 @@
       </v-card-text>
     </v-card>
 
-    <FormBuilder :form="formSchema" @change="onChangeMethod" :options="{}" />
+    <FormBuilder :form="() => formSchema" @change="onChangeMethod" :options="{}" />
   </div>
 </template>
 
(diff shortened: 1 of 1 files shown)


app/frontend/src/components/designer/FormDesigner.vue
```diff
@@ -74,7 +74,7 @@
       </v-card-text>
     </v-card>
 
-    <FormBuilder :form="formSchema" @change="onChangeMethod" :options="{}" />
+    <FormBuilder :form="() => formSchema" @change="onChangeMethod" :options="{}" />
```
jujaga, line 77: Since formSchema is now only one-way bound, it needs to be presented as a functional getter; otherwise Vue will warn about it not being a reactive element.

bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-wxpbtr-tools.pathfinder.gov.bc.ca/dashboard?id=common-hosted-form-service-pr-27

loneil (approved):
Does the trick, tested it out

PR #32, opened 2020-09-25
2 Step Create form

loneil:
# Description
This is an early cut of this but some of the more complex setup with state and saving I was trying was causing headaches, and we keep changing the viewer and designer components and events.
Need to get UX feedback ASAP so going with this design for now with room for some improvement coming to make the code more extensible. So the designer and viewer components aren't state based for now. Things seem to be working fine anyways but can envision upcoming features where it's going to be more necessary.

Leaving in draft as Matt is demoing today and don't want to get this in until he's done.

## Types of changes

## Checklist


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-wxpbtr-tools.pathfinder.gov.bc.ca/dashboard?id=common-hosted-form-service-pr-32

app/frontend/src/components/designer/FormDesigner.vue
```diff
@@ -96,19 +140,89 @@ export default {
     ID_PROVIDERS() {
       return IdentityProviders;
     },
+    designerOptions() {
```
loneil, line 143: Going to move this to its own static file but there are multiple of us working on it so not yet.

app/frontend/src/store/index.js
```diff
@@ -1,10 +1,12 @@
 import Vue from 'vue';
 import Vuex from 'vuex';
 
+import form from '@/store/modules/form.js';
+
 Vue.use(Vuex);
 
 export default new Vuex.Store({
-  modules: {},
+  modules: { form },
```
jujaga, line 9: I'm still slightly wary about having this always be mounted - it's quite possible for a user to just visit the about page or myForms page and end up doing nothing else -> in this situation they'd never need the form module tree ever mounted. Honestly can go both-ways, but anytime we could leverage module tree-shaking I think we should be taking advantage of it.

app/frontend/src/store/modules/form.js
```diff
@@ -0,0 +1,104 @@
...
+    },
+    formList: [],
+    formSubmission: {
+      confirmationId: '',
+      submission: {
+        data: {}
+      }
+    },
+    submissionList: [],
+    version: {}
```
jujaga, line 20: Right now this state module focuses on only one form. Do we have any need down the line to be able to store and "remember" multiple different forms within one session? If we do, we should perhaps consider mounting the top level state object as a guid instead; that way we could have multiple form trees stored in stead instead of just one (thinking ahead to potential vuex-persist stuff)

app/frontend/src/store/modules/form.js
```diff
@@ -0,0 +1,104 @@
...
+      state.formSubmission = submission;
+    },
+    SET_SUBMISSIONLIST(state, submissions) {
+      state.submissionList = submissions;
+    },
+    SET_VERSION(state, version) {
+      state.version = version;
+    },
+  },
+  actions: {
```
jujaga, line 52: As an extension to my previous comment - one of the thoughts I had for potential offline support would be to make all of the state actions be pushed into a memory-managed queue - so instead of directly just fetching the form and acting on state immediately, the action would first enqueue the action, and then the queue itself would then manage the state and update as needed. This would afford us the ability to have more fail-safe behavior in the event network traffic isn't going through (trying to submit in terrible network environment or poor cell service for example).
Probably no action at this time, but it is something we should consider if it's on the roadmap.

app/frontend/src/components/designer/FormDesigner.vue
loneil: Yeah will discuss with Mina when she has a look at Basic mode. 👍

app/frontend/src/store/index.js
loneil: Discussed

app/frontend/src/components/forms/SubmissionsTable.vue
```diff
@@ -68,46 +65,40 @@ export default {
...
       } finally {
         this.loading = false;
       }
-    },
-    showTableAlert(typ, msg) {
-      this.alertShow = true;
-      this.alertType = typ;
-      this.alertMessage = msg;
-      this.loading = false;
-    },
```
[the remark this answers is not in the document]
loneil: Fixed

app/frontend/src/store/modules/form.js
loneil: 👍

app/frontend/src/store/modules/form.js
loneil: Trying to keep the store as mostly just the central location for all form managing actions that a page or pages would need.
So lists of forms for the list pages, the form for the page you're on for a specific form. It can manage for now, but yeah if sections of the app need to pertain to one form it could be changed.

bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-wxpbtr-tools.pathfinder.gov.bc.ca/dashboard?id=common-hosted-form-service-pr-32

jujaga (approved):

PR #34, opened 2020-09-25
Submission Access

parc-jason:
Make direct submission routes. GET /submissions/:id
Do more fine-grained permissions checks (user can have permissions on the submission itself).
New table and view for user permissions on a submission, allows us to have many users have access to the submission (ex. creator and a staff reviewer can have read/write access).

**NOTE**: the API result for read is an object that has {form, version, submission}, this allows the caller to create a complete view of the submission without calling for form metadata or version/schema data separately.

See [SHOWCASE-1554](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-1554)

# Description

## Types of changes

## Checklist

## Further comments
Once this is in place and being used for reading submission, we should remove the other routes/services for read and update and force all reads/writes through this service.

### DIFF
--- a/app/src/forms/common/models/utils.js
+++ b/app/src/forms/common/models/utils.js
@@ -1,0 +1,20 @@
+const toArray = (values) => {
+  if (values) {
+    return Array.isArray(values) ? values.filter(p => p && p.trim().length > 0) : [values].filter(p => p && p.trim().length > 0);
+  }
+  return [];
+};
+const inArrayClause = (column, values) => {
+  return values.map(p => `'${p}' = ANY("${column}")`).join(' or ');
+};
+
+const inArrayFilter = (column, values) => {
+  const clause = inArrayClause(column, values);
+  return `(array_length("${column}", 1) > 0 and (${clause}))`;
+};
+
+module.exports = {
+  toArray: toArray,
+  inArrayClause: inArrayClause,
+  inArrayFilter: inArrayFilter
+};
--- a/app/src/forms/submission/service.js
+++ b/app/src/forms/submission/service.js
@@ -1,0 +1,100 @@
+const { Form, FormVersion, FormSubmission, FormSubmissionUserPermissions, SubmissionMetadata, UserFormAccess } = require('../common/models');
+
+const Permissions = require('../common/constants').Permissions;
+
+const Problem = require('api-problem');
+const {transaction} = require('objection');
+
+const service = {
+
+  _fetchSubmissionData: async(formSubmissionId) => {
+    const meta = await SubmissionMetadata.query()
+      .where('submissionId', formSubmissionId)
+      .first()
+      .throwIfNotFound();
+
+    const submission = await FormSubmission.query()
+      .findById(meta.submissionId)
+      .throwIfNotFound();
+    const version = await FormVersion.query()
+      .findById(meta.formVersionId)
+      .throwIfNotFound();
+    const form = await Form.query()
+      .findById(meta.formId)
+      .allowGraph('identityProviders')
+      .withGraphFetched('identityProviders(orderDefault)')
+      .throwIfNotFound();
+
+    return {
+      form: form,
+      version: version,
+      submission: submission
+    };
+  },
+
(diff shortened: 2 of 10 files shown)


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-wxpbtr-tools.pathfinder.gov.bc.ca/dashboard?id=common-hosted-form-service-pr-34

app/src/forms/common/models/utils.js
```diff
@@ -0,0 +1,20 @@
+const toArray = (values) => {
+  if (values) {
+    return Array.isArray(values) ? values.filter(p => p && p.trim().length > 0) : [values].filter(p => p && p.trim().length > 0);
```
jujaga, line 3: `p.trim().length > 0` can be changed to just evaluating `p.trim().length` as they are equivalent.

jujaga (commented):
Looks pretty good. Only high-level thought I have is whether we'd also need similar logic for querying and accessing a specific form schema as you won't be able to render a submission without its associated schema version.

parc-jason:
> Looks pretty good. Only high-level thought I have is whether we'd also need similar logic for querying and accessing a specific form schema as you won't be able to render a submission without its associated schema version.

It returns the form, the schema and the submission in the same call.

app/src/forms/submission/service.js
```diff
@@ -0,0 +1,100 @@
...
+        .modify('filterByPermissions', permissions)
+        .first();
+    };
+
+    const isDeleted = result.submission.deleted;
+    const isDraft = result.submission.draft;
+    const publicAllowed = result.form.identityProviders.find(p => p.code === 'public');
+    const idpAllowed = result.form.identityProviders.find(p => p.code === currentUser.idp);
+    const formSubmissionsPermission = await checkFormSubmissionsPermission();
+    const submissionPermission = await checkSubmissionPermission();
```
loneil, line 61: Depending on speed of this, would the second lookup potentially be avoidable based on the formSubmissionsPermission?

app/src/forms/submission/service.js
```diff
@@ -0,0 +1,100 @@
...
+      .findById(meta.submissionId)
+      .throwIfNotFound();
+    const version = await FormVersion.query()
+      .findById(meta.formVersionId)
+      .throwIfNotFound();
+    const form = await Form.query()
+      .findById(meta.formId)
+      .allowGraph('identityProviders')
+      .withGraphFetched('identityProviders(orderDefault)')
+      .throwIfNotFound();
```
jujaga, line 26: Looking at this, may be possible to have these three queries done in a `Promise.all(...)` wrapper so that we can parallelize them. Realistically, these three queries "should" be pretty fast already, but there isn't anything preventing them from being completely asynchronous until the return statement (or if it throws)

jujaga (commented):
Ah ok - missed that on first read-through; I see it now. 👍 

My only remaining recommendation is to add/update the OpenAPI 3.0 spec with this new endpoint, and to also double check to make sure all of the references in there point to the right thing. Right now ReDocs appears to have some wierd behavior on some endpoints when clicked on the sidebar, and then jumps somewhere else unintended (I suspect it may be a copy paste error somewhere).

app/src/forms/submission/service.js
parc-jason: logically, there would be a path to optimize this, but until we need to do so, i'd rather leave it.  it will be a lot easier to debug with a straightforward logic path.  if speed is an issue, we can shuffle this around.

parc-jason: changed it up, calls when next permission set needed.

app/src/forms/submission/service.js
parc-jason: putting in promise.all(), will create another ticket to clean up the open api spec, no need holding this pr up.

bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-wxpbtr-tools.pathfinder.gov.bc.ca/dashboard?id=common-hosted-form-service-pr-34

loneil (approved):

jujaga (approved):

PR #35, opened 2020-09-25
Feature/qr code

loneil:
# Description
Go to the Manage > Share popup and you can generate a QR code.

## Types of changes

## Checklist


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-wxpbtr-tools.pathfinder.gov.bc.ca/dashboard?id=common-hosted-form-service-pr-35

jujaga (commented):
In terms of UI flow, I think it'd be best if the ShareForm modal just rendered the QR code all the time by default in Small format. We don't necessarily need to give the user a sliding bar for choosing how large the QR code is either as it's going to be an SVG. Saves a button press, and makes the feature much more obvious.

loneil:
Leaving as canvas with downloadable png, can copy, can right-click download as well.
If there's a need the library can render as svg too. User could choose.

jujaga (approved):

PR #44, opened 2020-10-08
Apply bcgov style to the Form Builder and the forms it builds

TimCsaky:
# Description

supports [work on styling the form ](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-1608)

We need the form builder and the forms it creates to have the look and feel of the bcgov style guidelines.
long term approach could be:
- modify the html markeup produced by the formio form builder, so that each component has html elements (tags) that vuetify can use.

for now..
not sure if there is a viable way to modify html markup produced by the form builder. 
apply css/scss rules to the formio markup to make it appear like the vuetify stuff

## Checklist


app/frontend/src/assets/scss/variables.scss
```diff
@@ -0,0 +1,7 @@
+// Variables
+$bcgov-blue: #003366;
+$bcgov-blue-secondary: #38598a;
+$bcgov-orange: #fcba19;
+$bcgov-grey: #efefef;
+$bcgov-font: #494949;
+$bcgov-error: #D8292F;
```
TimCsaky, line 7: moving  these into a separate variables file, so we can use them from individual components

app/frontend/src/components/designer/FormDesigner.vue
```diff
@@ -410,3 +410,16 @@ export default {
...
+ @import '~bootstrap/dist/css/bootstrap.min.css';
+ @import 'https://unpkg.com/formiojs@4.11.2/dist/formio.builder.min.css';
+
+// override formio form builder styles here
+ .form-builder-wrapper ::v-deep .formbuilder{
+   .btn-primary{
+     background-color: #003366 !important;
+   }
+ }
+
```
TimCsaky, line 424: This is one way of overriding the formio styles. (?)

bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-wxpbtr-tools.pathfinder.gov.bc.ca/dashboard?id=common-hosted-form-service-pr-44

app/frontend/src/components/designer/FormViewer.vue
```diff
@@ -220,3 +220,13 @@ export default {
...
 </script>
+
+<style lang="scss" scoped>
+
+ .form-wrapper ::v-deep .formio-form{
+   .form-control{
+     border: 1px solid gray;
+   }
+ }
+
```
TimCsaky, line 231: style the form here.
i'm wondering if we can add vuetify class names

app/frontend/src/components/designer/FormDesigner.vue
```diff
@@ -410,3 +410,16 @@ export default {
   },
 };
 </script>
+
+<style lang="scss" scoped>
+ @import '~bootstrap/dist/css/bootstrap.min.css';
+ @import 'https://unpkg.com/formiojs@4.11.2/dist/formio.builder.min.css';
```
loneil, line 416: Yeah this looks promising. If we can only import the bootstrap on the components scoped to hold the form designer that might be the best. If this works out, ideally we wrap the form io component in a slimmer component so the bootstrap is at the lowest level it needs to be. Can coordinate with Jeremy since he's moving the builder/designer components around as it is.

TimCsaky: great. yeah, this bit seems to be working. i'll commit some more changes now for review.. because there seems to be a simpler way of overriding the formio styles.

app/frontend/src/components/designer/FormViewer.vue
```diff
@@ -220,3 +220,8 @@ export default {
   }
 };
 </script>
+
+<style lang="scss" scoped>
+@import '~bootstrap/dist/css/bootstrap.min.css';
+@import 'https://unpkg.com/formiojs@4.11.2/dist/formio.builder.min.css';
+</style>
```
TimCsaky, line 227: to read our custom css rules after formio's we should put them here. 
i've put them in the main styles.scss to keep things organised but i dont believe they have a higher specificity when there in that file.

bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-wxpbtr-tools.pathfinder.gov.bc.ca/dashboard?id=common-hosted-form-service-pr-44

TimCsaky:
i used a merge commit to get this pr up to date with bcgov/master.. and it was messingthings up for me.
so i've created a new PR 48 instead. 🙄 
https://github.com/bcgov/common-hosted-form-service/pull/48

PR #45, opened 2020-10-08
Export submissions UI

loneil:
# Description
https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-1617

Export submissions popup on the View Submissions page. Calls new api endpoint asking for CSV and downloads.
Some minor style tweaks upcoming but will wait until bootstrap stuff isn't overriding.

Tested on Chrome, FF, and iPad safari.

## Types of changes

## Checklist

### DIFF
--- a/app/frontend/src/components/forms/ExportSubmissions.vue
+++ b/app/frontend/src/components/forms/ExportSubmissions.vue
@@ -1,0 +1,42 @@
+<template>
+  <span>
+    <v-dialog v-model="dialog" width="900">
+      <template v-slot:activator="{ on, attrs }">
+        <v-btn color="blue" text small v-bind="attrs" v-on="on">
+          <v-icon class="mr-1">cloud_download</v-icon>
+          <span>Export CSV</span>
+        </v-btn>
+      </template>
+
+      <v-card>
+        <v-card-title class="headline">
+          Export CSV
+          <hr />
+        </v-card-title>
+
+        <v-card-text>
+          <h3>Select the submission date</h3>
+        </v-card-text>
+
+        <v-card-actions class="justify-center">
+          <v-btn class="mb-5 mr-5" color="primary" @click="dialog=false">
+            <span>Export</span>
+          </v-btn>
+          <v-btn class="mb-5" outlined @click="dialog=false">
+            <span>Cancel</span>
+          </v-btn>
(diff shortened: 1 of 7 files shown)


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-wxpbtr-tools.pathfinder.gov.bc.ca/dashboard?id=common-hosted-form-service-pr-45

app/frontend/src/components/forms/ExportSubmissions.vue
```diff
@@ -0,0 +1,174 @@
...
+        </v-card-actions>
+      </v-card>
+    </v-dialog>
+  </span>
+</template>
+
+<script>
+import moment from 'moment';
+import { mapGetters } from 'vuex';
+import formService from '@/services/formService.js';
```
jujaga, line 121: Going to start advocating importing our services via index instead of directly referencing the file
```json
import { formService } from '@/services';
```

TimCsaky (approved):
You've done a great job with this.. it's works beautifully from a ux perspective and the code is MUCH clearer with that vuex plugin. looks great on mobile too.

minor point: in the modal, it could just say 'File Name' (instead of 'file name and type') as the label. But maybe leave that decision if we're doing json export as well.

i guess the csv file should list rows by date created desc, or the same default order as the table in the app (?). you could have validation to ensure startdate is before end date.. but i think returning an empty file is just fine - kind of serves them right.. seriously! 
strange that when i export 'all' firefox offers to open with excel.exe and when i export using a date range, it defaults to notepad.exe. regardless of previous selections.

loneil:
Hmm firefox defaults to submlime for me for me in all cases... chrome goes to excel.

parc-jason:
> Hmm firefox defaults to submlime for me for me in all cases... chrome goes to excel.

Each browser could be different.  One thing to look at is the content-type I put in the header - it is just text/csv, is there a better header?  still think each browser has a different mime-type association to application.

app/frontend/src/components/forms/ExportSubmissions.vue
loneil: We can go round and update them all

parc-jason (approved):
👍

PR #60, opened 2020-10-17
Share/qr/launch should not be version specific

loneil:
# Description
https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-1632
When Launching the form, or getting the Share URL, the version ID should not be relevant. The form submit page should just call the endpoint that fetches the currently published form version.
Old submissions should call the endpoint to get the specific version schema.

https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-1639
Remove breadcrumbs for now

Also was missing the update form call when updating settings on manage, so added that back in.

## Types of changes

## Checklist

### DIFF
--- a/app/frontend/src/components/designer/FormViewer.vue
+++ b/app/frontend/src/components/designer/FormViewer.vue
@@ -41,6 +41,7 @@ export default {
       currentForm: {},
       submissionRecord: {},
       version: 0,
+      versionIdToSubmitTo: this.versionId
     };
   },
   computed: {
@@ -61,11 +62,9 @@ export default {
         const response = await formService.getSubmission(this.submissionId);
         this.submissionRecord = Object.assign({}, response.data.submission);
         this.submission = this.submissionRecord.submission;
-        // simulate getFormName()
         this.formName = response.data.form.name;
-        // simulate getFormSchema()
         this.formSchema = response.data.version.schema;
-        // this.version = response.data.version.version;
+        this.version = response.data.version.version;
       } catch (error) {
         this.addNotification({
           message: 'An error occurred fetching the submission for this form',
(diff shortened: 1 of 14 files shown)


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-wxpbtr-tools.pathfinder.gov.bc.ca/dashboard?id=common-hosted-form-service-pr-60

app/frontend/src/components/designer/FormViewer.vue
```diff
@@ -41,6 +41,7 @@ export default {
       currentForm: {},
       submissionRecord: {},
       version: 0,
+      versionIdToSubmitTo: this.versionId
```
jujaga, line 44: Optional: Could be "versionIdTarget" instead

parc-jason (approved):
👍

app/frontend/src/components/designer/FormViewer.vue
```diff
@@ -61,46 +62,48 @@ export default {
...
       try {
-        const response = await formService.readVersion(
-          this.formId,
-          this.versionId
-        );
-        if (!response.data || !response.data.schema) {
-          throw new Error(
-            `No schema in response. VersionId: ${this.versionId}`
-          );
+        let response = undefined;
```
jujaga, line 78: It doesn't look like we use the response variable outside of the if/else block. In this case, it's better to declare const response within the if/else blocks instead as it would be a tighter scope.

jujaga (approved):
Looks pretty good. A few minor tweaks for your consideration:

app/frontend/src/components/designer/FormViewer.vue
loneil: Ah yeah, leftover from earlier attempt. Fixed

PR #64, opened 2020-10-19
Try a custom component fix for Edge

loneil:
# Description

## Checklist


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-wxpbtr-tools.pathfinder.gov.bc.ca/dashboard?id=common-hosted-form-service-pr-64

jujaga (approved):
Moving this out into the component is a good idea anyways as it allows us to have less in the initial application loading. I think we should go through with this 👍

loneil:
Sadly this doesn't work, as nice as it would be to have them in the formio components themselves it seems to need to load before the keycloak adapter or KC just breaks (polyfill clash I think)

PR #68, opened 2020-10-20
Export Submissions - ordering, UX changes

parc-jason:
# Description
Export Submissions Ordering on backend, proper mime type on frontend download.
[SHOWCASE-1657](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-1657)
Some styling on Submissions dialog - Date Range outline, Radio button line height.
[SHOWCASE-1661](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-1661)
Fix Header text alignment
[SHOWCASE-1667](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-1667)
Add Timestamp to Submissions table
[SHOWCASE-1668](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-1668)
Rebrand Basic Mode to Simple Mode for form designer
[SHOWCASE-1669](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-1669)
Update Vuetify Checkbox and Radio Button label styles to correct vertical alignment
[SHOWCASE-1656](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-1656)

## Types of changes

## Checklist

### DIFF
--- a/app/frontend/src/components/forms/ExportSubmissions.vue
+++ b/app/frontend/src/components/forms/ExportSubmissions.vue
@@ -178,7 +178,7 @@ export default {
 };
 </script>
 
-<style lang="scss" scoped>
+<style lang="scss">
 
 .export-submissions-dlg {
   // vuetify radio button labels need alignment help
--- a/app/frontend/src/components/forms/ExportSubmissions.vue
+++ b/app/frontend/src/components/forms/ExportSubmissions.vue
@@ -137,15 +137,16 @@ export default {
     ...mapActions('notifications', ['addNotification']),
     async callExport() {
       try {
+        // UTC start of selected start date...
         const from =
           this.dateRange && this.startDate
-            ? moment.utc(this.startDate).format()
+            ? moment(this.startDate, 'YYYY-MM-DD hh:mm:ss').utc().format()
             : undefined;
+        // UTC end of selected end date...
         const to =
           this.dateRange && this.endDate
-            ? moment.utc(this.endDate).format()
+            ? moment(`${this.endDate} 23:59:59`, 'YYYY-MM-DD hh:mm:ss').utc().format()
             : undefined;
-
         const response = await formService.exportSubmissions(
           this.form.id,
           from,
--- a/app/frontend/src/components/forms/ExportSubmissions.vue
+++ b/app/frontend/src/components/forms/ExportSubmissions.vue
@@ -182,12 +182,6 @@ export default {
 <style lang="scss">
 
 .export-submissions-dlg {
-  // vuetify radio button labels need alignment help
-  .v-radio {
-    > label {
-      margin-bottom: 0;
-    }
-  }
 
   // legend with no text breaks outline on date picker
   .v-input__slot {
--- a/app/frontend/src/components/forms/ExportSubmissions.vue
+++ b/app/frontend/src/components/forms/ExportSubmissions.vue
@@ -1,6 +1,6 @@
 <template>
   <span>
-    <v-dialog v-model="dialog" width="900">
+    <v-dialog v-model="dialog" width="900" content-class="export-submissions-dlg">
       <template v-slot:activator="{ on, attrs }">
         <v-btn color="primary" text small v-bind="attrs" v-on="on">
           <v-icon class="mr-1">cloud_download</v-icon>
@@ -48,7 +48,6 @@
                     v-model="startDate"
                     data-test="picker-form-startDate"
                     @input="startDateMenu = false"
-                    :readonly="reviewMode"
                   ></v-date-picker>
                 </v-menu>
               </v-col>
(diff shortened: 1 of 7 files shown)


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-wxpbtr-tools.pathfinder.gov.bc.ca/dashboard?id=common-hosted-form-service-pr-68

jujaga (approved):
Looks pretty good. Only concern thusfar:
* When I export CSV with select date range set to start today and end on today, I get no results. Is this intended? (May be inclusion/exclusion or off-by-one behavior)

parc-jason:
> Looks pretty good. Only concern thusfar:
> 
> * When I export CSV with select date range set to start today and end on today, I get no results. Is this intended? (May be inclusion/exclusion or off-by-one behavior). Exporting all behaves as intended.

The FE should be sending in a better range.  I'll change the logic to be startdate 00:00:00 and endDate 23:59:59.

app/frontend/src/components/forms/SubmissionsTable.vue
```diff
@@ -90,6 +90,9 @@ export default {
   },
   methods: {
     ...mapActions('form', ['fetchForm', 'fetchSubmissions']),
+    formatDate(date) {
+      return date ? new Date(date).toLocaleString() : 'N/A';
+    },
```
loneil, line 95: This should use the `| formatDate` filter so there don't have to be date formatter methods down in each component.

loneil (changes requested):

app/frontend/src/components/forms/SubmissionsTable.vue
```diff
@@ -99,7 +102,7 @@ export default {
           const tableRows = this.submissionList.map((s) => {
             return {
               confirmationId: s.confirmationId,
-              date: s.createdAt,
+              date: this.formatDate(s.createdAt),
```
loneil, line 105: See above. Filters are used everywhere date formatting needs to be to encapsulate that logic. There's a `formatDate` and `formatDateLong` filter there already. Can add additional flags (can pass params to filters) or another filter if a different format is wanted.

app/frontend/src/components/forms/ExportSubmissions.vue
```diff
@@ -176,3 +178,22 @@ export default {
...
+
+<style lang="scss">
+
+.export-submissions-dlg {
+  // vuetify radio button labels need alignment help
+  .v-radio {
+    > label {
+      margin-bottom: 0;
+    }
+  }
```
loneil, line 190: Left this comment on Tim's PR #69 looks like he's adjusting it there too so would conflict with this PR. Should scope higher than the export dialog anyways.

_If we're going to adjust the radio button vertical alignment issue it should be a global style setting. It's not just this export dialog with the issue
See "Vertical alignment of text for Checkboxes and Radio buttons (see Create New Form) too high." in https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-1656

This is (another) issue introduced by Bootstrap it looks like. So if we do want to address it quickly with CSS overrides can apply a class to places it needs to be fixed globablly in style.scss maybe rather than scoped.

We should double check the radios and checkboxes on the actual FormIO forms themselves too._

app/frontend/src/components/forms/SubmissionsTable.vue
parc-jason: it did have formatDate, didn't know about those filters.  will set it to formatDateLong.

loneil: Filters are great, didn't use them on any other Vue apps until now. The existing one uses the long (words) date format, which I think Mina has used in mockups but if we want something else can pass a format to a filter too.

app/frontend/src/components/forms/ExportSubmissions.vue
parc-jason: yup that can come out when we get a global solution.

loneil: One of you or @TimCsaky will have to take out your implementation though since you're both doing it in pr 68 and 69 in different ways. (or both take it out and move to style.scss)

app/frontend/src/components/forms/SubmissionsTable.vue
parc-jason: done

app/frontend/src/components/forms/ExportSubmissions.vue
parc-jason: left a comment in @TimCsaky PR that this has moved globally.

loneil (approved):
Great stuff, good call on the different times on the start/end date

PR #69, opened 2020-10-20
style fixes

TimCsaky:
# Description

- date range outlines are partially missing (this might be a simpler way of doing it until we restructure css stuff)
- Export and Cancel buttons appear to be missing side padding
- Button styling for Secondary... Blue outline with white background.
- readonly select boxes need to look disabled (greyed out)
- formio component button styles
- alignment of links with icons in my forms table

## Checklist


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-wxpbtr-tools.pathfinder.gov.bc.ca/dashboard?id=common-hosted-form-service-pr-69

app/frontend/src/components/forms/ExportSubmissions.vue
```diff
@@ -176,3 +172,8 @@ export default {
   },
 };
 </script>
+<style scoped>
+.export-radios ::v-deep .v-label {
+  margin-bottom: 0 !important;
+}
```
loneil, line 178: If we're going to adjust the radio button vertical alignment issue it should be a global style setting. It's not just this export dialog with the issue
See "Vertical alignment of text for Checkboxes and Radio buttons (see Create New Form) too high." in https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-1656

This is (another) issue introduced by Bootstrap it looks like. So if we do want to address it quickly with CSS overrides can apply a class to places it needs to be fixed globablly in style.scss maybe rather than scoped.

We should double check the radios and checkboxes on the actual FormIO forms themselves too.

loneil (changes requested):
Looks good, just should address the radio spacing at a higher level to fix other spots as well.

app/frontend/src/components/forms/ExportSubmissions.vue
parc-jason: i've added a global style in my PR for vuetify checkbox and radio button labels, so this isn't needed.

TimCsaky: ok. thank you. i'll remove this and test.

TimCsaky: formio checkbox/radio labels look good on firefox and edge.

app/frontend/src/components/forms/ExportSubmissions.vue
```diff
@@ -39,9 +39,7 @@
                       readonly
                       v-on="on"
                       dense
-                      flat
```
parc-jason, line 42: This works a LOT better and is simpler than what I did in my PR!
We will have to remove my style and use this approach. 👍

parc-jason (commented):
Is there another ticket to change some more styles?
The formio buttons need to have their text bolded and uppercased like our other buttons.
I also have some tweak that isn't a ticket yet - the text in the my forms table (a > i ~ span should be vertical align middle to have the text line up with the icon...

app/frontend/src/components/forms/ExportSubmissions.vue
TimCsaky: I've added the two new style tweaks you mentioned in the comment above.
- links in my forms table
- formio button style

loneil (approved):
Looking really good

PR #72, opened 2020-10-21
About page video and images

loneil:
# Description
https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-1548

Add final (for now) video and images. Media is linked from storage in github wiki.
Pop-out images when clicked to show full size.

## Types of changes

## Checklist


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-wxpbtr-tools.pathfinder.gov.bc.ca/dashboard?id=common-hosted-form-service-pr-72

bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-wxpbtr-tools.pathfinder.gov.bc.ca/dashboard?id=common-hosted-form-service-pr-72

app/frontend/tests/unit/views/About.spec.js
```diff
@@ -1,18 +1,35 @@
...
-    vuetify = new Vuetify();
+    store = new Vuex.Store();
+    store.registerModule('auth', {
+      namespaced: true,
+      getters: {
+        authenticated: () => true,
+        isAdmin: () => false,
+        keycloakReady: () => true,
+        // eslint-disable-next-line no-unused-vars
+        createLoginUrl: () => options => 'testurl'
```
jujaga, line 23: This can just be
```js
createLoginUrl: () => () => 'testurl'
```
Goal of this is to make it return a function which when called returns string 'testurl' - no need to specify a param when it's not used.

TimCsaky (approved):
nice! 
There are more features worth advertising (submission export, form sharing, access etc) but I guess that will come in a later version.
That image modal thing will be useful.
Also probably not necessary at this stage, but 'Sign in now to try it out ' could only show if you're not logged in.

app/frontend/src/views/form/Create.vue
```diff
@@ -70,5 +70,13 @@ export default {
...
     },
   },
+  beforeRouteLeave(to, from, next) {
+    const answer = window.confirm('Do you really want to leave this page?');
+    if (answer) {
+      next();
+    } else {
+      next(false);
+    }
+  },
```
jujaga, line 80: Could do
``` js
(answer) ? next() : next(false);
```
Also as a further note, I wonder if it would be prudent down the line to also move the standard nav leave stuff into here instead of having it reside in the component?

loneil: This... isn't supposed to be in this PR. I must have comitted accidentally. The nav leave stuff isn't working any more as well... I'm looking at all that right now.
This can merge in though but will be adjusted.

loneil: The warning on leave when creating a form broke here
image

PR #76, opened 2020-10-22
more style fixes

TimCsaky:
# Description

- show red borders on invalid input fields (previously overridden by mistake)
- remove red alert icon on invalid fields

## Checklist

### DIFF
--- a/app/frontend/src/assets/scss/style.scss
+++ b/app/frontend/src/assets/scss/style.scss
@@ -284,6 +284,30 @@ a,
     background-image: none;
   }
 
+  // add rounded corners to Date / Time Component input groups and red border if invalid
+  .input-group > .flatpickr-input.form-control {
+    & + .form-control.input {
+      border-top-left-radius: 0.25rem;
+      border-bottom-left-radius: 0.25rem;
+    }
+    & ~ .input-group-append {
+      border: 1px solid $bcgov-input-placeholder;
+      border-top-right-radius: 0.25rem;
+      border-bottom-right-radius: 0.25rem;
+    }
+
+    // if validation error make borders red;
+    &.is-invalid + .form-control.input {
+      border: 1px solid #dc3545;
+      border-right: 1px solid $bcgov-input-placeholder;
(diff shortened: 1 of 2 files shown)


app/frontend/src/assets/scss/style.scss
```diff
@@ -264,15 +264,23 @@ a,
...
       color: $bcgov-input-placeholder;
       opacity: inherit;
     }
   }
 
+  .form-control.is-invalid,
+  .was-validated .form-control:invalid {
+    background-image: none;
+  }
+
```
loneil, line 283: I think we need to be really careful with this. Are there other background images used for other formio functionality that having a `background-image: none` could override? Like some other control using an icon that's neccessary and having it disappear on validation could be unintended behaviour.

What's the intent of removing this as well? It's a FormIO default isn't it. We've decided not to mess around with basic FormIO behaviour of the rendered controls.

Personally I'd rather stay away from removing things like this unless it's an absolute requirement.
@parc-jason you've played more with the actual formIO controls than I have.

bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-wxpbtr-tools.pathfinder.gov.bc.ca/dashboard?id=common-hosted-form-service-pr-76

app/frontend/src/assets/scss/style.scss
TimCsaky: yes. good point. maybe we can just apply this to select fields.
this was done to address the issue with the red circle icon appearing over the dropdown arrow. 
image

parc-jason: @loneil  - i don't really know what styles etc apply to formio.  but I agree, we have to be super particular with any style override.  between vuetify and bootstrap and formio, there are a lot of implications.  i think we are going to have to accept that not everything will look like the bc gov design system, and any style/class override that must be done (like above) has to be done in the most specific way possible.  the finest grained scope that will correct the issue.

loneil (approved):

PR #78, opened 2020-10-26
Route/Controller Basic Unit Tests

parc-jason:
This is just to ensure proper response codes and routes exist.
This DOES NOT test middleware or keycloak protect, assume caller always has permissions to call the route.
This DOES NOT test the body of the response, or any data access layer code.

See [SHOWCASE-1650](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-1650)

# Description

## Types of changes

## Checklist

### DIFF
--- a/app/tests/unit/routes/v1/admin.spec.js
+++ b/app/tests/unit/routes/v1/admin.spec.js
@@ -1,0 +1,144 @@
+const request = require('supertest');
+const Problem = require('api-problem');
+
+const helper = require('../../../common/helper');
+
+//
+// mock middleware
+//
+const keycloak = require('../../../../src/components/keycloak');
+//
+// test assumes that caller has appropriate token, we are not testing middleware here...
+//
+keycloak.protect = jest.fn(() => {
+  return jest.fn((req, res, next) => {
+    next();
+  });
+});
+
+const userAccess = require('../../../../src/forms/auth/middleware/userAccess');
+userAccess.currentUser = jest.fn((req, res, next) => {
+  next();
+});
+
+//
+// we will mock the underlying data service calls...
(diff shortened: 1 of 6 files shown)


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-wxpbtr-tools.pathfinder.gov.bc.ca/dashboard?id=common-hosted-form-service-pr-78

app/tests/unit/routes/v1/admin.spec.js
```diff
@@ -0,0 +1,144 @@
...
+//
+const keycloak = require('../../../../src/components/keycloak');
+//
+// test assumes that caller has appropriate token, we are not testing middleware here...
+//
+keycloak.protect = jest.fn(() => {
+  return jest.fn((req, res, next) => {
+    next();
+  });
+});
```
jujaga, line 17: This appears pretty frequently in all the new unit test files. As this is mocking out keycloak.protect, ideally this should be wrapped into a beforeAll or beforeEach clause, and then a `keycloak.protect.mockRestore()` in an afterEach or afterAll clause so that there's a stronger guarantee that these unit tests don't accidentally end up trampling on other unrelated unit tests that also happen to mock out keycloak.protect in some different capacity.

Same advice applies for the other two mocks below (userAccess, service).

jujaga (approved):
Looks awesome! 👍
Only other thing I can think of when looking at the coverage table on Jenkins is that we should ignore out the `db/migrations` and `db/seeds` directories... IIRC that'd need to be added to Line 4 of `app/jest.config.js`.

loneil (approved):

PR #81, opened 2020-10-28
Wording to make form confirmation ID optional for now

loneil:
# Description
For the COVID desk form (until we give the designer the ability to choose whether the confirmation message shows), make the wording to the user not seem mandatory.

## Types of changes

## Checklist

### DIFF
--- a/app/frontend/src/views/form/Success.vue
+++ b/app/frontend/src/views/form/Success.vue
@@ -8,7 +8,8 @@
             Your form has been submitted successfully
           </h1>
           <h3>
-            Please keep the following Confirmation ID for your records:
+            If you wish to keep a record of this submission, you can keep the
+            following Confirmation ID:
             <strong>{{ s.substring(0, 8).toUpperCase() }}</strong>
           </h3>
           <hr />


app/frontend/src/views/form/Success.vue
```diff
@@ -8,7 +8,8 @@
             Your form has been submitted successfully
           </h1>
           <h3>
-            Please keep the following Confirmation ID for your records:
+            If you wish to keep a record of this submission, you can keep the
+            following Confirmation ID:
```
jujaga, line 12: Could do `Your submission's Confirmation ID is:` - if we state it factually instead of as a hypothetical, it could achieve the same thing and also be a shorter message.

jujaga (approved):

bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-wxpbtr-tools.pathfinder.gov.bc.ca/dashboard?id=common-hosted-form-service-pr-81

app/frontend/src/views/form/Success.vue
loneil: Need to drive home that they really don't need it for this form. Going to be changed in short order anyways when we remove it for the COVID form

PR #83, opened 2020-10-28
API for form delete (soft).

parc-jason:
# Description
Use DELETE /forms/:formId for delete, allows no confusion between update and delete.  

After discussions with the team, by default, we do not return deleted/inactive forms - treat as 404 if queried by id, do not return in lists.  The caller must be explicit in getting deleted forms.

Examples:
GET /forms:/formId?active=false is valid to retrieve an inactive form 
GET /forms:/formId/version?active=false is valid to retrieve an inactive form and current version
GET /forms?active=false is valid to return a list of invalid forms
GET /rbac/currentUser?active=false

We will leave the resurrection and management of deleted forms to CHEFS Admin users under the /admin API (TBD).

See [SHOWCASE-1724](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-1724)

## Types of changes

## Checklist

### DIFF
--- a/app/src/forms/auth/middleware/userAccess.js
+++ b/app/src/forms/auth/middleware/userAccess.js
@@ -45,10 +45,14 @@ const hasFormPermissions = (permissions) => {
       // cannot find the currentUser... guess we don't have access... FAIL!
       return next(new Problem(401, {detail: 'Form Id not found on request.'}));
     }
-    const form = req.currentUser.forms.find(f => f.formId === formId);
+    let form = req.currentUser.forms.find(f => f.formId === formId);
     if (!form) {
-      // cannot find the form... guess we don't have access... FAIL!
-      return next(new Problem(401, {detail: 'Current user has no access to form.'}));
+      // check deleted... (this allows 404 on other queries later)
+      form = req.currentUser.deletedForms.find(f => f.formId === formId);
+      if (!form) {
+        // cannot find the form... guess we don't have access... FAIL!
+        return next(new Problem(401, {detail: 'Current user has no access to form.'}));
+      }
     }
 
     if (!Array.isArray(permissions)) {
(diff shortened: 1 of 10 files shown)


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-wxpbtr-tools.pathfinder.gov.bc.ca/dashboard?id=common-hosted-form-service-pr-83

bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-wxpbtr-tools.pathfinder.gov.bc.ca/dashboard?id=common-hosted-form-service-pr-83

app/src/forms/auth/middleware/userAccess.js
```diff
@@ -45,10 +45,14 @@ const hasFormPermissions = (permissions) => {
       // cannot find the currentUser... guess we don't have access... FAIL!
       return next(new Problem(401, {detail: 'Form Id not found on request.'}));
     }
-    const form = req.currentUser.forms.find(f => f.formId === formId);
+    let form = req.currentUser.forms.find(f => f.formId === formId);
     if (!form) {
-      // cannot find the form... guess we don't have access... FAIL!
-      return next(new Problem(401, {detail: 'Current user has no access to form.'}));
+      // check deleted... (this allows 404 on other queries later)
+      form = req.currentUser.deletedForms.find(f => f.formId === formId);
```
jujaga, line 51: If we're just using form once for the boolean check on line 52, it's better to keep the form variable declaration as a const on line 48, and then declare a new const in here like `deletedForms` or something. Generally better to avoid variable reassignments unless absolutely necessary.

parc-jason: The form variable is used later to determine the permissions (line 63)

loneil (approved):

PR #86, opened 2020-10-29
Delete button

loneil:
# Description
https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-1725
Enable Delete button on management page and call new api endpoint to soft delete.
Other places that load forms seem to handle this fine you either get
- Form missing in lists when inactive
- Error (404) trying to get an inactive form. Will create a ticket to determine if we want a different view behaviour here in the different cases.

## Types of changes

## Checklist

### DIFF
--- a/app/frontend/src/components/base/BaseDialog.vue
+++ b/app/frontend/src/components/base/BaseDialog.vue
@@ -2,7 +2,7 @@
   <v-dialog
     :max-width="width"
     persistent
-    v-if="show"
+    v-bind:value="value"
     @click:outside="closeDialog"
     @keydown.esc="closeDialog"
   >
@@ -31,14 +31,19 @@
           </v-btn>
         </div>
         <div v-else-if="type === 'CONTINUE'">
-          <v-btn class="mb-5 mr-5" outlined @click="closeDialog">
-            <slot name="button-text-cancel">
-              <span>Cancel</span>
-            </slot>
-          </v-btn>
-          <v-btn class="mb-5" color="primary" depressed @click="continueDialog">
+          <v-btn
+            class="mb-5 mr-5"
+            color="primary"
+            depressed
+            @click="continueDialog"
+          >
             <slot name="button-text-continue">
               <span>Continue</span>
+            </slot>
+          </v-btn>
+          <v-btn class="mb-5" outlined @click="closeDialog">
+            <slot name="button-text-cancel">
+              <span>Cancel</span>
             </slot>
           </v-btn>
         </div>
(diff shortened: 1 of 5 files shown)


app/frontend/src/components/base/BaseDialog.vue
```diff
@@ -2,7 +2,7 @@
   <v-dialog
     :max-width="width"
     persistent
-    v-if="show"
+    v-bind:value="value"
```
jujaga, line 5: Can shorthand this to just `:value="value"`

app/frontend/src/views/form/Manage.vue
```diff
@@ -18,9 +18,25 @@
...
+          <BaseDialog
+            v-model="showDeleteDialog"
+            type="CONTINUE"
+            @close-dialog="showDeleteDialog = false"
+            @continue-dialog="deleteForm"
+          >
+            <template #title>Confirm Deletion</template>
+            <template #text>
+              Are you sure you wish to delete <strong>{{ form.name }}</strong
+              >? This form will no longer be accessible.
```
jujaga, line 34: Strange vetur bracket wraparounds.

bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-wxpbtr-tools.pathfinder.gov.bc.ca/dashboard?id=common-hosted-form-service-pr-86

app/frontend/src/views/form/Manage.vue
```diff
@@ -203,15 +217,17 @@ export default {
       this.formSettingsDisabled = true;
       this.fetchForm(this.f);
     },
+    deleteForm() {
+      this.showDeleteDialog = false;
+      this.deleteCurrentForm();
```
jujaga, line 222: Do we need to make this an async/await to make sure it's deleted before navigating?

app/frontend/src/components/base/BaseDialog.vue
loneil: I kind of want to leave this as v-bind since it's so prominent in the docs of how to pass a v-model to a component

app/frontend/src/views/form/Manage.vue
loneil: No was going to navigate away regardless and the error will pop if it fails. Not losing any work if they have to click back and try again.

bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-wxpbtr-tools.pathfinder.gov.bc.ca/dashboard?id=common-hosted-form-service-pr-86

app/frontend/src/views/form/Manage.vue
loneil: Fixed

parc-jason (approved):
👍 👍

app/frontend/src/views/form/Manage.vue
jujaga: For the current case it should be fine as all operations get shunted to vuex anyways (error notification, etc). It's more in the case of if it fails to delete, we don't want them to navigate away, and leave them on the Manage page with the error instead.

PR #95, opened 2020-11-06
File Uploads / Submission Attachments

parc-jason:
# Description
Formio component to upload file(s) to CHEFS.
Tried to alter as little of existing formio code for File Upload component using URL provider.

File Storage handling - local files, object storage files.
Associate uploaded files with submissions when a submission is created.
Files API protected for read/delete by submission permissions (once associated with submission).

## Types of changes

## Checklist


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-wxpbtr-tools.pathfinder.gov.bc.ca/dashboard?id=common-hosted-form-service-pr-95

bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-wxpbtr-tools.pathfinder.gov.bc.ca/dashboard?id=common-hosted-form-service-pr-95

app/config/test.json
```diff
@@ -3,6 +3,27 @@
...
+      "fileCount": "1",
+      "fileKey": "files",
+      "fileMaxSize": "1MB",
+      "fileMinSize": "0KB",
+      "path": "files"
+    },
+    "permanent": "localStorage",
+    "localStorage" : {
+      "path": ""
+    },
```
jujaga, line 18: test.json only needs to define values that aren't already defined in default.json - likely don't need any of these attributes highlighted.

app/frontend/src/assets/scss/style.scss
```diff
@@ -443,6 +443,24 @@ a,
...
+
+.form-wrapper {
+   // the drag/drop area needs to line up with vuetify v-application ul styles...
+  .formio-component-file {
+    .fileSelector {
+      margin-left: 24px;
+    }
+    .file {
+      .row {
+        margin-left: 24px;
```
jujaga, line 460: If possible, would be better to define this in rem units instead of px (I think this would be 1.5rem).

app/package.json
```diff
@@ -41,23 +41,28 @@
...
     "axios-oauth-client": "^1.3.0",
     "axios-token-interceptor": "^0.2.0",
     "body-parser": "^1.19.0",
+    "bytes": "^3.1.0",
     "compression": "^1.7.4",
     "config": "^3.3.2",
     "crypto": "^1.0.1",
     "express": "^4.17.1",
     "express-rate-limit": "^5.1.3",
+    "express-rate-limiter": "^1.3.1",
```
jujaga, line 55: Why do we now have two express-rate-limit packages? Should only have one of them if possible and drop the other.

app/src/forms/file/controller.js
```diff
@@ -0,0 +1,81 @@
...
+        // will throw permissions error if user not allowed to read
+        await submissionService.read(fileStorage.formSubmissionId, req.currentUser);
+      }
+
+      // ok, let's go get the binary...
+      const stream = await storageService.read(fileStorage);
+
+      stream.on('error', function error(err) {
+        throw (err);
+      });
```
jujaga, line 46: This can be an arrow function callback
```js
stream.on('error', (err) => throw (err));
```

app/src/forms/file/storage/localStorageService.js
```diff
@@ -0,0 +1,70 @@
...
+    }
+  },
+
+  uploadFile: async(fileStorage) => {
+    // we do not upload, so return current state.
+    return {
+      path: fileStorage.path,
+      storage: fileStorage.storage
+    };
+  }
```
jujaga, line 66: Just as a note - if most fs file operations are using the synchronous variant, these services may not necessarily need to be async at all. It's either that, or we could consider changing the file operation calls to their asynchronous version where appropriate (this may improve microservice performance on heavier loads)

app/src/forms/file/storage/objectStorageService.js
```diff
@@ -0,0 +1,213 @@
...
+
+}
+
+const endpoint = config.get('files.objectStorage.endpoint');
+const bucket = config.get('files.objectStorage.bucket');
+const key = config.get('files.objectStorage.key');
+const accessKeyId = config.get('files.objectStorage.accessKeyId');
+const secretAccessKey = config.get('files.objectStorage.secretAccessKey');
+
+let objectStorageService = new ObjectStorageService({accessKeyId: accessKeyId, secretAccessKey: secretAccessKey, endpoint: endpoint, bucket: bucket, key: key});
```
jujaga, line 211: As an aside, you could potentially get away with just getting `config.get('files.objectStorage')` and then pass that object directly to the ObjectStorageService constructor (all the object parameters are named the same way already).

Also, more async and file sync considerations as with earlier comment.

jujaga (commented):
Looks like it's working as you'd expect. Definitely should take a look over the backend async service calls that do fs synchronous calls and see whether they should really be asynchronous or synchronous in implementation.

app/config/test.json
parc-jason: mostly just there to ensure that we don't use defaults or env vars for testing.  don't necessarily know what values would be in env vars, and probably don't want large values or objectstorage... who knows? but it should be known what values are in use for tests, and this is the way.

app/frontend/src/assets/scss/style.scss
parc-jason: only in place because we are making this match exactly what vuetify does to the ul under the file selector - and it uses 24px.

app/package.json
parc-jason: yup, we don't use that one... removing express-rate-limiter

app/src/forms/file/storage/localStorageService.js
parc-jason: async are in place to use the same 'interface' as other storage calls.  so the caller always expects async and awaits if wanted.  if we find performance lags on fs sync, we can swap to async without changing the wrapping storageService.

app/src/forms/file/storage/objectStorageService.js
parc-jason: could do, but prefer to be explicit and have it blow up earlier in the call stack if values are not in config.  line error will point out which one is failing.
async is more for the storageService interface, if we ever put in storage that only has async calls, we don't have to change the interface at all. if we see speed issues with any sync calls in *StorageService, we can flip to async (if supported) without changing storageService or it's callers.

jujaga (approved):

loneil (approved):

TimCsaky (approved):
Good stuff!

PR #99, opened 2020-11-16
Allow non-admin form creation and management

loneil:
# Description
https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-1708
Remove ADMIN blocks for form actions

Also:
Fix an issue where the checkbox-disabling logic was flipped on the Submitter column on Team Management (should be disabled for **non**-TEAM forms.

## Types of changes

## Checklist

## Further comments

Removed Admin checks from BaseSecure. So you can still URL into a page where you can't do anything if you don't have the appropriate rights (like can go to Manage page), but this is fine for now since you can't read any data or do any action that you don't have rights for anyways (buttons are controlled by access rights, and of course the API is as well).
Could look into a more complex BaseSecure control later if we really want to present "You don't have permission to do anything here" pages instead.

### DIFF
--- a/app/src/forms/auth/middleware/userAccess.js
+++ b/app/src/forms/auth/middleware/userAccess.js
@@ -43,7 +43,7 @@ const hasFormPermissions = (permissions) => {
     // If we invoke this middleware and the caller is acting on a specific formId, whether in a param or query (precedence to param)
     const formId = req.params.formId || req.query.formId;
     if (!formId) {
-      // cannot find the currentUser... guess we don't have access... FAIL!
+      // No form provided to this route that secures based on form... that's a problem!
       return next(new Problem(401, {detail: 'Form Id not found on request.'}));
     }
     let form = req.currentUser.forms.find(f => f.formId === formId);
--- a/app/src/forms/auth/middleware/userAccess.js
+++ b/app/src/forms/auth/middleware/userAccess.js
@@ -40,7 +40,8 @@ const hasFormPermissions = (permissions) => {
       // cannot find the currentUser... guess we don't have access... FAIL!
       return next(new Problem(401, {detail: 'Current user not found on request.'}));
     }
-    const formId = req.params.formId;
+    // If we invoke this middleware and the caller is acting on a specific formId, whether in a param or query (precedence to param)
+    const formId = req.params.formId || req.query.formId;
     if (!formId) {
       // cannot find the currentUser... guess we don't have access... FAIL!
       return next(new Problem(401, {detail: 'Form Id not found on request.'}));
--- a/app/src/forms/rbac/routes.js
+++ b/app/src/forms/rbac/routes.js
@@ -1,4 +1,3 @@
-const config = require('config');
 const routes = require('express').Router();
 const currentUser = require('../auth/middleware/userAccess').currentUser;
 
--- a/app/src/forms/rbac/routes.js
+++ b/app/src/forms/rbac/routes.js
@@ -15,39 +15,39 @@ routes.get('/idps', async (req, res, next) => {
   await controller.getIdentityProviders(req, res, next);
 });
 
-routes.get('/forms', keycloak.protect(`${config.get('server.keycloak.clientId')}:admin`), async (req, res, next) => {
+routes.get('/forms', async (req, res, next) => {
   await controller.getFormUsers(req, res, next);
 });
 
-routes.put('/forms', keycloak.protect(`${config.get('server.keycloak.clientId')}:admin`), async (req, res, next) => {
+routes.put('/forms', async (req, res, next) => {
   await controller.setFormUsers(req, res, next);
 });
 
-routes.get('/users', keycloak.protect(`${config.get('server.keycloak.clientId')}:admin`), async (req, res, next) => {
+routes.get('/users', async (req, res, next) => {
   await controller.getUserForms(req, res, next);
 });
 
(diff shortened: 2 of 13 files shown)


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-wxpbtr-tools.pathfinder.gov.bc.ca/dashboard?id=common-hosted-form-service-pr-99

bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-wxpbtr-tools.pathfinder.gov.bc.ca/dashboard?id=common-hosted-form-service-pr-99

jujaga (approved):

app/src/forms/rbac/routes.js
```diff
@@ -15,39 +14,39 @@ routes.get('/idps', async (req, res, next) => {
   await controller.getIdentityProviders(req, res, next);
 });
 
-routes.get('/forms', keycloak.protect(`${config.get('server.keycloak.clientId')}:admin`), async (req, res, next) => {
+routes.get('/forms', async (req, res, next) => {
```
parc-jason, line 17: For getUserForms/getFormUsers and setUserForms/setFormUsers, I think you will want to do a check (in middleware possibly, or controller or service) that the current user has team management permissions on any form id. Most importantly on the set* calls.  Don't want someone mucking with forms they don't have permission on.

app/src/forms/rbac/routes.js
```diff
@@ -15,39 +14,39 @@ routes.get('/idps', async (req, res, next) => {
...
   await controller.getUserForms(req, res, next);
 });
 
-routes.put('/users', keycloak.protect(`${config.get('server.keycloak.clientId')}:admin`), async (req, res, next) => {
+routes.put('/users', async (req, res, next) => {
   await controller.setUserForms(req, res, next);
 });
 
-routes.get('/', keycloak.protect(`${config.get('server.keycloak.clientId')}:admin`), async (req, res, next) => {
+routes.get('/', async (req, res, next) => {
```
parc-jason, line 33: I don't think these functions/endpoints (list, create, read, update, delete) are in use on the UX, I would either remove them, deprecate them (410?) or leave the admin check.  Don't want these "wide" open.  Admin check left in at a minimum - maybe moved under /admin/rbac (would also apply to roles  & permissions)?

parc-jason (changes requested):
Some important endpoints unprotected.  Think we need business level checks.

app/src/forms/rbac/routes.js
loneil: Hmm damn, yeah. For the set ones I think we can alter the userAccess.hasFormPermissions check to look for a query param (if no path param) and use that middleware as it's used elsewhere.
For getUserForms not really sure.. that one is based on just userID isn't it. So no permission check would work.

app/src/forms/rbac/routes.js
loneil: Yeah we should take them out I think until needed if ever.

loneil: Yeah will leave downstream methods but chop the route for now.

app/src/forms/rbac/routes.js
parc-jason: probably put it in the service, filter out form ids user doesn't have access to before acting on them.  don't think need to raise 4** errors, just remove them from the work and the response should indicate what was actually done.

loneil: Ok @parc-jason does this change make sense to you (on this PR already and I've tested all cases out)?
- Change the usedAccess.hasFormPermissions to check a query param if it's optionally invoked on a path with a formId query param (line 48 of that file)

In Rbac routes
- getFormUsers
- setFormUsers
- setUserForms
All these operate on a specific formID, and only on that form. So invoke the new hasFormPermissions check with the query param.

For
- getUserForms
This one is really an administrative function, it returns all form for a user that the caller specifies. (the FE should be using /current). So secure with admin flag.
This is actually not called by the FE right now always so doesn't matter. But would be useful for an admin panel.
Will rethink some of these routes in context of an admin panel in the future (likely duplicate with different permissions under a /admin or something)

parc-jason: yeah, i think that's the best initial go.  you need to make sure the UX is calling setUserForms with a formId parameter (because I don't believe it is mandatory, but is needed for hasFormPermissions.

app/src/forms/auth/middleware/userAccess.js
```diff
@@ -40,7 +40,8 @@ const hasFormPermissions = (permissions) => {
       // cannot find the currentUser... guess we don't have access... FAIL!
       return next(new Problem(401, {detail: 'Current user not found on request.'}));
     }
-    const formId = req.params.formId;
+    // If we invoke this middleware and the caller is acting on a specific formId, whether in a param or query (precedence to param)
+    const formId = req.params.formId || req.query.formId;
     if (!formId) {
       // cannot find the currentUser... guess we don't have access... FAIL!
```
parc-jason, line 46: i think this is my comment and it should be cannot find the form for the currentUser... d'oh.

parc-jason (approved):
I like that change for checking user/form permissions on managing the team 👍 👍

app/src/forms/rbac/routes.js
loneil: Great, thanks for the review and catching this!

app/src/forms/auth/middleware/userAccess.js
loneil: 👍

PR #108, opened 2020-12-03
Default ID mode in the form settings

loneil:
# Description
We set the default userType to blank in the store since it was used by Team Management as well. However, when creating a new form we still want the Team option checked by default, so specify in there when that component is created.

## Types of changes

## Checklist

### DIFF
--- a/app/frontend/src/components/designer/FormSettings.vue
+++ b/app/frontend/src/components/designer/FormSettings.vue
@@ -211,6 +211,9 @@ export default {
     if (this.formId) {
       this.id = this.formId;
       await this.fetchForm(this.id);
+    } else {
+      // default to TEAM on the form Settings when starting new
+      this.userType = this.ID_MODE.TEAM;
     }
     this.loading = false;
   },


app/frontend/src/components/designer/FormSettings.vue
```diff
@@ -211,6 +211,9 @@ export default {
     if (this.formId) {
       this.id = this.formId;
       await this.fetchForm(this.id);
+    } else {
+      // default to TEAM on the form Settings when starting new
+      this.userType = this.ID_MODE.TEAM;
```
jujaga, line 216: It may be worth calling directly to the imported variable instead to eliminate one function call in the stack:
```js
this.userType = IdentityMode.TEAM;
```

jujaga (approved):

bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-wxpbtr-tools.pathfinder.gov.bc.ca/dashboard?id=common-hosted-form-service-pr-108

app/frontend/src/components/designer/FormSettings.vue
loneil: Hmm doesn't like that or this.IdentityMode in created.

`TypeError: _utils_constants__WEBPACK_IMPORTED_MODULE_9__.IdentityMode.ID_MODE is undefined
    _callee$ FormSettings.vue:216
    Babel 7`

PR #117, opened 2020-12-16
Feature/admin stuff

loneil:
# Description
https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-1795

All the structure for the Admin side of the application.
Shows the tables to display forms and users in the system, can navigate into each one to get more info.
Pages to actually show the specific form/user info are not fleshed out but show the raw response.

Think this is enough for this ticket for now and can flesh out those exact UIs as needed in the future. Should get this reviewed and in to ensure structure/security and requirements are good for now.

## Types of changes

## Checklist

## Further comments

Added new /admin routing, views, and components on the FE.
Added a separated out admin store to section off administrative data and functions from main form data. Only register this module when in the admin section of the app.

Admin API endpoints expanded on.

More unit tests to come...


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-wxpbtr-tools.pathfinder.gov.bc.ca/dashboard?id=common-hosted-form-service-pr-117

app/frontend/src/store/modules/admin.js
```diff
@@ -0,0 +1,99 @@
...
+    async getUsers({ commit, dispatch }) {
+      try {
+        // Get all users
+        commit('SET_USERLIST', []);
+        const response = await adminService.listUsers();
+        commit('SET_USERLIST', response.data);
+      } catch (error) {
+        dispatch('notifications/addNotification', {
+          message:
+            'An error occurred while fetching forms.',
```
TimCsaky, line 79: ..while fetching users

TimCsaky:
this is looking awesome. looks like a great structure/design.

app/frontend/src/router/index.js
```diff
@@ -38,6 +38,33 @@ export default function getRouter(basePath = '/') {
...
       },
+      {
+        path: '/admin',
+        component: () => import(/* webpackChunkName: "admin" */ '@/views/Admin.vue'),
+        children: [
+          {
+            path: '',
+            name: 'Admin',
+            component: () => import(/* webpackChunkName: "adminroot" */ '@/views/admin/Root.vue')
+          },
```
jujaga, line 49: Just thinking about how this page is behaving atm - on the root admin page, we have a tab system where you can look at 3 subpanels - I'm half considering whether we should have those nested subpanels be explicit routes too (when you visit one of the later admin pages, then hit back, it always goes back to first tab instead of remembering what state it was previously). This could be dealt with via route state, or we could shove the tab state into vuex under admin tree or something. Let's talk about this some more first.

app/frontend/src/store/modules/admin.js
```diff
@@ -0,0 +1,99 @@
+import { adminService } from '@/services';
+
+/**
+ * Admin Module
+ */
+export default {
```
jujaga, line 6: I'm half wondering if we need this new admin vuex module tree - feels like to me we can still do most operations directly in the components without requiring centralized state management for this. This is closely tied to the existing forms module already, so this potentially risks state duplication in the overall store.

app/frontend/src/views/Admin.vue
```diff
@@ -0,0 +1,41 @@
+<template>
+  <BaseSecure :admin="true">
```
jujaga, line 2: I'm "pretty" sure if you can just do the following and it'll still pass in admin as true (no v-bind required):
```html
<BaseSecure admin>
```

app/frontend/src/views/Admin.vue
```diff
@@ -0,0 +1,41 @@
...
+    }
+    if (this.isAdmin) {
+      this.$store.registerModule('admin', admin);
+    }
+  },
+};
+</script>
+
+<style scoped>
+.component-fade-enter-active,
```
jujaga, line 33: We might need to make these transition things get pushed up to top-level scss - it's starting to show up in multiple places now.

app/src/forms/admin/service.js
```diff
@@ -27,7 +51,7 @@ const service = {
       // grab all users that have roles on this form...
       const users = [...new Set(accessItems.filter(x => x.formId === formId && x.roles.length))];
       const form = accessItems.find(x => x.formId === formId);
-      const userRoles = users.map(x => { return {userId: x.userId, username: x.username, roles: x.roles}; });
+      const userRoles = users.map(x => { return { userId: x.userId, username: x.username, roles: x.roles }; });
```
jujaga, line 54: The map callback function can be more cleanly defined using a closure to avoid needing a return statement (recommended for direct object returns or one liners):
```js
const userRoles = users.map(x => ({ userId: x.userId, username: x.username, roles: x.roles }));
```
If you really wanted to be fancy... could also use Object Destructuring and Property Shorthand:
```js
const userRoles = users.map(({ userId, username, roles }) => ({ userId, username, roles })(object));
```
Ref: https://stackoverflow.com/a/39333479

For readability's sake though, I'd suggest the simple closure option over the destructuring one.

jujaga (commented):
Looks pretty good. Let's chat a bit further with the routes and vuex proposals.

app/frontend/src/router/index.js
loneil: Changing tab to be remembered in state in next PR as discussed.

app/frontend/src/views/Admin.vue
loneil: Yeah will look, have a todo with table styles in the same way

app/src/forms/admin/service.js
loneil: Yeah first just looks more readable to me

jujaga (approved):

PR #127, opened 2021-01-05
Hide / show 'file upload' component depending on form user type

TimCsaky:
deployed here:
https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-127/

# Description

ticket [https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-1838](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-1838)

- force a re-render of the formBuilder component (with injected options) when the form userType (public, idir, team, etc) is changed.

- add file upload component to for builder in 'advanced' mode
It sounded like we will need to re-visit UX issues when switching between basic and advanced. 

## Checklist

### DIFF
--- a/app/frontend/src/components/designer/FormDesigner.vue
+++ b/app/frontend/src/components/designer/FormDesigner.vue
@@ -439,11 +439,11 @@ export default {
     }
   },
   watch: {
-    // if form userType (public, idir, team, etc) changes, re-render the form builder
-    userType() {
+    advancedForm() {
       this.reRenderFormIo += 1;
     },
-    advancedForm() {
+    // if form userType (public, idir, team, etc) changes, re-render the form builder
+    userType() {
       this.reRenderFormIo += 1;
     },
   },
--- a/app/frontend/src/components/designer/FormDesigner.vue
+++ b/app/frontend/src/components/designer/FormDesigner.vue
@@ -260,6 +260,12 @@ export default {
         return {
           builder: {
             premium: false,
+            basic: {
+              components: {
+                // add 'simplefile' file upload component to formBuilder in advanced mode
+                simplefile: this.userType !== this.ID_MODE.PUBLIC
+              }
+            }
           },
         };
       }
@@ -433,6 +439,10 @@ export default {
     }
   },
   watch: {
+    // if form userType (public, idir, team, etc) changes, re-render the form builder
+    userType() {
+      this.reRenderFormIo += 1;
+    },
     advancedForm() {
       this.reRenderFormIo += 1;
     },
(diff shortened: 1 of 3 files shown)


app/frontend/src/lib/components/SimpleFile/Component.js
```diff
@@ -36,7 +36,7 @@ import editForm from './Component.form';
 import { Constants } from '../Common/Constants';
 var uniqueName = Utils.uniqueName;
 var ID = 'simplefile';
-var DISPLAY = 'Upload';
+var DISPLAY = 'File Upload';
```
jujaga, line 39: We should not be directly editing the lib/components files ad this is directly generated by the Typescript compiler from the top-level components folder. It should be changed there instead and then recompiled... but there's other build mismatch issues associated with that atm.

app/frontend/src/components/designer/FormDesigner.vue
```diff
@@ -433,6 +439,10 @@ export default {
     }
   },
   watch: {
+    // if form userType (public, idir, team, etc) changes, re-render the form builder
+    userType() {
+      this.reRenderFormIo += 1;
+    },
```
jujaga, line 445: Alphabetical ordering - this should come after advancedForm.

bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-127

app/frontend/src/components/designer/FormDesigner.vue
```diff
@@ -260,6 +260,12 @@ export default {
         return {
           builder: {
             premium: false,
+            basic: {
+              components: {
+                // add 'simplefile' file upload component to formBuilder in advanced mode
+                simplefile: this.userType !== this.ID_MODE.PUBLIC
+              }
```
jujaga, line 267: Should we be doing this? I know this fileUpload isn't exactly just a "simple component", but this also asks questions about how unified our custom components vs standard components should be -> mixing both could cause quite a bit of confusion in the long run for our users.

loneil: What is this doing? We don't add the orgbook manually for example but it's in both basic and advanced.

TimCsaky: Agreed. i guess the file upload component should be re-named from 'simplefile' to 'fileupload'
the same way the OrgBook component is just called 'orgbook'
seems out of scope for this ticket.

app/frontend/src/lib/components/SimpleFile/Component.js
loneil: Yeah would figure out the rebuild process soon for stuff like this. Again maybe move out of scope for this fix.

TimCsaky:
deployed here:
https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-127/

jujaga (approved):

PR #137, opened 2021-02-04
Unused form fetch in settings

loneil:
# Description
https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-1915
Settings was not displaying the IDP correctly, always showing TEAM.

Found this was because the form fetch condition was not actually being called in this component, and was not needed since the form is in state already from the parent. Refactored a little to fix this.

## Types of changes

## Checklist

### DIFF
--- a/app/frontend/src/store/modules/form.js
+++ b/app/frontend/src/store/modules/form.js
@@ -1,6 +1,6 @@
 import { getField, updateField } from 'vuex-map-fields';
 
-import { NotificationTypes } from '@/utils/constants';
+import { IdentityMode, NotificationTypes } from '@/utils/constants';
 import { formService, rbacService } from '@/services';
 import { generateIdps, parseIdps } from '@/utils/transformUtils';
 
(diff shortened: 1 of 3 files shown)


bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-137

app/frontend/src/store/modules/form.js
```diff
@@ -13,7 +13,7 @@ const genInitialForm = () => ({
   sendSubRecieviedEmail: false,
   showSubmissionConfirmation: true,
   submissionReceivedEmails: [],
-  userType: ''
+  userType: IdentityMode.TEAM
```
jujaga, line 16: I can't remember but didn't we originally have this set to Team by default, then changed it to be empty by default so that it wouldn't populate incorrectly or something?

loneil: Yes but looks like the refactor (the same one that lead to not needing the fetch in FormSettings) has fixed that. I regression tested creating forms, modifying existing settings, and permissions on form load and it's all good with the default set here now.

jujaga (approved):
Awesome work. LGTM. 👍

TimCsaky (approved):

PR #156, opened 2021-03-24
Unit tests for middleware and text change

loneil:
# Description
Unit tests for permission middleware.

Help text on status

## Types of changes

## Checklist

### DIFF
--- a/app/frontend/src/components/forms/submission/StatusPanel.vue
+++ b/app/frontend/src/components/forms/submission/StatusPanel.vue
@@ -30,7 +30,21 @@
             <div v-show="statusFields">
               <div v-if="showAsignee">
                 <!-- {{ formReviewers }} -->
-                <label>Assign To</label>
+                <label>
+                  Assign To
+                  <v-tooltip bottom>
+                    <template #activator="{ on, attrs }">
+                      <v-icon color="primary" v-bind="attrs" v-on="on">
+                        help_outline
+                      </v-icon>
+                    </template>
+                    <span>
+                      Submissions can be assigned to Form Reviewers. <br />
+                      To add more team members as Form Reviewers, go to the
+                      Manage page for this form.
+                    </span>
+                  </v-tooltip>
+                </label>
                 <v-autocomplete
                   v-model="assignee"
                   clearable
@@ -38,6 +52,7 @@
                   :filter="autoCompleteFilter"
                   :items="formReviewers"
                   :loading="loading"
+                  no-data-text="No Form Reviewers found with search. Add Form Reviewers on the Manage page."
                   outlined
                   return-object
                   :rules="[(v) => !!v || 'Assignee is required']"
@@ -290,7 +305,7 @@ export default {
               submissionId: this.submissionId,
               submissionStatusId: submissionStatusId,
               note: this.note,
-              userId: user.data.id
+              userId: user.data.id,
             };
             const response = await formService.addNote(
               this.submissionId,
--- a/app/tests/unit/forms/auth/authService.spec.js
+++ b/app/tests/unit/forms/auth/authService.spec.js
@@ -1,0 +1,81 @@
+const service = require('../../../../src/forms/auth/service');
+
+afterEach(() => {
+  jest.clearAllMocks();
+});
+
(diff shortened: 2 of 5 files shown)


app/frontend/src/components/forms/submission/StatusPanel.vue
```diff
@@ -30,14 +30,29 @@
...
+                  <v-tooltip bottom>
+                    <template #activator="{ on, attrs }">
+                      <v-icon color="primary" v-bind="attrs" v-on="on">
+                        help_outline
+                      </v-icon>
+                    </template>
+                    <span>
+                      Submissions can be assigned to Form Reviewers. <br />
+                      To add more team members as Form Reviewers, go to the
+                      Manage page for this form.
```
jujaga, line 44: I wonder if it would be a good idea to make the "Manage page" part of text hyperlinkable?

bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-156

app/tests/unit/forms/auth/authService.spec.js
```diff
@@ -0,0 +1,81 @@
...
+      labels: 4,
+      idps: 5,
+      active: 6,
+      formVersionId: 7,
+      version: 8,
+      published: 9,
+      versionUpdatedAt: 10,
+      roles: 11,
+      permissions: 12
+    });
```
jujaga, line 53: It looks like those two objects are identical, but declared explicitly twice. You could do something like the following to make it more concise:
```js
const form = {
  formId: 1,
  formName: 2,
  ...
  permissions: 12
};
const result = service.formAccessToForm(form);
expect(result).toEqual(form);
```

app/tests/unit/forms/auth/authService.spec.js
```diff
@@ -0,0 +1,81 @@
...
+    const token = 'token';
+    const result = await service.login(token, params);
+    expect(service.parseToken).toHaveBeenCalledTimes(1);
+    expect(service.parseToken).toHaveBeenCalledWith(token);
+    expect(service.getUserId).toHaveBeenCalledTimes(1);
+    expect(service.getUserId).toHaveBeenCalledWith('userInf');
+    expect(service.getUserForms).toHaveBeenCalledTimes(2);
+    expect(service.getUserForms).toHaveBeenCalledWith({ user: 'me' }, { p: 1, active: false });
+    expect(result).toBeTruthy();
+    expect(result).toEqual({ user: 'me', forms: [{ formID: 1 }, { formId: 2 }], deletedForms: [{ formID: 1 }, { formId: 2 }] });
```
jujaga, line 73: This object looks complex enough that it may be better to space it out over multiple lines instead for readability. Same with Line 79 on this file below.
Also the fact that it's the same object for both this line and 79, you could define it as a const on Line 58 so that it's available for use for both tests under the `login` suite and only defined once.

app/tests/unit/forms/auth/authService.spec.js
loneil: That'll be a reference equality though right? And the point of the test is to see that the contents of the 2 objects are the same and not that the reference is just being returned. (or does `toEqual` de-reference?)

app/frontend/src/components/forms/submission/StatusPanel.vue
loneil: Problem is it's possible that you can be here (you're a reviewer) but don't have access to the Manage page at all...
I think this'll need some iteration after getting users on it but yeah a link would be good, but would have to do the same role detection.

app/tests/unit/forms/auth/authService.spec.js
loneil: NM, answered my own question. So I think you're right it can be shortedned
https://stackoverflow.com/questions/45195025/what-is-the-difference-between-tobe-and-toequal-in-jest

app/tests/unit/forms/auth/authService.spec.js
loneil: Fixed.

jujaga (approved):

PR #160, opened 2021-04-15
New submission permission middleware

loneil:
# Description
New middleware to check SUBMISSION permissions on route. This is to prep for all our submission edit workflow work and fixes the bug in https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-1986.

Include logging (kube probe) change too.

## Types of changes

## Checklist

### DIFF
--- a/app/src/forms/auth/middleware/userAccess.js
+++ b/app/src/forms/auth/middleware/userAccess.js
@@ -74,6 +74,59 @@ const hasFormPermissions = (permissions) => {
   };
 };
 
+const hasSubmissionPermissions = async (permissions) => {
+  return async (req, res, next) => {
+
+    if (!req.currentUser) {
+      // cannot find the currentUser... guess we don't have access... FAIL!
+      return next(new Problem(401, { detail: 'Current user not found on request.' }));
+    }
+    // Get the provided submission ID whether in a param or query (precedence to param)
+    const submissionId = req.params.submissionId || req.query.submissionId;
+    if (!submissionId) {
+      // No submission provided to this route that secures based on form... that's a problem!
+      return next(new Problem(401, { detail: 'Submission Id not found on request.' }));
+    }
+
+    // Get the submission results so we know what form this submission is for
+    const submissionForm = await service.getSubmissionForm(submissionId);
+
+    // Does the user have permissions for this submission due to their FORM permissions
+    let formFromCurrentUser = req.currentUser.forms.find(f => f.formId === submissionForm.form.id);
+    if (formFromCurrentUser) {
+      if (!Array.isArray(permissions)) {
+        permissions = [permissions];
+      }
+      // Do they have the submission permissions being requested on this FORM
+      const intersection = permissions.filter(p => {
+        return formFromCurrentUser.permissions.includes(p);
+      });
+      if (intersection.length == permissions.length) {
+        return next();
+      }
+    }
+
+    const isDeleted = submissionForm.submission.deleted;
+    const isDraft = submissionForm.submission.draft;
+    const publicAllowed = submissionForm.form.identityProviders.find(p => p.code === 'public') !== undefined;
+    const idpAllowed = submissionForm.form.identityProviders.find(p => p.code === currentUser.idp) !== undefined;
+
+    // check against the public and user's identity provider permissions...
+    if (!isDraft && !isDeleted) {
+      if (publicAllowed || idpAllowed) return next();
+    }
+
+    // check against the submission level permissions assigned to the user...
+    const submissionPermission = service.checkSubmissionPermission();
+    if (submissionPermission) return next();
+
+    // no access to this submission...
+    return next(new Problem(401, { detail: 'You do not have access to this submission.' }));
+  };
+};
+
+
 
 module.exports.currentUser = currentUser;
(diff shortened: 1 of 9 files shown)


bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-160

bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-160

app/src/forms/auth/middleware/userAccess.js
```diff
@@ -74,6 +76,60 @@ const hasFormPermissions = (permissions) => {
...
+      // No submission provided to this route that secures based on form... that's a problem!
+      return next(new Problem(401, { detail: 'Submission Id not found on request.' }));
+    }
+
+    // Get the submission results so we know what form this submission is for
+    const submissionForm = await service.getSubmissionForm(submissionId);
+    const isDeleted = submissionForm.submission.deleted;
+
+    // Deleted submissions are inaccessible
+    if (isDeleted) {
```
jujaga, line 97: Doesn't seem like isDeleted is used anywhere else. Could streamline this into
```js
if (submissionForm.submission.deleted) {
```

app/src/forms/auth/middleware/userAccess.js
```diff
@@ -74,6 +76,60 @@ const hasFormPermissions = (permissions) => {
...
+    // Does the user have permissions for this submission due to their FORM permissions
+    if (req.currentUser) {
+      let formFromCurrentUser = req.currentUser.forms.find(f => f.formId === submissionForm.form.id);
+      if (formFromCurrentUser) {
+
+        // Do they have the submission permissions being requested on this FORM
+        const intersection = permissions.filter(p => {
+          return formFromCurrentUser.permissions.includes(p);
+        });
+        if (intersection.length == permissions.length) {
```
jujaga, line 116: ===

app/src/forms/auth/service.js
```diff
@@ -212,8 +214,41 @@ const service = {
...
+  // Submission Data
+  // -------------------------------------------------------------------------------------------------------------
+  checkSubmissionPermission: async (currentUser, submissionId, permissions) => {
+    if (currentUser.public) return false;
+    return FormSubmissionUserPermissions.query()
+      .modify('filterSubmissionId', submissionId)
+      .modify('filterUserId', currentUser.id)
+      .modify('filterByPermissions', permissions)
+      .first();
+  },
```
jujaga, line 230: Is this query missing a `throwIfNotFound()` in the chain?

app/src/forms/form/service.js
```diff
@@ -281,26 +281,6 @@ const service = {
     }
   },
 
-  updateSubmission: async (formSubmissionId, data, currentUser) => {
```
jujaga, line 284: I'm not sure if we should be removing this function yet? We'll likely be using this in the future when you end up editing a submission (submitter and reviewer back and forth interaction). It'd just be missing the logic to append to a log table (doesn't exist yet obviously).

app/src/forms/submission/routes.js
```diff
@@ -2,34 +2,36 @@ const routes = require('express').Router();
 
 const controller = require('./controller');
 const currentUser = require('../auth/middleware/userAccess').currentUser;
+const hasSubmissionPermissions = require('../auth/middleware/userAccess').hasSubmissionPermissions;
```
jujaga, line 5: If we change the userAccess middleware file down the line to module.export an object with those functions as attributes, we can then do something nicer like the following:
```js
const { currentUser, hasSubmissionPermissions } = require('../auth/middleware/userAccess');
```

app/src/forms/submission/service.js
```diff
@@ -38,60 +35,15 @@ const service = {
...
-
-    // check against the submission level permissions assigned to the user...
-    const submissionPermission = await checkSubmissionPermission();
-    if (submissionPermission) return result;
-
-    // no access to this submission...
-
-    throw new Problem(401, 'You do not have access to this submission.');
+  read: async (formSubmissionId) => {
+    return await service._fetchSubmissionData(formSubmissionId);
```
jujaga, line 39: If we're expecting a promise from _fetchSubmissionData and read function is already async, we could make this function a synchronous passthrough instead and avoid unnecessary async/await tackons:
```js
  read: (formSubmissionId) => service._fetchSubmissionData(formSubmissionId),
```

app/src/forms/auth/middleware/userAccess.js
loneil: Yeah true

app/src/forms/form/service.js
loneil: Yeah it's unused in the form service but definitely will be used in the submission service (duplicated there so removed this one)

app/src/forms/auth/service.js
loneil: No this one is supposed to return which permission intersections you have for the supplied id, user, permission list. So the array of permissions if so, or falsey if nothing.

jujaga (approved):

TimCsaky (approved):
this is great. seems like it will make applying permissions a lot easier

PR #167, opened 2021-04-29
Add extra local navigation links on the UI for a form

TimCsaky:
# Description

supports ticket: https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-1629

- added 'view all submissions' link on individual submission view page
- added 'manage form' link on 'submissions table' page
- added 'view all submissions' link on 'manage form' page
- tried a different icon for 'submissions'
- removed default SUBMISSION_DELETE permission for submitter

this feels a bit half-baked. not sure i'm checking permissions the best way.
there's a bit of code duplication. did consider having a 'navigation' component. that could be used something like this:
<FormNavigation links="['manage_form', 'manage_team', export_submissions'] />
depends how much we want to restructure stuff.

## Types of changes

## Checklist

### DIFF
--- a/app/frontend/src/components/forms/FormSubmission.vue
+++ b/app/frontend/src/components/forms/FormSubmission.vue
@@ -19,7 +19,7 @@
         </v-col>
         <v-spacer />
         <v-col class="text-sm-right" cols="12" sm="6">
-          <span v-if="canViewAllSubmissions()">
+          <span>
             <v-tooltip bottom>
               <template #activator="{ on, attrs }">
                 <router-link
@@ -33,7 +33,6 @@
                     v-on="on"
                   >
                     <v-icon class="mr-1">content_copy</v-icon>
-                    <span>Submissions</span>
                   </v-btn>
                 </router-link>
               </template>
@@ -132,7 +131,6 @@
 
 <script>
 import { mapActions, mapGetters } from 'vuex';
-import { FormPermissions } from '@/utils/constants';
 
 import AuditHistory from '@/components/forms/submission/AuditHistory.vue';
 import DeleteSubmission from '@/components/forms/submission/DeleteSubmission.vue';
@@ -169,20 +167,12 @@ export default {
       this.submissionReadOnly = !editing;
       this.reRenderSubmission += 1;
     },
-    canViewAllSubmissions() {
-      const perms = [
-        FormPermissions.SUBMISSION_READ,
-        FormPermissions.SUBMISSION_UPDATE,
-      ];
-      return this.permissions.some((p) => perms.includes(p));
-    },
   },
   async mounted() {
     await this.fetchSubmission({ submissionId: this.submissionId });
-    this.loading = false;
-
     // get current user's permissions on associated form
     await this.getFormPermissionsForUser(this.form.id);
+    this.loading = false;
   },
 };
 </script>
--- a/app/frontend/src/components/forms/FormSubmission.vue
+++ b/app/frontend/src/components/forms/FormSubmission.vue
@@ -28,11 +28,11 @@
                   <v-btn
                     class="mx-1"
                     color="primary"
-                    text
+                    icon
                     v-bind="attrs"
                     v-on="on"
                   >
-                    <v-icon class="mr-1">content_copy</v-icon>
+                    <v-icon class="mr-1">view_list</v-icon>
                   </v-btn>
                 </router-link>
               </template>
--- a/app/frontend/src/components/forms/FormSubmission.vue
+++ b/app/frontend/src/components/forms/FormSubmission.vue
@@ -1,5 +1,5 @@
 <template>
-  <div>
+  <div class="mt-5">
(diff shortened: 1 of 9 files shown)


bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-167

app/frontend/src/components/forms/FormSubmission.vue
```diff
@@ -19,6 +19,28 @@
...
+                >
+                  <v-btn
+                    class="mx-1"
+                    color="primary"
+                    text
+                    v-bind="attrs"
+                    v-on="on"
+                  >
+                    <v-icon class="mr-1">content_copy</v-icon>
+                    <span>Submissions</span>
```
jujaga, line 36: This button should be an icon only - it should not have text built-into the button as that goes against our top-right button navigation paradigm.

app/frontend/src/components/forms/FormSubmission.vue
```diff
@@ -87,16 +110,27 @@ export default {
...
     refreshNotes() {
       this.$refs.notesPanel.getNotes();
     },
+
+    canViewAllSubmissions() {
+      const perms = [
+        FormPermissions.SUBMISSION_READ,
+        FormPermissions.SUBMISSION_UPDATE,
+      ];
+      return this.permissions.some((p) => perms.includes(p));
```
jujaga, line 125: Not sure if we should be blocking navigation from the view submission page to the submissions table. At this time, the /form/view page requires you to be authenticated according to the route, so public users won't be able to look at this page anyways and thus makes this check not relevant. However, if we do plan on making this view not authentication forced, then this permission check would be useful as a submitter user (non-staff) would never need to see this button nor know about it.

app/frontend/src/components/forms/FormSubmission.vue
```diff
@@ -87,16 +110,27 @@ export default {
...
+    },
   },
   async mounted() {
     await this.fetchSubmission({ submissionId: this.submissionId });
     this.loading = false;
+
+    // get current user's permissions on associated form
+    await this.getFormPermissionsForUser(this.form.id);
```
jujaga, line 133: If you're checking for permissions, this MUST be done before loading is set to false. Loading should ONLY flip to false when all async XHR requests have completed.

app/frontend/src/components/forms/SubmissionsTable.vue
```diff
@@ -140,6 +176,12 @@ export default {
       }
     },
   },
+
+  created(){
+    // Get the permissions for this form
+    this.getFormPermissionsForUser(this.formId);
+  },
+
```
jujaga, line 184: Why do we need to do getFormPermissionsForUser in the created lifecycle? It is sufficient enough to do this as part of the mounted lifecycle.

jujaga (changes requested):

app/frontend/src/components/forms/FormSubmission.vue
loneil: If you're viewing a submission you're a reviewer so there's no way to get into that without having the ability to view submissions in the current model. So yeah being able to go from a submission to the submissions table is kind of a given.

app/frontend/src/components/forms/SubmissionsTable.vue
```diff
@@ -113,7 +137,19 @@ export default {
     },
   },
   methods: {
-    ...mapActions('form', ['fetchForm', 'fetchSubmissions']),
+    ...mapActions('form', ['fetchForm', 'fetchSubmissions', 'getFormPermissionsForUser']),
+
+    checkFormManage() {
```
loneil, line 142: Lets use the checkFormManage in permissionUtils so this list of permissions isn't duplicated accross the code (1 place to change it)

app/frontend/src/components/forms/SubmissionsTable.vue
TimCsaky: i forget exactly why i was doing this on the created() event. I've put it back in mounted() and it seems to be working

app/frontend/src/components/forms/SubmissionsTable.vue
TimCsaky: checkManageForm() takes a simplified form object:
```
currentVersionId: ​​
description: ​​
id: ​​
idps: ​​
name: ​​
permissions: 
```
so I can build that object, but the problem I'm having is that this.permissions arent committed and available in the store until getFormPermissionsForUser(this.formId) has completed.
I'll have to take another look

loneil: Ok i would just move on and just make that array a constant so it's not duplicated here and permissionUtils

TimCsaky: ok. i tried doing something else but have gone with this solution

jujaga (approved):

PR #185, opened 2021-06-14
re-order submission data headers in csv export to match form design

TimCsaky:
# Description

[ticket 2114](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-2114)
Order of JSON and CSV columns does not match the top-down order of the form

Submission fields in the export are ordered  as they exist in the database (key length and then alphabetical)
We can re-order the columns in the CSV export with the jsonexport 'headers' argument.
.. passing in the meta columns, followed by the fields as they appear in the form schema.
submission fields that arent in the most recent form schema are appended to columns on the end.
This can be done just for the CSV export.

## Types of changes

## Checklist

### DIFF
--- a/app/src/forms/form/exportService.js
+++ b/app/src/forms/form/exportService.js
@@ -136,6 +136,7 @@ const service = {
   _readLatestFormSchema: async (formId) => {
     return FormVersion.query()
       .select('schema')
+      .withGraphFetched('submissions')
       .where('formId', formId)
       .modify('orderVersionDescending')
       .first().then((row) => row.schema);
--- a/app/src/forms/form/exportService.js
+++ b/app/src/forms/form/exportService.js
@@ -41,6 +41,28 @@ const EXPORT_FORMATS = Object.freeze({
 });
 
 const service = {
+
+  _buildCsvHeaders: async (form, data) => {
+    // -- get column order to match field order in form design
+    // object key order is not preserved when submission JSON is saved to jsonb field type in postgres.
+    // get correctly ordered field names (keys) from latest form version
+    const latestFormDesign = await service._readLatestFormSchema(form.id);
+
+    const fieldNames = latestFormDesign.components
+      .filter(x => x.input === true)
+      .map(({key}) => key);
+
+    // get meta properties in 'form.<child.key>' string format
+    const metaKeys = Object.keys(data[0].form);
+    const metaHeaders = metaKeys.map(x => 'form.' + x);
+
+    // -- make other changes to headers here if required
+    // eg: use field labels as headers
+    // see: https://github.com/kaue/jsonexport
+
+    return metaHeaders.concat(fieldNames);
+  },
+
   _checkPermission: (currentUser, formId, permission, exportType) => {
     const form = currentUser.forms.find(x => x.formId === formId);
     if (!form || !form.permissions.includes(permission)) {
@@ -133,29 +155,7 @@ const service = {
     };
   },
 
-  _readLatestFormSchema: async (formId) => {
-    return FormVersion.query()
-      .select('schema')
-      .withGraphFetched('submissions')
-      .where('formId', formId)
-      .modify('orderVersionDescending')
-      .first().then((row) => row.schema);
-  },
-
   _formatSubmissionsCsv: async (form, data) => {
-
-    // -- get column order to match field order in form design
-    // object key order is not preserved when submission JSON is saved to jsonb field type in postgres.
-    // get correctly ordered field names (keys) from latest form version
-    let latestFormDesign = await service._readLatestFormSchema(form.id);
-    let fieldNames = latestFormDesign.components
-      .filter(x => x.input === true)
-      .map(({key}) => key);
-
-    // get meta properties in 'form.<child.key>' string format
-    let metaKeys = Object.keys(data[0].form);
-    let metaHeaders = metaKeys.map(x => 'form.' + x);
-
     try {
       const options = {
         fillGaps: true,
@@ -176,7 +176,7 @@ const service = {
         // }
 
         // show meta headers first, then field names from form schema, remaining fields in data have columns automatically appended
-        headers: metaHeaders.concat(fieldNames)
+        headers: await service._buildCsvHeaders(form, data)
       };
 
       const csv = await jsonexport(data, options);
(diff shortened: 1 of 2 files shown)


bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-185

app/src/forms/form/exportService.js
```diff
@@ -133,7 +133,29 @@ const service = {
     };
   },
 
+  _readLatestFormSchema: async (formId) => {
+    return FormVersion.query()
+      .select('schema')
+      .withGraphFetched('submissions')
+      .where('formId', formId)
+      .modify('orderVersionDescending')
+      .first().then((row) => row.schema);
```
jujaga, line 142: The then function should probably be on a separate line for readability.
You can probably safely remove the async from this function as it is not awaiting anywhere.

app/src/forms/form/exportService.js
```diff
@@ -133,7 +133,29 @@ const service = {
...
+  },
+
   _formatSubmissionsCsv: async (form, data) => {
+
```
jujaga, line 146: Extra whitespace line

app/src/forms/form/exportService.js
```diff
@@ -133,7 +133,29 @@ const service = {
...
+    // object key order is not preserved when submission JSON is saved to jsonb field type in postgres.
+    // get correctly ordered field names (keys) from latest form version
+    let latestFormDesign = await service._readLatestFormSchema(form.id);
+    let fieldNames = latestFormDesign.components
+      .filter(x => x.input === true)
+      .map(({key}) => key);
+
+    // get meta properties in 'form.<child.key>' string format
+    let metaKeys = Object.keys(data[0].form);
+    let metaHeaders = metaKeys.map(x => 'form.' + x);
```
jujaga, line 157: All 4 of these variables should be `const` as they are never reassigned to later in the block scope.

loneil, line 157: Can extract to unit tests and use that in options.headers

app/src/forms/form/exportService.js
```diff
@@ -133,7 +133,29 @@ const service = {
     };
   },
 
+  _readLatestFormSchema: async (formId) => {
+    return FormVersion.query()
+      .select('schema')
+      .withGraphFetched('submissions')
```
loneil, line 139: Does this need a join on submissions? Are we just getting the form version.
(actually important to note withGraphFetched isn't a join it's a second query that does a WHERE IN on the parent results, so big time increase if it's needed)

TimCsaky: ok. good to know.. i thought this might be unnecessary. The idea was to only find the latest form version that has submissions. but you're right.. this could be a big overhead with a lot of submissions. Guess we should remove this clause and just get latest form version, regardless of whether it has submissions

app/src/forms/form/exportService.js
```diff
@@ -152,6 +174,9 @@ const service = {
         //     return value;
```
MichaelChung123, line 174: Can we remove this comment?

app/src/forms/form/exportService.js
```diff
@@ -133,7 +133,29 @@ const service = {
     };
   },
 
+  _readLatestFormSchema: async (formId) => {
```
MichaelChung123, line 136: can remove async since there is no await for this function

MichaelChung123: Could we also remove the async call on the function above called _formatSubmissionsJson?

bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-185

app/src/forms/form/exportService.js
TimCsaky: good catch. i've removed it from the _formatSubmissionsJson function.

loneil (approved):
👍

MichaelChung123 (approved):
LGTM

app/tests/unit/forms/form/exportService.spec.js
```diff
@@ -0,0 +1,57 @@
...
+      },
+    ];
+
+    // mock latestFormSchema
+    exportService._readLatestFormSchema = jest.fn(() => { return formSchema; });
+
+    // build csv headers
+    const result = await exportService._buildCsvHeaders(form, submissionsData);
+
+    // expect headers array to have 3 element - [ 'form.confirmationId', 'firstName', 'lastName' ]
```
MichaelChung123, line 53: Should we put more expects to assert the values in the result array?

app/tests/unit/forms/form/exportService.spec.js
```diff
@@ -0,0 +1,57 @@
+const exportService = require('../../../../src/forms/form/exportService');
+
+describe('_buildCsvHeaders', () => {
+
+  it.only('should build correct csv headers', async () => {
```
jujaga, line 5: Only should not be here. Probably forgot to remove it before commiting.

app/tests/unit/forms/form/exportService.spec.js
```diff
@@ -0,0 +1,57 @@
...
+    ];
+
+    // mock latestFormSchema
+    exportService._readLatestFormSchema = jest.fn(() => { return formSchema; });
+
+    // build csv headers
+    const result = await exportService._buildCsvHeaders(form, submissionsData);
+
+    // expect headers array to have 3 element - [ 'form.confirmationId', 'firstName', 'lastName' ]
+    expect(result.length === 3).toBeTruthy();
```
jujaga, line 54: These tests should be a bit more verbose and cover more aspects of the result:
```js
expect(result.length).toHaveLength(3);
expect(result).toEqual(expect.arrayContaining([ 'form.confirmationId', 'firstName', 'lastName' ]));
expect(exportService._readLatestFormSchema).toHaveBeenCalledTimes(1);
expect(exportService._readLatestFormSchema).toHaveBeenCalledWith(form, submissionsData);
```
Also, we *should* be mockRestoring or mockReseting anything we mock out at the end of the function. While this test works because there's only one it block, if there were multiple, `_readLatestFormSchema ` would collide with the next test since it never got reset and cleaned up.

jujaga (changes requested):

app/tests/unit/forms/form/exportService.spec.js
TimCsaky: yes. probably. i'll update this now

jujaga (approved):

PR #187, opened 2021-06-15
Making Links Open in new Tab

MichaelChung123:
# Description
Edit the Custom Content Component in FormIO Making <a> tags open links in new tab instead of trying to redirect the current page.

## Types of changes
Added the "target='_blank'" attribute to all <a> tags.

## Checklist

### DIFF
--- a/components/src/components/SimpleContent/Component.ts
+++ b/components/src/components/SimpleContent/Component.ts
@@ -31,4 +31,17 @@ export default class Component extends (ParentComponent as any) {
             schema: Component.schema()
         };
     }
+
+    addTargetBlankToLinks = (html) => {
+      let result = '';
+      if(html.includes('<a href=')) {
+        result = html.replace(/<a href=/g, '<a target="_blank" href=');
+      }
+      return result;
+    };
+
+    constructor(component, options, data) {
+      super(component, options, data);
+      component.html = this.addTargetBlankToLinks(component.html);
+    }
 }
--- a/components/src/components/SimpleContent/Component.ts
+++ b/components/src/components/SimpleContent/Component.ts
@@ -32,20 +32,24 @@ export default class Component extends (ParentComponent as any) {
     };
   }
 
-  get defaultSchema() {
-    return Component.schema();
-  }
-
   addTargetBlankToLinks = (html) => {
-    let result = '';
-    if (html.includes('<a href=')) {
-      result = html.replace(/<a href=/g, '<a target="_blank" href=');
+    if (html.includes('<a')) {
+      html = html.replace(/<a(?![^>]+target)/g, '<a target="_blank"');
     }
-    return result;
+    return html;
   }
 
   attach(element) {
     this.loadRefs(element, { html: 'single' });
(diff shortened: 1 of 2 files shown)


bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-187

components/src/components/SimpleContent/Component.ts
```diff
@@ -9,26 +10,43 @@ const ID = 'simplecontent';
...
-            schema: Component.schema()
-        };
+  get defaultSchema() {
+    return Component.schema();
+  }
+
+  addTargetBlankToLinks = (html) => {
+    let result = '';
+    if (html.includes('<a href=')) {
+      result = html.replace(/<a href=/g, '<a target="_blank" href=');
```
jujaga, line 43: While this works in the case where the anchor does not already have a target blank in the element, it's possible to have a situation where if a user manually writes in anchor html that has other stuff between the a and href, you'll miss inserting a target blank into that DOM element. Not too sure if there's a way to parse the html DOM string and then search on that instead of doing it via regex, or if that would be less computationally expensive.

TimCsaky:
i noticed some unexpected behavior when testing.
Entering just plain text (paragraph) into the wysiwyg editor didn't show up.
Also, saving the component and then going back to edit didn't work.

components/src/components/SimpleContent/Component.ts
```diff
@@ -9,26 +9,43 @@ const ID = 'simplecontent';
...
     }
```
TimCsaky, line 43: i noticed some unexpected behavior when testing.
Entering just plain text (paragraph) into the wysiwyg editor didn't show up.
Also, saving the component and then going back to edit didn't work.

i think you need to return the original html in an else statement (or something with that effect)

TimCsaky (changes requested):

components/src/components/SimpleContent/Component.ts
MichaelChung123: After entering a anchor tags manually through the component's textfields, it looks like my code and FormIO is able to differentiate between rendered HTML vs the content of the component.

components/src/components/SimpleContent/Component.ts
MichaelChung123: I have changed the structure of my addTargetBlankToLinks to always return the html whether it has been edit or not. Confirmed the errors you were getting previously are fixed on local.

jujaga: I would recommend using the regex
```js
/<a(?![^>]+target=)/g
```
as it'd be more specific than just target. As unlikely as it may be, there could be a random situations like the following where it'd misbehave with your current regex:
```html
<a target-value="true" href="foobar">
<a href="https://www.target.com">
```

MichaelChung123: Added the new regex in latest commit

jujaga (approved):

loneil (approved):

TimCsaky (approved):

PR #191, opened 2021-06-16
Static Text Fix

MichaelChung123:
# Description
Fixing issue with PR 187. Issue caused with conditional rendering of Static Content.

## Types of changes
Moved the previous logic for the SimpleContent into a constructor block instead of the built in attach method.

## Checklist

## Further comments
Will be doing a QA screen in Test env to confirm it is safe before merging.

### DIFF
--- a/app/frontend/src/components/designer/FormViewer.vue
+++ b/app/frontend/src/components/designer/FormViewer.vue
@@ -129,6 +129,22 @@ export default {
         this.loadingSubmission = false;
       }
     },
+    // Attaches attributes to <a> Link tags to open in a new tab.
+    attachAttributesToLinks() {
+      let simpleContentComponents = formioUtils.searchComponents(this.formSchema.components, {
+        type: 'simplecontent'
+      });
+      let advancedContent = formioUtils.searchComponents(this.formSchema.components,{
+        type: 'content'
+      });
+      console.log('simpleContentComponents: ', simpleContentComponents);
+      const combinedLinks = [...simpleContentComponents, ...advancedContent];
+      combinedLinks.map((component) => {
+        if (component.html && component.html.includes('<a ')) {
+          component.html = component.html.replace(/<a(?![^>]+target=)/g,'<a target="_blank" rel="noopener"');
+        }
+      });
+    },
     // Get the form definition/schema
     async getFormSchema() {
       try {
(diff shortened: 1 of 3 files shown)


bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-191

app/frontend/src/components/designer/FormViewer.vue
```diff
@@ -129,6 +129,22 @@ export default {
...
     },
+    // Attaches attributes to <a> Link tags to open in a new tab.
+    attachAttributesToLinks() {
+      let simpleContentComponents = formioUtils.searchComponents(this.formSchema.components, {
+        type: 'simplecontent'
+      });
+      let advancedContent = formioUtils.searchComponents(this.formSchema.components,{
+        type: 'content'
+      });
+      console.log('simpleContentComponents: ', simpleContentComponents);
```
jujaga, line 140: I think you forgot to remove this console log.

app/frontend/src/components/designer/FormViewer.vue
```diff
@@ -129,6 +129,22 @@ export default {
         this.loadingSubmission = false;
       }
     },
+    // Attaches attributes to <a> Link tags to open in a new tab.
+    attachAttributesToLinks() {
+      let simpleContentComponents = formioUtils.searchComponents(this.formSchema.components, {
```
jujaga, line 134: This should be a const - it's not being reassigned anywhere else in this function scope.

app/frontend/src/components/designer/FormViewer.vue
```diff
@@ -127,6 +129,22 @@ export default {
...
+        type: 'simplecontent'
+      });
+      let advancedContent = formioUtils.searchComponents(this.formSchema.components,{
```
jujaga, line 137: This should be a const - it's not being reassigned anywhere else in this function scope.
Also needs a space after the comma and before the open curly brace.

jujaga (commented):
Looks pretty good semantically. A future thought is that if we do end up needing to leverage formioUtils more heavily, it may be worth writing out these new functions as standalone JS modules so that it can be siloed away, imported and used as needed (but no need to do that yet since it'd only exist in this Vue component for the moment).

app/frontend/src/components/designer/FormViewer.vue
MichaelChung123: Moving this fix into a new PR-196
https://github.com/bcgov/common-hosted-form-service/pull/196

app/frontend/src/components/designer/FormViewer.vue
MichaelChung123: Moving this fix into a new PR-196
https://github.com/bcgov/common-hosted-form-service/pull/196

app/frontend/src/components/designer/FormViewer.vue
MichaelChung123: Moving this fix into a new PR-196
https://github.com/bcgov/common-hosted-form-service/pull/196

PR #195, opened 2021-06-22
order csv export columns

TimCsaky:
# Description

see: [ticket 2114](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-2114)

This is an extra fix to my last PR to get submissions export to match the order of the fields as they appear in the form schema.
`_buildCsvHeaders` references the list of fields created in '_readSchemaFields' function.
we need to go recursively into each component in case that component contains nested components.

The way the form schema is structured makes this process not very reliable. this is a best effort.

because any fields in the submission data that arent defined in headers are appended to the export we should be safe here.. 

-Jujaga -> Fixed some of the nasty recursion and edge cases

## Types of changes

## Checklist

### DIFF
--- a/app/src/forms/form/exportService.js
+++ b/app/src/forms/form/exportService.js
@@ -41,42 +41,54 @@ const EXPORT_FORMATS = Object.freeze({
 });
 
 const service = {
-
+  /**
+   * @function _readSchemaFields
+   * Returns a flattened, ordered array of relevant content field names with topology
+   * @param {Object} schema A form.io schema
+   * @returns {String[]} An array of strings
+   */
   _readSchemaFields: (schema) => {
+    /**
+     * @function findFields
+     * Recursively traverses the form.io schema to extract all relevant content field names
+     * @param {Object} obj A form.io schema or subset of it
+     * @returns {String[]} An array of strings
+     */
+    const findFields = (obj) => {
+      const fields = [];
 
-    const fields = [];
-
-    // function to recursively find all input field names
-    const findFields = function (obj) {
-      // if component is an input field showing in the form,
-      if (obj.input && !obj.hidden) {
+      // Handle current object level
+      if (obj.key && obj.input && !obj.hidden) {
+        // Add current field key if it is non-hidden input field
         fields.push(obj.key);
-
-        // if object is a checkbox group
-        if (obj.values && obj.type && obj.type.includes('checkbox')) {
-          // add checkbox fields now
-          obj.values.forEach(element => fields.push(obj.key + '.' + element.value));
+        // Look-ahead and add all sub-values as fields if type is simplecheckbox or checkbox
+        // TODO: there may be more cases where attribute is an object tree
+        if (obj.type && obj.type.includes('checkbox')) {
+          if (obj.values) obj.values.forEach(e => fields.push(`${obj.key}.${e.value}`));
         }
       }
 
-      // else not an input (eg: panel, column, etc)
-      else {
-        // for each property of the component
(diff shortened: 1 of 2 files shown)


bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-195

app/src/forms/form/exportService.js
```diff
@@ -42,15 +42,49 @@ const EXPORT_FORMATS = Object.freeze({
...
+            // repeat this function recursively
+            obj[prop].map(e => findFields(e));
+          }
+        });
+      }
+    };
+
+    schema.components.forEach(function (item) {
+      findFields(item);
+    });
```
jujaga, line 76: Use arrow notation for anonymous functions
```js
schema.components.forEach((item) => findFields(item));
```

app/src/forms/form/exportService.js
```diff
@@ -42,15 +42,49 @@ const EXPORT_FORMATS = Object.freeze({
...
+    const fields = [];
+
+    // function to recursively find all input field names
+    const findFields = (obj) => {
+      // if component is an input field showing in the form,
+      if (obj.input && !obj.hidden) {
+        fields.push(obj.key);
+
+        // if object is a checkbox group
+        if (obj.type && obj.type.includes('checkbox') && obj.values) {
```
jujaga, line 55: May be worth inverting conditional evaluation to short circuit more effectively:
```js
if (obj.values && obj.type && obj.type.includes('checkbox')) {
```

app/src/forms/form/exportService.js
```diff
@@ -42,15 +42,49 @@ const EXPORT_FORMATS = Object.freeze({
...
+          // add checkbox fields now
+          obj.values.forEach(element => fields.push(obj.key + '.' + element.value));
```
jujaga, line 57: The key seems "ok" for singly nested values, but I'm concerned this will not work for deeper nested situations. May need to think about how this recursion is actually working and its edge cases.

app/src/forms/form/exportService.js
```diff
@@ -42,15 +42,49 @@ const EXPORT_FORMATS = Object.freeze({
...
+        }
+      }
+
+      // else not an input (eg: panel, column, etc)
+      else {
+        // for each property of the component
+        Object.keys(obj).forEach(function (prop) {
```
jujaga, line 64: Use arrow notation for anonymous functions:
```js
Object.keys(obj).forEach((prop) => {
```

jujaga (commented):
I'm not sure if the findFields recursive function is designed correctly as it is pushing out the result as a side effect instead of returning the results of the children array that it discovers in its level.

app/src/forms/form/exportService.js
TimCsaky: can you explain your reasoning for that advice?

app/src/forms/form/exportService.js
```diff
@@ -42,15 +42,49 @@ const EXPORT_FORMATS = Object.freeze({
 
 const service = {
 
+  _readSchemaFields: async (schema) => {
```
MichaelChung123, line 45: Would it be fine to remove this async call? I dont believe this is making any await calls.

jujaga (approved):

bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-195

MichaelChung123 (approved):
LGTM

loneil (approved):

PR #197, opened 2021-06-23
Form Header Title Width Fix

MichaelChung123:
# Description
Letting the Form's Title header span the  entire width of it's container.

## Types of changes
Adding Vuetify bootstrap grid changes.

## Checklist

### DIFF
--- a/app/frontend/src/components/designer/FormDesigner.vue
+++ b/app/frontend/src/components/designer/FormDesigner.vue
@@ -1,12 +1,14 @@
 <template>
   <div>
     <v-row class="mt-6" no-gutters>
-      <v-col cols="12" sm="6">
+      <v-col cols="12" sm="6" order-md="1">
         <h1>Form Design</h1>
+      </v-col>
+      <v-col cols="12" sm="12" order-md="3">
         <h3 v-if="name">{{ name }}</h3>
       </v-col>
       <v-spacer />
-      <v-col class="text-sm-right" cols="12" sm="6">
+      <v-col class="text-sm-right" cols="12" sm="6" order-md="1">
         <v-tooltip bottom>
           <template #activator="{ on, attrs }">
             <v-btn
--- a/app/frontend/src/components/designer/FormDesigner.vue
+++ b/app/frontend/src/components/designer/FormDesigner.vue
@@ -2,16 +2,11 @@
   <div>
     <v-row class="mt-6" no-gutters>
       <!-- page title -->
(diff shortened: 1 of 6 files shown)


bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-197

app/frontend/src/components/designer/FormDesigner.vue
```diff
@@ -1,16 +1,21 @@
...
         <h1>Form Design</h1>
+      </v-col>
+      <!-- form name -->
+      <v-col cols="12" order-md="3" order-sm="2">
         <h3 v-if="name">{{ name }}</h3>
       </v-col>
       <v-spacer />
-      <v-col class="text-sm-right" cols="12" sm="6">
+      <!-- buttons -->
+      <v-col class="text-md-right text-sm-left" cols="12" md="6" order="4" order-md="2" order-sm="4">
```
loneil, line 14: Does this need to be split up into mutliple columns with responsive orders?
Can it just be 
| form design,  name, version (vertical)|-----------------------| buttons |
?

loneil: Ohhh no is this so the form title can run **under** the buttons if it's long? In that case this probably works.
Though I'm not sure on line 14 if you need both `order="4"` AND `order-sm="4"`? Can the `sm` one be removed because the basic 4 one runs from xs to md?

TimCsaky: thats what i thought.. but i found you needed both wen testing

jujaga (commented):
I feel like we should be forcing our action buttons to show up above the page title and form title. This would also affect the management page as well if you end up having a degenerate form name like the example below:

Capture

app/frontend/src/components/designer/FormDesigner.vue
loneil: That's weird, the spec says they're default and then breakpoint and **greater**

image

MichaelChung123: If we do it that way, the mobile ordering gets messed up. That's why we have to structure it in this fashion.

bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-197

loneil (approved):

jujaga (approved):

PR #201, opened 2021-06-29
Submissions

MichaelChung123:
# Description
Fixing Header alignment issues on Submission View page. Also fixed Header Icons showing up in Print View.

## Checklist

### DIFF
--- a/app/frontend/src/components/forms/FormSubmission.vue
+++ b/app/frontend/src/components/forms/FormSubmission.vue
@@ -19,7 +19,7 @@
           </p>
         </v-col>
         <!-- buttons -->
-        <v-col class="text-right text-sm-right" cols="12" sm="6" order="1" order-sm="2">
+        <v-col class="text-right text-sm-right d-print-none" cols="12" sm="6" order="1" order-sm="2">
           <span>
             <BasePrintButton />
           </span>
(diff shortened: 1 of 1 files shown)


bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-201

jujaga (approved):

app/frontend/src/components/forms/FormSubmission.vue
```diff
@@ -20,22 +35,7 @@
...
-            {{ formSubmission.createdAt | formatDateLong }}
-            <br />
-            <strong>Confirmation ID:</strong>
-            {{ formSubmission.confirmationId }}
-            <br />
-            <strong>Submitted By:</strong>
-            {{ formSubmission.createdBy }}
-            <br />
-          </p>
+          <DeleteSubmission class="d-print-none" @deleted="onDelete" :submissionId="submissionId" />
```
loneil, line 38: Does this need d-print class? It's on the containing column so the whole div containing all this would hide right?

jujaga: Good catch - yea that class may not be necessary at all due to inheritance.

MichaelChung123: good catch, removing this class on the component in new commit.

TimCsaky (approved):

PR #204, opened 2021-06-30
exclude 'deleted' and 'draft' submissions from export

TimCsaky:
# Description

[ticket 2113](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-2113)

## Types of changes

## Checklist

### DIFF
--- a/app/frontend/src/components/forms/ExportSubmissions.vue
+++ b/app/frontend/src/components/forms/ExportSubmissions.vue
@@ -168,11 +168,16 @@ export default {
               .utc()
               .format()
             : undefined;
+
+        // include deleted submissions in export. Default to false, but this could be set by user if implemented)
+        const showDeleted = false;
+
         const response = await formService.exportSubmissions(
           this.form.id,
           from,
           to,
-          this.exportFormat
+          this.exportFormat,
+          showDeleted
         );
         if (response && response.data) {
           const blob = new Blob([response.data], {
--- a/app/frontend/src/components/forms/ExportSubmissions.vue
+++ b/app/frontend/src/components/forms/ExportSubmissions.vue
@@ -172,12 +172,16 @@ export default {
         // include deleted submissions in export. Default to false, but this could be set by user if implemented)
         const showDeleted = false;
 
+        // include draft submissions in export. Default to false, but this could be set by user if implemented)
+        const showDrafts = false;
+
         const response = await formService.exportSubmissions(
           this.form.id,
           from,
           to,
           this.exportFormat,
-          showDeleted
+          showDeleted,
+          showDrafts
         );
         if (response && response.data) {
           const blob = new Blob([response.data], {
--- a/app/frontend/src/components/forms/ExportSubmissions.vue
+++ b/app/frontend/src/components/forms/ExportSubmissions.vue
@@ -169,19 +169,15 @@ export default {
               .format()
             : undefined;
 
-        // include deleted submissions in export. Default to false, but this could be set by user if implemented)
-        const showDeleted = false;
-
-        // include draft submissions in export. Default to false, but this could be set by user if implemented)
-        const showDrafts = false;
-
         const response = await formService.exportSubmissions(
           this.form.id,
-          from,
-          to,
           this.exportFormat,
-          showDeleted,
-          showDrafts
+          {
+            minDate: from,
+            maxDate: to,
+            // deleted: true,
+            // drafts: true
+          }
         );
         if (response && response.data) {
           const blob = new Blob([response.data], {
--- a/app/frontend/src/services/formService.js
+++ b/app/frontend/src/services/formService.js
@@ -261,23 +261,18 @@ export default {
   /**
    * @function exportSubmissions
    * Get the export file for a range of form submittions
-   * @param {string} minDate The form uuid
-   * @param {string} maxDate The form uuid
-   * @param {string} format The file format csv or json
-   * @param {string} showDeleted should deleted submissions be included in export? true or false
-   * @param {string} showDrafts should draft submissions be included in export? true or false
+   * @param {string} formId The form uuid
+   * @param {string} format The export file format csv or json
+   * @param {object} options options for the export (eg: minDate, maxDate, deleted, drafts)
    * @returns {Promise} An axios response
    */
-  exportSubmissions(formId, minDate, maxDate, format, showDeleted, showDrafts) {
+  exportSubmissions(formId, format, options = {}) {
     return appAxios().get(`${ApiRoutes.FORMS}/${formId}/export`,
       {
         params: {
           format: format,
           type: 'submissions',
-          minDate: minDate ? minDate : undefined,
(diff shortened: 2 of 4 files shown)


bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-204

app/src/forms/form/exportService.js
```diff
@@ -188,6 +188,7 @@ const service = {
     return SubmissionData.query()
       .column(service._submissionsColumns(form, params))
       .where('formId', form.id)
+      .where('deleted', false)
```
jujaga, line 191: This should not be hard coded IMHO - instead, this should come through the params argument so that it is configurable and can be overridden if required.

loneil: This would be good for data exports yeah, Then the FE can just pass in an active or deleted flag

TimCsaky: done. just saw your comment Luke.. i called it showDeleted.. let me know if you want that changed

app/src/forms/form/exportService.js
```diff
@@ -22,6 +22,11 @@ class SubmissionData extends Model {
           query.where('createdAt', '<=', maxDate);
         }
       },
+      filterDeleted(query, showDeleted) {
+        if(showDeleted == 'false') {
```
TimCsaky, line 26: i couldnt figure out why this is value seems to be a string

app/frontend/src/services/formService.js
```diff
@@ -264,16 +264,20 @@ export default {
...
           format: format,
           type: 'submissions',
           minDate: minDate ? minDate : undefined,
-          maxDate: maxDate ? maxDate : undefined
+          maxDate: maxDate ? maxDate : undefined,
+          showDeleted: showDeleted ? showDeleted : false,
+          showDrafts: showDrafts ? showDrafts : false
         },
         responseType: 'blob'
       }
```
jujaga, line 283: The exportSubmissions function is starting to become a bit hefty with function arguments. General ES conventions usually recommend to not have more than 3 or 4 arguments to a function call - any more than that and we'll want to start considering passing in options as an object argument instead of explicitly as separated arguments. So we could do something like the following instead:
```js
exportSubmissions(formId, format, options = {}) {
  return appAxios().get(`${ApiRoutes.FORMS}/${formId}/export`,
    {
      params: {
        format: format,
        type: 'submissions',
        ...options
      },
      responseType: 'blob'
    }
  );
},
```
This would give us the liberty of handling optionality of the minDate, maxDate, showDrafts and showDeleted attributes. If they're not defined in the object, destructuring by default will have them be undefined, and not show up in the network query request at all - a much cleaner approach.

app/src/forms/form/exportService.js
```diff
@@ -22,6 +22,16 @@ class SubmissionData extends Model {
...
+          query.where('deleted', false);
+        }
+      },
+      filterDrafts(query, showDrafts) {
+        if(showDrafts == 'false') {
+          query.where('draft', false);
+        }
+      },
```
jujaga, line 34: This can be simplified to be more concise and leverage truthiness checks:
```js
filterDeleted(query, showDeleted) {
  if(!showDeleted) query.where('deleted', false);
},
filterDrafts(query, showDrafts) {
  if(!showDrafts) query.where('draft', false);
},
```
If we explicitly check for "false" instead of falsiness, we may miss a few edge cases such as when they're undefined (as they're not explicitly "false").

app/frontend/src/components/forms/ExportSubmissions.vue
```diff
@@ -168,11 +168,20 @@ export default {
...
+
         const response = await formService.exportSubmissions(
           this.form.id,
           from,
           to,
-          this.exportFormat
+          this.exportFormat,
+          showDeleted,
+          showDrafts
         );
```
jujaga, line 185: Following up on the signature change proposal below, we can simplify this to the following:
```js
const response = await formService.exportSubmissions(
  this.form.id,
  this.exportFormat,
  {
    minDate: from,
    maxDate: to
    // showDeleted: undefined
    // showDrafts: undefined
  }
);
```
By having it as an options object as an argument, we can ensure that by default if a user calling this service function doesn't specify that option, it'll just default to false. As a bonus, this also makes the API call have less query parameters too simply by taking advantage of truthiness/falsiness JS behavior.

jujaga (commented):
Let's consider how we're expanding this frontend service call signature so that it's more extensible. I've given a proposal for this in the comments.

app/src/forms/form/exportService.js
loneil: We do this in a couple spots I think
```
      filterDraft(query, value) {
        if (value !== undefined) {
          query.where('draft', value);
        }
      },
      filterDeleted(query, value) {
        if (value !== undefined) {
          query.where('deleted', value);
        }
      },
 ```

But whatever works and is concise, we should look later at making these standard.

Also this model should go in the model area with the other ones (not sure why it was originally here) but that can go when we refactor all those to their own files.

File: app/frontend/src/components/forms/ExportSubmissions.vue [hunk 6]
```diff
@@ -168,11 +168,16 @@ export default {
...
         const response = await formService.exportSubmissions(
           this.form.id,
-          from,
-          to,
-          this.exportFormat
+          this.exportFormat,
+          {
+            minDate: from,
+            maxDate: to,
+            // deleted: true,
```
MichaelChung123, line 178: do we need these comments here for the final merge?

TimCsaky: not really. i left these comments in to show that they should be passed as true (boolean) if desired.
or not passed (or passed as undefined) if not desired.

jujaga: In this situation I think it's worth keeping the comments here as they'll provide breadcrumbs on how we can extend this call down the line to allow a user to choose these options.

app/src/forms/form/exportService.js
```diff
@@ -184,10 +193,12 @@ const service = {
 
   _getSubmissions: (form, params, currentUser) => {
     service._checkPermission(currentUser, form.id, Permissions.SUBMISSION_READ, EXPORT_TYPES.submissions);
-    // possible params for this export include minDate and maxDate (full timestamp dates).
+    // params for this export include minDate and maxDate (full timestamp dates).
     return SubmissionData.query()
       .column(service._submissionsColumns(form, params))
       .where('formId', form.id)
+      .modify('filterDeleted', params.deleted)
+      .modify('filterDrafts', params.drafts)
```
jujaga, line 201: These modify functions should be added in after filterCreatedAt to stay alphabetically consistent.

jujaga (approved):
Looks good. Minor alphabetical ordering tweak and it should be ready for merge. Tentatively approving it. 👍

loneil (approved):

PR #205, opened 2021-07-02
Draft link refactor

MichaelChung123:
# Description
Making links open in a new tab for Submission Draft View.

## Types of changes
Re-structured the function to attach attributes, and moved functionality into lifecycle method in the Component.

## Checklist

### DIFF
--- a/app/frontend/src/utils/transformUtils.js
+++ b/app/frontend/src/utils/transformUtils.js
@@ -57,7 +57,8 @@ export function attachAttributesToLinks(formSchemaComponents) {
     type: 'content'
   });
   const combinedLinks = [...simpleContentComponents, ...advancedContent];
-  combinedLinks.map((component) => {
+
+  combinedLinks.forEach((component) => {
     if (component.html && component.html.includes('<a ')) {
       component.html = component.html.replace(/<a(?![^>]+target=)/g,'<a target="_blank" rel="noopener"');
     }
--- a/app/frontend/src/utils/transformUtils.js
+++ b/app/frontend/src/utils/transformUtils.js
@@ -1,4 +1,5 @@
 import { IdentityMode } from '@/utils/constants';
+import formioUtils from 'formiojs/utils';
 
 //
 // Transformation Functions for converting form objects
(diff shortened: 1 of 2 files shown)


bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-205

app/frontend/src/utils/transformUtils.js
```diff
@@ -1,4 +1,5 @@
 import { IdentityMode } from '@/utils/constants';
+import formioUtils from 'formiojs/utils';
```
jujaga, line 2: Style: Normally we tend to import external libraries first before we import our internal codebase. Would be good to have this line be above Line 1.

app/frontend/src/utils/transformUtils.js
```diff
@@ -42,3 +43,24 @@ export function parseIdps(identityProviders) {
   }
   return result;
 }
+
+/**
+ * @function attachAttributesToLinks
+ * Attaches attributes to <a> Link tags to open in a new tab
+ * @param {Component[]} formSchemaComponents An array of Components
```
jujaga, line 50: IIRC the JSDoc doesn't support custom types like Component unless you define it somewhere else what that type is. I'd recommend changing this to just be `Object[]` and let the description handle describing what kind of object array it is.

app/frontend/src/utils/transformUtils.js
```diff
@@ -42,3 +43,24 @@ export function parseIdps(identityProviders) {
...
+ * @returns {void} Does not return anything. Simply transforms the Components using FormIO Utils.
```
jujaga, line 51: If there is no return value in a function, this line can be completely omitted.

jujaga (approved):
Looks good. Tentatively approving with some minor changes requested. 👍

TimCsaky:
looks good. 
i forget why this function needs to be called beforeUpdate.. 

also I kind of wonder if we should have a 'do all formio fixes' function called something like 'processFormSchema()" in /utils/formShemaUtils.js. as there may be more of these fixes and it seems a bit unexpected to just call this minor change.

loneil (approved):

TimCsaky (approved):

MichaelChung123:
Rabased

app/frontend/src/utils/transformUtils.js
MichaelChung123: oved this import in new commit

PR #219, opened 2021-08-10
Emails Restructuring

MichaelChung123:
# Description
- Added email instances when an Admin invites or uninvites a user to a Submission Draft
- Changed wording and structure of the 5 existing emails being sent out

## Types of changes
- Added new paths in the emailService.js file to send an email when a user is invited AND uninvited to a Submission Draft
- Changed wording in the email template files for the messages being sent.

## Checklist


bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-219

app/frontend/src/components/forms/submission/ManageSubmissionUsers.vue
```diff
@@ -228,7 +229,7 @@ export default {
         this.isLoadingTable = false;
       }
     },
-    async modifyPermissions(userId, permissions) {
+    async modifyPermissions(userId, permissions, isAssigned) {
```
loneil, line 232: I don't think(?) we need this new `isAssigned` flag on the endpoint?
The API side logic know if it's removing someone from a form because the permissions array is blank. Like how in \rbac\service.js it knows whether it's assigning a user to a form or removing them based on just body.permissions.

So the additional param all through the stack might be extraneous.

app/src/forms/rbac/controller.js
```diff
@@ -85,7 +86,13 @@ module.exports = {
   },
   setSubmissionUserPermissions:  async (req, res, next) => {
     try {
+      const submission = await formService.read(req.query.formSubmissionId, req.currentUser);
       const response = await service.modifySubmissionUser(req.query.formSubmissionId, req.query.userId, req.body, req.currentUser);
+
+      JSON.parse(req.query.isAssigned) ?
```
loneil, line 92: See comment above about body.permissions functioning as the assign/remove flag.

app/src/forms/rbac/controller.js
```diff
@@ -85,7 +86,13 @@ module.exports = {
...
+        emailService.submissionAssigned(submission.form.id, response[0], req.body.email, req.headers.referer).catch(() => { })
+        :
+        emailService.submissionUnassigned(submission.form.id, response[0], req.body.email, req.headers.referer).catch(() => { });
```
loneil, line 95: Is the email included in the body?

loneil (changes requested):
I'm not getting the invitee email when I add to a draft via Manage Team Members. The email address of the person being invited doesn't look to be sent in the manage submissions api call?

image

app/src/forms/rbac/controller.js
MichaelChung123: Made changes to target the email properly. Also removed the isAssigned and instead am checking for the permissions array being empty or populated with values.

app/src/forms/email/assets/bodies/submission-unassigned.html
```diff
@@ -0,0 +1,21 @@
+<table border="0" cellpadding="0" cellspacing="0" role="presentation" style="" width="100%">
+  <tr>
+    <td data-slot-container="1" align="center" style="
+      font-size: 0px;
```
jujaga, line 4: This line seems strange esp with a 14px font style one child DOM element down. What is it supposed to do?

app/src/forms/email/emailService.js
```diff
@@ -12,20 +12,28 @@ const generateContexts = (type, configData, submission, referer) => {
...
-    userTypePath = '/view?s=';
+    userTypePath = 'form/view?s=';
   }
   return [{
     context: {
       confirmationNumber: submission.confirmationId,
       title: configData.title,
       messageLinkText: configData.messageLinkText,
-      messageLinkUrl: `${service._appUrl(referer)}/form${userTypePath}${submission.id}`,
+      messageLinkUrl: `${service._appUrl(referer)}/${userTypePath}${submission.id}`,
```
jujaga, line 34: It'd be much better to have the query param part `?s=` be embedded into `messageLinkUrl`'s string directly instead of repeating it in the `userTypePath` variable.

app/src/forms/email/emailService.js
```diff
@@ -88,6 +97,53 @@ const service = {
...
+        messageLinkText: `You have been invited to a ${form.name} submission draft. You can review your submission draft details by visiting the following links:`,
+        priority: 'normal',
+        assignmentNotificationEmail,
+        form,
+      };
+      return service._sendEmailTemplate('sendSubmissionAssigned', configData, submission, referer);
+    } catch (e) {
+      log.error(`status: ${currentStatus}, referer: ${referer}`, { function: 'statusAssigned' });
+      log.error(e.message, { function: 'submissionAssigned' });
+      log.error(e);
```
jujaga, line 118: I'd strongly recommend utilizing more of winston's new logging features instead of multi-line reporting the same error repeatedly here. Winston logs are already output in structured JSON, so we should be taking advantage of that instead. Repeat for other similar sections. For example, you can try experimenting with the following to be more precise with your logging:
```js
log.error(e.message, {
  function: 'statusAssigned',
  status: currentStatus,
  referer: referer
});
```

app/src/forms/rbac/controller.js
```diff
@@ -85,7 +86,15 @@ module.exports = {
...
   setSubmissionUserPermissions:  async (req, res, next) => {
     try {
+      const submission = await formService.read(req.query.formSubmissionId, req.currentUser);
       const response = await service.modifySubmissionUser(req.query.formSubmissionId, req.query.userId, req.body, req.currentUser);
+      if (req.body && Array.isArray(req.body.permissions) && req.query.selectedUserEmail) {
+        // Check if we are adding or removing a user from the draft invite list. empty permissions signifies that we are removing permissions from a user.
+        req.body.permissions.length ?
+          emailService.submissionAssigned(submission.form.id, response[0], req.query.selectedUserEmail, req.headers.referer).catch(() => { })
+          :
+          emailService.submissionUnassigned(submission.form.id, response[0], req.query.selectedUserEmail, req.headers.referer).catch(() => { });
```
jujaga, line 96: Doing an in-line promise style chain catch like this isn't a good idea especially when this function is already set up to be written in async/await style. It'd be better to instead let the error actually be caught by the try catch and propagate.
```js
if(req.body.permissions.length) {
  await emailService.submissionAssigned(submission.form.id, response[0], req.query.selectedUserEmail, req.headers.referer);
} else {
  await emailService.submissionUnassigned(submission.form.id, response[0], req.query.selectedUserEmail, req.headers.referer);
}
```
Remove the `await` in my example if you don't actually need to evaluate if the call actually worked or not before returning response.

app/tests/unit/routes/v1/rbac.spec.js
```diff
@@ -327,9 +329,13 @@ describe(`GET ${basePath}/submissions`, () => {
 describe(`PUT ${basePath}/submissions`, () => {
 
   it('should return 200', async () => {
+    formService.read = jest.fn().mockReturnValue([]);
     // mock a success return value...
     service.modifySubmissionUser = jest.fn().mockReturnValue([]);
 
+    emailService.submissionAssigned = jest.fn().mockReturnValue([]);
+    emailService.submissionUnassigned = jest.fn().mockReturnValue([]);
```
jujaga, line 337: Looks like these were newly added in. Make sure when writing unit tests for things that need to be mocked that they are always `mockReset` or `mockRestore` after each unit test executed; otherwise you will have very strange side effects that are hard to catch in the future as more tests are added to this suite.
Ref: https://jestjs.io/docs/mock-function-api#mockfnmockreset

app/src/forms/email/emailService.js
MichaelChung123: removed the '?s=' from the userTypePath variable and put directly into the messageLinkUrl.

app/src/forms/email/assets/bodies/submission-unassigned.html
MichaelChung123: removed that font-size assignment. tested and looks good in emails still.

app/frontend/src/components/forms/submission/ManageSubmissionUsers.vue
```diff
@@ -231,16 +231,22 @@ export default {
...
             formSubmissionId: this.submissionId,
             userId: userId,
+            selectedUserEmail: selectedEmail,
           }
         );
         if (response.data) {
           this.userTableList = this.transformResponseToTable(response.data);
+          this.addNotification({
+            ...NotificationTypes.SUCCESS,
+            message: permissions.length ? `Invite email sent to ${selectedEmail}` : `Sent uninvited email to ${selectedEmail}`,
```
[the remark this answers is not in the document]
MichaelChung123: Changed the wording to match.

app/src/forms/email/emailService.js
MichaelChung123: Changed the error logs accordingly for all the emailService functions.

app/src/forms/rbac/controller.js
MichaelChung123: good catch, im making this an if else block instead. I will also be omitting the await call since we do not need to await them for the response.

app/tests/unit/routes/v1/rbac.spec.js
MichaelChung123: right, I have implemented a mockReset after the asserts of this test since this is the only location we are mocking the submissionAssigned and submissionUnassigned functions.

app/src/forms/email/emailService.js
```diff
@@ -96,41 +156,51 @@ const service = {
...
+      });
       throw e;
     }
   },
 
-  submissionReceived: async (formId, submissionId, referer) => {
+  // Sending confirmation of Form Submission
+  submissionConfirmation: async (formId, submissionId, body, referer) => {
     try {
       const form = await formService.readForm(formId);
```
loneil, line 205: Believe this should have the identity provider in it to figure out if the form is public or not (see review comment)

loneil (changes requested):
A **public** form submission can't have the link back to "View your * submissions" since the user who filled out the form is unauthenticated, so there's no tracking of their submissions.
So will have to do an identity provider check and then conditionally remove that second link

image

app/src/forms/email/emailService.js
MichaelChung123: doing a check for the identityProviders to remove the link on public.

app/src/forms/email/emailService.js
```diff
@@ -182,15 +188,19 @@ const service = {
...
       return service._sendEmailTemplate('sendSubmissionReceived', configData, submission, referer);
     } catch (e) {
-      log.error(`formId: ${formId}, submissionId: ${submissionId}, body: ${JSON.stringify(body)}, referer: ${referer}`, { function: 'submissionConfirmation' });
-      log.error(e.message, { function: 'submissionReceived' });
-      log.error(e);
+      log.error(e.message, e, {
+        function: 'submissionReceived',
+        formId: formId,
+        submissionId: submissionId,
+        body: JSON.stringify(body),
```
jujaga, line 195: It may be fine to leave the body as a JSON object directly, as stringifying it will likely add a bunch of character encoding wrappers on the content. Winston can handle direct object references while logging in JSON mode.

app/src/forms/email/emailService.js
```diff
@@ -200,13 +210,18 @@ const service = {
...
       return service._sendEmailTemplate('sendSubmissionConfirmation', configData, submission, referer);
     } catch (e) {
-      log.error(`formId: ${formId}, submissionId: ${submissionId}, referer: ${referer}`, { function: 'submissionReceived' });
-      log.error(e.message, { function: 'submissionConfirmation' });
-      log.error(e);
+      log.error(e.message, e, {
+        function: 'submissionConfirmation',
+        formId: formId,
+        submissionId: submissionId,
+        body: JSON.stringify(body),
```
jujaga, line 224: It may be fine to leave the body as a JSON object directly, as stringifying it will likely add a bunch of character encoding wrappers on the content. Winston can handle direct object references while logging in JSON mode.

jujaga (approved):

loneil (changes requested):
The **invite** email link (see screenshot below) to the list of submissions is going back to the staff page rather than the submitter page (ie the /user one)

image

app/src/forms/email/emailService.js
MichaelChung123: removing the JSON.stringify call on the body

app/src/forms/email/emailService.js
MichaelChung123: removing the JSON.stringify call on the body

app/src/forms/email/emailService.js
```diff
@@ -204,7 +204,7 @@ const service = {
     try {
       const form = await formService.readForm(formId);
       const submission = await formService.readSubmission(submissionId);
-      const bodyTemplate = form.identityProviders[0].idp.includes('public') ? 'submission-received-confirmation-public.html' : 'submission-received-confirmation-login.html';
+      const bodyTemplate = form.identityProviders.length > 0 && form.identityProviders[0].idp.includes('public') ? 'submission-received-confirmation-public.html' : 'submission-received-confirmation-login.html';
```
loneil, line 207: `.idp` is a string or another array? If it's a string it wouldn't be the includes check would it?

Includes would be for if you were checking the `.identityProvider` array for at least 1 object that has an `idp === 'public'

loneil (approved):
Functionally working, don't think the includes check in the IDP (see last comment) is in the right spot but for the way you're allowed to do IDP setup will work as is

PR #229, opened 2021-09-08
Fix hidden file support

loneil:
# Description
https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-2205

When there's a hidden file upload (conditional or anything) there was a 500 error occurring. Fix that, extract the logic where the bug was, and put unit tests to handle all cases.

## Types of changes

## Checklist

## Further comments

Was adding an undefined element to the array returned by flatMap (if the file control was not in the schema but not the data). Tweak so the flatMap skips in that case
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap#for_adding_and_removing_items_during_a_map

### DIFF
--- a/app/src/forms/form/service.js
+++ b/app/src/forms/form/service.js
@@ -24,6 +24,18 @@ const Rolenames = [Roles.OWNER, Roles.TEAM_MANAGER, Roles.FORM_DESIGNER, Roles.S
 
 const service = {
 
+  // Get the list of file IDs from the submission
+  _findFileIds: (schema, data) => {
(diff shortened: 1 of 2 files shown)


jujaga (approved):

bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-229

TimCsaky (approved):
looks very good

PR #230, opened 2021-09-09
find fields nested in columns in a form schema

TimCsaky:
# Description

the `readVersionFields()` function that is called when customizing columns doesn't find the fields for this form design:
https://chefs.nrs.gov.bc.ca/app/form/submissions?f=b6c5b8c8-0961-4347-b0fe-4153b3c67b31
It should look for components nested in the 'columns' component.

the unit tests i added were more of an afterthought.. 
A real form schema (fixtures to come in another pr) will be more useful.

## Checklist

### DIFF
--- a/app/src/forms/form/service.js
+++ b/app/src/forms/form/service.js
@@ -268,11 +268,17 @@ const service = {
     // TODO: Consider if this should be a form utils function instead?
     const findFields = (obj) => {
       const fields = [];
-      // Only add key if it is an input and visible
-      if (obj.input && !obj.hidden) fields.push(obj.key);
-      // Check children components that aren't inputs
-      else if (!obj.hidden && obj.components && obj.components.length) {
-        fields.push(obj.components.flatMap(o => findFields(o)));
+      if (!obj.hidden) {
+        // Only add key if it is an input and visible
+        if (obj.input) fields.push(obj.key);
+        // Recursively check all children attributes that are arrays
+        else {
+          Object.keys(obj).forEach(key => {
+            if (Array.isArray(obj[key]) && obj[key].length) {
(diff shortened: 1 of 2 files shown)


bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-230

loneil (approved):

app/src/forms/form/service.js
```diff
@@ -274,6 +274,10 @@ const service = {
       else if (!obj.hidden && obj.components && obj.components.length) {
         fields.push(obj.components.flatMap(o => findFields(o)));
       }
+      // check for components inside columns
+      else if (!obj.hidden && obj.columns && obj.columns.length) {
+        fields.push(obj.columns.flatMap(o => findFields(o)));
+      }
```
jujaga, line 280: It may be better to consider rewriting the findFields function definition to be more generic with its recursion approach by checking all keys instead of only a few hard-coded ones like components and columns:
```js
    // Recursively find all field key names
    // TODO: Consider if this should be a form utils function instead?
    const findFields = (obj) => {
      const fields = [];
      if (!obj.hidden) {
        // Only add key if it is an input and visible
        if (obj.input) fields.push(obj.key);
        // Recursively check all children attributes that are arrays
        else {
          Object.keys(obj).forEach(key => {
            if (Array.isArray(obj[key]) && obj[key].length) {
              fields.push(obj[key].flatMap(o => findFields(o)));
            }
          });
        }
      }
      return fields.flat();
    };
```

TimCsaky: good idea. thanks for the code.

PR #235, opened 2021-09-20
css change to remove asterisk after inputs in data-grid

TimCsaky:
# Description
formio default shows asterisks next to required input in a data grid, as well as column headers.
Looks a bit extraneous. I'm hoping this is an unobtrusive fix without side-affects.

## Checklist

### DIFF
--- a/app/frontend/src/assets/scss/style.scss
+++ b/app/frontend/src/assets/scss/style.scss
@@ -437,6 +437,13 @@ a,
     color: $bcgov-error;
   }
 
+  // hide asterisks after input in Data Grid when labels appear as table headers
+  .formio-component-datagrid {
+    label.field-required.control-label--hidden::after {
+      content: "";
+    }
+  }
+
   // dropdowns
   .choices__list--dropdown {
     .choices__item {


bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-235

app/frontend/src/assets/scss/style.scss
```diff
@@ -437,6 +437,13 @@ a,
     color: $bcgov-error;
   }
 
+  // hide asterisks in after input in Data Grid becasue they appear after column headers
+  .formio-component-datagrid {
+    label.field-required:after {
```
loneil, line 442: Should psuedo elements like hover and after and stuff use `::`?
https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements

or does it work either way in SCSS

loneil (approved):
👍 missed the discussion about going back to red asterisks when I was away I guess but I like it.

jujaga (approved):

PR #239, opened 2021-09-29
include custom form component templates a better way

TimCsaky:
# Description

My pr to add aria/required tags to inputs (pr 233) introduced some console statements saying 'x plugin not found' when you view a form.
I believe the way i was including the templates (in the main.js) was not waiting for the formio base templates and plugins to load.
i think this is a better way of doing it.

## Types of changes

## Checklist

### DIFF
--- a/app/frontend/src/components/designer/FormViewer.vue
+++ b/app/frontend/src/components/designer/FormViewer.vue
@@ -76,15 +76,12 @@ import Vue from 'vue';
 import { mapActions, mapGetters } from 'vuex';
 import { Form } from 'vue-formio';
 
+import templateExtensions from '@/plugins/templateExtensions';
 import { formService, rbacService } from '@/services';
 import FormViewerActions from '@/components/designer/FormViewerActions.vue';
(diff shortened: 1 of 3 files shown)


bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-239

app/frontend/src/components/designer/FormViewer.vue
```diff
@@ -75,6 +75,8 @@
 import Vue from 'vue';
 import { mapActions, mapGetters } from 'vuex';
 import { Form } from 'vue-formio';
+import templateExtensions from '@/plugins/templateExtensions';
+Form.use(templateExtensions);
```
jujaga, line 79: These lines should be inserted after the second block of imports because templateExtensions is something we wrote, not an external library import.

MichaelChung123 (approved):

TimCsaky:
found another way to include template overrides, in the 'options' passed when you initialize the form renderer and builder.

jujaga (approved):

PR #247, opened 2021-10-27
changing global username

MichaelChung123:
# Description
Adding the appropriate IDP at the end of the usernames to write to the DB.

https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-2242

## Types of changes
Made changes to the auth service layer to give the appropriate username format to all the routes/controllers.

## Checklist


bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-247

loneil (changes requested):
Blocking until we discuss the additional changes to user db record happening

loneil (approved):

app/src/forms/auth/service.js
```diff
@@ -129,8 +130,10 @@ const service = {
       // what if name or email changed?
       user = await service.updateUser(user.id, obj);
     }
+
+    const usernameIdp = user.idpCode ? `${user.username}@${user.idpCode}` : user.username;
     // return with the db id...
-    return { id: user.id, ...userInfo };
+    return { id: user.id, usernameIdp: usernameIdp,  ...userInfo };
```
jujaga, line 136: Now that I look at this, may not even need to do that const declaration and in-line the ternary directly as an option:
```js
return {
  id: user.id,
  usernameIdp: user.idpCode ? `${user.username}@${user.idpCode}` : user.username,
  ...userInfo
};
```

jujaga (approved):
Logically looks pretty good. Just a minor comment to consider. Tentatively approving as I'm not seeing anything too major otherwise. 👍

app/src/forms/auth/service.js
MichaelChung123: okay, will change this to be an inline ternary value

app/src/docs/v1.api-spec.yaml
```diff
@@ -1723,6 +1723,11 @@ paths:
           schema:
             type: string
           description: user username
+        - in: query
+          name: usernameIdp
+          schema:
+            type: string
+          description: user username with idp type appended
```
loneil, line 1730: These ones not needed

TimCsaky (approved):

PR #261, opened 2021-12-01
Feat/global token user context

naomiaro:
# Description
Uses formio's option `evalContext` to change their default `user` and `token` form context to be retrieved from our Vuex store instead.

## Types of changes

## Checklist

https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-2266

Front-end user object is taken from looking at this middleware:
https://github.com/bcgov/common-hosted-form-service/blob/master/app/src/forms/auth/service.js#L75

To overwrite the Formio user context: https://github.com/formio/formio.js/blob/73bb21a6d158db2d6eb33250ceff2997390ac398/src/Element.js#L549
so we can avoid all reads to localstorage.

Docs: https://github.com/bcgov/common-hosted-form-service/wiki/Custom-Default-Value


bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-261

bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-261

app/frontend/src/store/modules/auth.js
```diff
@@ -34,23 +41,46 @@ export default {
...
     tokenParsed: () => Vue.prototype.$keycloak.tokenParsed,
-    userName: () => Vue.prototype.$keycloak.userName
+    userName: () => Vue.prototype.$keycloak.userName,
+
+    user: (_state, getters) => {
+      const username = getters.authenticated
+        ? getters.tokenParsed.identity_provider_identity
+          ? getters.tokenParsed.identity_provider_identity
+          : getters.tokenParsed.preferred_username
+        : '';
```
jujaga, line 63: I would recommend avoiding nested ternaries. While they definitely work the way its written, they are cognitively harder to process. It would be better to write this as a `let username = ''` and then if certain conditions are met, reassigned instead in a more classic if/else format.

This also needs to be unit tested. Refer to the `auth.getters.spec.js` file.

app/frontend/src/store/modules/auth.js
```diff
@@ -34,23 +41,46 @@ export default {
...
+        username,
+        firstName: getters.authenticated ? getters.tokenParsed.given_name : '',
+        lastName: getters.authenticated ? getters.tokenParsed.family_name : '',
+        fullName: getters.authenticated ? getters.tokenParsed.name : '',
+        email: getters.authenticated ? getters.tokenParsed.email : '',
+        idp: getters.authenticated
+          ? getters.tokenParsed.identity_provider
+          : 'public',
+        public: !!getters.authenticated,
+      };
```
jujaga, line 75: Not quite too sure what this getter function is needed for - we're constructing an object that a user can easily access directly by referencing token (tokenParsed) on formio side and going for the raw data directly if available.

If we do need this, should we consider having the default be assigned as undefined instead of an empty string? It's generally better to have an attribute be undefined if they're falsy.

app/frontend/src/store/modules/auth.js
```diff
@@ -34,23 +41,46 @@ export default {
...
     moduleLoaded: () => !!Vue.prototype.$keycloak,
     realmAccess: () => Vue.prototype.$keycloak.tokenParsed.realm_access,
-    redirectUri: state => state.redirectUri,
+    redirectUri: (state) => state.redirectUri,
     resourceAccess: () => Vue.prototype.$keycloak.tokenParsed.resource_access,
     token: () => Vue.prototype.$keycloak.token,
     tokenParsed: () => Vue.prototype.$keycloak.tokenParsed,
-    userName: () => Vue.prototype.$keycloak.userName
+    userName: () => Vue.prototype.$keycloak.userName,
+
```
jujaga, line 57: Unnecessary whitespace line.

jujaga (commented):
I think evalContext for token should be sufficient as that would provide the parsed JWT contents (user/username, email, auth type iirc). Not convinced we need the user evalContext atm as that looks to just be an object restructure of tokenParsed.

app/frontend/src/components/designer/FormDesigner.vue
```diff
@@ -182,6 +194,7 @@ export default {
     };
   },
   computed: {
+    ...mapGetters('auth', ['tokenParsed', 'user']),
```
loneil, line 197: No sure if we need stuff in the Form Designer for the token? Could be wrong, worth a discussion.

Like is a token needed at design time at all? That would be the logged in user _creating_ the form at that time...

app/frontend/src/store/modules/auth.js
loneil: Not sure we need the user object and the token object?
If there's a formIO instruction concept of `user` as well as `token` it would be helpful, as people might refer to formIO documentation. So if that case I'd leave it in (then we'd want to ask ourself if `token` should be the parsed one or the encoded one...) I haven't dug into the FormIO expected behaviour so we could check up on that and try to match it.

loneil: I see this in the existing form designer UI, so decoded token is expected yeah

image

app/frontend/src/components/designer/FormDesigner.vue
```diff
@@ -277,11 +290,15 @@ export default {
             weight: 60,
             components: {
               orgbook: true,
-              simplefile: this.userType !== this.ID_MODE.PUBLIC
+              simplefile: this.userType !== this.ID_MODE.PUBLIC,
             },
           },
         },
         templates: templateExtensions,
+        evalContext: {
```
loneil, line 298: Did you find any good source of documentation on this?
I've seen it referenced a bunch in the past but never a good central explainer...

components/src/overrides/editform/utils.ts
```diff
@@ -0,0 +1,13 @@
+import EditFormUtils from 'formiojs/components/_classes/component/editForm/utils';
+
+const originalLogicVariablesTable = EditFormUtils.logicVariablesTable;
+
+// decorate this function to include global updates to evalContext.
+function logicVariablesTable(additional = '') {
+  const customEval =
```
loneil, line 7: Looks like this is adding "token" twice to the "Calculated Value" section

image

app/frontend/src/components/designer/FormDesigner.vue
loneil: Figured it out, this lets you see the result of what you set in Custom Default Value at design time 👍

app/frontend/src/components/designer/FormDesigner.vue
loneil: Figured it out, this lets you see the result of what you set in Custom Default Value at design time 👍

app/frontend/src/store/modules/auth.js
naomiaro: Formio also has a user field looking into localstorage, which is why I overwrote it as well.

components/src/overrides/editform/utils.ts
naomiaro: hmm I actually don't get the token line showing up for me, I wonder why you had the original show up originally.

naomiaro: Do you still have a localstorage variable set? Maybe that is still updating this table somehow.

app/frontend/src/store/modules/auth.js
naomiaro: formio was also reading a user context from localstorage, so I created this to block any potential reads to localstorage without a `user` variable existing. The `user` variable could also be easier to reason about for the end user and isn't so reliant on any token structure itself.

app/frontend/src/store/modules/auth.js
naomiaro: https://github.com/formio/formio.js/blob/73bb21a6d158db2d6eb33250ceff2997390ac398/src/Element.js#L549

app/frontend/src/store/modules/auth.js
naomiaro: https://github.com/formio/formio.js/blob/73bb21a6d158db2d6eb33250ceff2997390ac398/src/Element.js#L549

app/frontend/src/store/modules/auth.js
loneil: Yeah makes sense to me if it's a formio existing field that people could come across to override that as well.

components/src/overrides/editform/utils.ts
naomiaro: Now realised it's a different table, checking to see how to avoid this..

app/frontend/src/store/modules/auth.js
naomiaro: As for undefined, more parsing would then be needed on both form viewer and designer to set to `''` so that the text inputs get a good default value.

bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-261

TimCsaky (approved):

loneil (approved):
Keeping `user` construct as well as token makes sense after discussions (formIO override)

app/frontend/src/store/modules/auth.js
```diff
@@ -34,23 +41,58 @@ export default {
...
+        username,
+        firstName,
+        lastName,
+        fullName,
+        email,
+        idp,
+        public: !getters.authenticated,
+      };
+
+      return user;
```
jujaga, line 89: You can optimize the number of assignments and discrete variables being used (and thus memory footprint) by initializing the user object earlier in the function and assigning into those attributes instead:
```js
const user = {
  username: '',
  firstName: '',
  lastName: '',
  fullName: '',
  email: '',
  idp: 'public',
  public: !getters.authenticated
};

if (getters.authenticated) {
  if (getters.tokenParsed.identity_provider_identity) {
    user.username = getters.tokenParsed.identity_provider_identity;
  } else {
    user.username = getters.tokenParsed.preferred_username;
  }

  user.firstName = getters.tokenParsed.given_name;
  user.lastName = getters.tokenParsed.family_name;
  user.fullName = getters.tokenParsed.name;
  user.email = getters.tokenParsed.email;
  user.idp = getters.tokenParsed.identity_provider;
}

return user;
```

jujaga (approved):

PR #264, opened 2021-12-03
Populating Recipient Email

MichaelChung123:
# Description
When selecting Revising option, the Recipient Email textbox now populates with the Submitter's email address.
Ref: https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-2297?

## Types of changes
Made changes to the StatusPanel Component on the Front End to make a call to the rbacService to filter the response and grab the submission owner's email.

## Checklist

### DIFF
--- a/app/frontend/src/components/forms/submission/MySubmissionsTable.vue
+++ b/app/frontend/src/components/forms/submission/MySubmissionsTable.vue
@@ -156,6 +156,9 @@ export default {
     // Status columns in the table
     getCurrentStatus(record) {
       if (record.draft) {
+        if (record.submissionStatus[0].code === 'REVISING') {
+          return 'REVISING';
+        }
         return 'DRAFT';
       } else {
         // Current status is most recent status (top in array, query returns in status created desc)
--- a/app/frontend/src/components/forms/submission/MySubmissionsTable.vue
+++ b/app/frontend/src/components/forms/submission/MySubmissionsTable.vue
@@ -155,16 +155,14 @@ export default {
 
     // Status columns in the table
     getCurrentStatus(record) {
-      if (record.draft) {
-        if (record.submissionStatus[0].code === 'REVISING') {
-          return 'REVISING';
-        }
+      // Current status is most recent status (top in array, query returns in status created desc)
+      const status = record.submissionStatus && record.submissionStatus[0]
+        ? record.submissionStatus[0].code
+        : 'N/A';
+      if (record.draft && status !== 'REVISING') {
         return 'DRAFT';
       } else {
-        // Current status is most recent status (top in array, query returns in status created desc)
-        return record.submissionStatus && record.submissionStatus[0]
-          ? record.submissionStatus[0].code
-          : 'N/A';
+        return status;
       }
     },
     getStatusDate(record, statusCode) {
--- a/app/frontend/src/components/forms/submission/StatusPanel.vue
+++ b/app/frontend/src/components/forms/submission/StatusPanel.vue
@@ -24,7 +24,7 @@
               item-value="code"
               v-model="statusToSet"
               :rules="[(v) => !!v || 'Status is required']"
-              @change="statusFields = true"
+              @change="onChange(statusToSet)"
             />
 
             <div v-show="statusFields" v-if="showRevising">
@@ -212,7 +212,7 @@ export default {
   },
   computed: {
     ...mapGetters('auth', ['keycloakSubject']),
-    ...mapGetters('form', ['formSubmission']),
+    ...mapGetters('form', ['form', 'formSubmission', 'submissionUsers']),
     // State Machine
     showRevising() {
       return ['REVISING'].includes(this.statusToSet);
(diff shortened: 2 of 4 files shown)


bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-264

app/frontend/src/components/forms/submission/StatusPanel.vue
```diff
@@ -224,6 +225,7 @@ export default {
   },
   methods: {
     ...mapActions('notifications', ['addNotification']),
+    ...mapActions('form', ['fetchSubmission']),
```
loneil, line 228: I think this `fetchSubmission` is unused?

jujaga: Yea it's currently unused - doesn't need to be imported in.

app/frontend/src/components/forms/submission/StatusPanel.vue
```diff
@@ -276,9 +278,6 @@ export default {
             (sc) => sc.code === this.currentStatus.code
           ).statusCode;
           this.items = this.currentStatus.statusCodeDetail.nextCodes;
-
-          // Hiding REVISING status until backend can receive submitter's email.
-          this.items = this.items.filter((item) => item !== 'REVISING');
```
jujaga, line 281: We can't wholesale remove this -> we still need to remove Revising from the list if and only if the form does NOT allow users to create drafts. That should be a boolean somewhere in the form definition level. We'll need to make sure that is in place before enabling return to user flow.

jujaga (commented):
Some potentially inconsistent behavior with cached responses and navigation from all submissions to specific submission page.

app/frontend/src/components/forms/submission/StatusPanel.vue
```diff
@@ -360,6 +378,7 @@ export default {
   },
   created() {
     this.getStatus();
+    this.setRecipientEmail();
```
jujaga, line 381: Consider having this method only being invoked when the user selects revising state instead of running it all the time on page/component load to reduce computation and network calls.

bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-264

app/frontend/src/store/modules/form.js
```diff
@@ -309,6 +314,18 @@ export default {
         }, { root: true });
       }
     },
+    async fetchSubmissionUsers({ commit, dispatch }, { formSubmissionId }) {
```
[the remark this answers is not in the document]
loneil: Yeah if it's a single one it can be a single stand alone argument, though if it's more than 1 you have to do an object like that. If you look in the file we've been a bit inconsistent for single args so this one could go either way fine.

loneil (changes requested):
Couple minor tweaks after trying it out and discussing.

On the submission table (for the submitting user after getting it sent back to them) we'd want to keep the status as 'REVISING' rather than setting to DRAFT for this point in the flow.
`MySubmissionsTable.vue : 158` can alter logic there to see if it has a status of Revising and not do the draft label logic if so. Might have to move the conditionals in `getCurrentStatus` around a little.

The Delete Draft doesn't work (intentionally, because it's been submitted) so that icon will  have to hide if the status is REVISING. I **think** just doing the 1 thing above will take care of this as well.

Everything looking good so far otherwise

app/frontend/src/components/forms/submission/MySubmissionsTable.vue
```diff
@@ -156,6 +156,9 @@ export default {
     // Status columns in the table
     getCurrentStatus(record) {
       if (record.draft) {
+        if (record.submissionStatus[0].code === 'REVISING') {
+          return 'REVISING';
+        }
         return 'DRAFT';
       } else {
         // Current status is most recent status (top in array, query returns in status created desc)
```
loneil, line 164: If we need the status now in both parts of the conditional, we'd still want to gaurd against exceptions, so I'd refactor this to 
```
      // Current status is most recent status (top in array, query returns in status created desc)
      const status = record.submissionStatus && record.submissionStatus[0]
          ? record.submissionStatus[0].code
          : 'N/A';
      if (record.draft && status !== 'REVISING') {
        return 'DRAFT';
      } else {
        return status
      }
 ```

**Review by loneil — APPROVED** [review]

**Review by jujaga — APPROVED** [review]

**Review by naomiaro — APPROVED** [review]

File: app/frontend/src/components/forms/submission/StatusPanel.vue [hunk 6]
```diff
@@ -276,7 +296,8 @@ export default {
             (sc) => sc.code === this.currentStatus.code
           ).statusCode;
           this.items = this.currentStatus.statusCodeDetail.nextCodes;
-
+        }
+        if (!this.form.enableSubmitterDraft) {
           // Hiding REVISING status until backend can receive submitter's email.
```
TimCsaky, line 301: this comment can be removed

PR #278, opened 2022-02-16
Completed Status Email Notification

MichaelChung123:
# Description
When an admin sets a submission to Completed, it sends a confirmation email to the submission user.

## Types of changes
Made changes to the StatusPanel and emailService to detect when an admin is trying to set the status for a submission to Completed.

## Checklist

### DIFF
--- a/app/frontend/src/components/forms/submission/StatusPanel.vue
+++ b/app/frontend/src/components/forms/submission/StatusPanel.vue
@@ -28,7 +28,7 @@
             />
 
             <div v-show="statusFields" v-if="showRevising">
-              <v-text-field v-model="revisionEmail" label="Recipient Email" outlined dense />
+              <v-text-field v-model="submissionUserEmail" label="Recipient Email" outlined dense />
               <v-checkbox v-model="addComment" :label="'Attach Comment to Email'" />
               <div v-if="addComment">
                 <label>Email Comment</label>
@@ -205,7 +205,7 @@ export default {
       loading: true,
       note: '',
       emailComment: '',
-      revisionEmail: '',
+      submissionUserEmail: '',
       statusHistory: {},
       statusFields: false,
       statusToSet: '',
@@ -250,7 +250,7 @@ export default {
             const username = data.user.idpCode ? `${data.user.username}@${data.user.idpCode}` : data.user.username;
             return username === this.formSubmission.createdBy;
           });
-          this.revisionEmail = submitterData.user.email;
+          this.submissionUserEmail = submitterData.user.email;
         } catch (error) {
           this.addNotification({
             message: 'An error occured while trying to fetch recipient emails for this submission.',
@@ -330,7 +330,7 @@ export default {
       this.emailComment = '';
       this.statusFields = false;
       this.$refs.form.resetValidation();
-      this.revisionEmail = '';
+      this.submissionUserEmail = '';
       this.statusToSet = null;
       this.note = '';
     },
@@ -343,7 +343,7 @@ export default {
 
           const statusBody = {
             code: this.statusToSet,
-            revisionNotificationEmail: this.revisionEmail,
+            submissionUserEmail: this.submissionUserEmail,
             revisionNotificationEmailContent: this.emailComment,
             confirmCompleted: this.addConfirmCompleted
           };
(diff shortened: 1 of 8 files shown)


bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-278

TimCsaky:
i might be forgetting something but i'm not seeing the 'Revising' option in the status dropdown anymore.

TimCsaky:
looks good . I guess we decided the 'comment' you can enter when you set the status to 'Revising' is not using the same 'note' construct used when you set to 'assigned' and 'completed'. i suppose this has the benefit of allowing the 'Comment' to be private (not visible to other form reviewers). I kind of wonder if it will be easier to make all status changes use the same 'note' but the way you have it is fine too.

Its very hard to test this with a second idir that actually has an email address 😆

loneil (changes requested):
Get an error on a **public** form when picking COMPLETED in the box.
Figure the Send Confirmation Email checkbox should be hidden in that case as well. (also remember you can change the login type of a form after submissions have already happened, but that might not be an edge case you want to handle, but consider it)

image

bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-278

loneil (approved):
Issue with public forms is fixed, IDIR forms email appears to be sending as expected

app/src/forms/submission/controller.js
```diff
@@ -62,16 +62,28 @@ module.exports = {
...
+          } else if (req.body.code === Statuses.REVISING && req.body.submissionUserEmail) {
+            emailService.statusRevising(submission.form.id, response[0], req.body.submissionUserEmail, req.body.revisionNotificationEmailContent, req.headers.referer).catch(() => { });
+          } else if (req.body.code === Statuses.COMPLETED && req.body.submissionUserEmail && req.body.confirmCompleted) {
+            emailService.statusCompleted(submission.form.id, response[0], req.body.submissionUserEmail, req.headers.referer).catch(() => { });
+          }
+          res.status(200).json(response);
+        })
+        .catch(error => {
+          next(error);
+        });
```
jujaga, line 86: You can avoid writing in promise-chain style since you're already awaiting on the Promise.all - don't need to catch twice. It's generally good practice to either stick with async/await style OR promise-chain style. Mixing both types can cause a lot of grief/confusion down the line with unintended/uncaught behavior.
```js
const results = await Promise.all(tasks);
const response = results[0];
const submission = results[1];
// send an email (async in the background)
if (req.body.code === Statuses.ASSIGNED && req.body.assignmentNotificationEmail) {
  emailService.statusAssigned(submission.form.id, response[0], req.body.assignmentNotificationEmail, req.headers.referer).catch(() => { });
} else if (req.body.code === Statuses.REVISING && req.body.submissionUserEmail) {
  emailService.statusRevising(submission.form.id, response[0], req.body.submissionUserEmail, req.body.revisionNotificationEmailContent, req.headers.referer).catch(() => { });
} else if (req.body.code === Statuses.COMPLETED && req.body.submissionUserEmail && req.body.confirmCompleted) {
  emailService.statusCompleted(submission.form.id, response[0], req.body.submissionUserEmail, req.headers.referer).catch(() => { });
  }

res.status(200).json(response);
```
The trailing `.catch(() => { })` on the emailService calls aren't great practice either; we should consider doing at minimum a `log.warn` of the error, seeing as an email would have failed to dispatch in those cases.

app/frontend/src/components/forms/submission/StatusPanel.vue
```diff
@@ -238,14 +244,20 @@ export default {
...
+          if (this.submissionUsers.data.length > 0) {
+            const submitterData = this.submissionUsers.data.find((data) => {
+              const username = data.user.idpCode ? `${data.user.username}@${data.user.idpCode}` : data.user.username;
+              return username === this.formSubmission.createdBy;
+            });
+            this.submissionUserEmail = submitterData.user.email;
+            if (status === 'COMPLETED') {
+              this.showSendConfirmEmail = true;
+            }
+          }
```
jujaga, line 260: Can avoid double-checking the same thing as Lucas mentioned:
```js
const submitterData = this.submissionUsers.data.find((data) => {
  const username = data.user.idpCode ? `${data.user.username}@${data.user.idpCode}` : data.user.username;
  return username === this.formSubmission.createdBy;
});
if (submitterData) {
  this.submissionUserEmail = submitterData.user ? submitterData.user.email : undefined;
  this.showSendConfirmEmail = status === 'COMPLETED';
}
```

jujaga (approved):

TimCsaky (approved):

PR #280, opened 2022-02-17
Stopping CHES Call when no email provided on Submit

MichaelChung123:
# Description
Stopping the CHES call when there are no emails provided.

## Types of changes
Made changes to the emailService so that it checks the contexts variable for any emails to send to.

## Checklist

### DIFF
--- a/app/src/forms/email/emailService.js
+++ b/app/src/forms/email/emailService.js
@@ -287,8 +287,10 @@ const service = {
   submissionReceived: async (formId, submissionId, body, referer) => {
     try {
       const { configData, contexts } = await buildEmailTemplate(formId, submissionId, EmailTypes.SUBMISSION_RECEIVED, referer, { body });
-      if (contexts[0].to.length > 0) {
+      if (contexts[0].to.length) {
         return service._sendEmailTemplate(configData, contexts);
+      } else {
+        return {};
       }
     } catch (e) {
       log.error(e.message, {
--- a/app/src/forms/email/emailService.js
+++ b/app/src/forms/email/emailService.js
@@ -287,8 +287,9 @@ const service = {
   submissionReceived: async (formId, submissionId, body, referer) => {
     try {
       const { configData, contexts } = await buildEmailTemplate(formId, submissionId, EmailTypes.SUBMISSION_RECEIVED, referer, { body });
-
-      return service._sendEmailTemplate(configData, contexts);
+      if (contexts[0].to.length > 0) {
+        return service._sendEmailTemplate(configData, contexts);
+      }
     } catch (e) {
       log.error(e.message, {
         function: EmailTypes.SUBMISSION_RECEIVED,


app/src/forms/email/emailService.js
```diff
@@ -287,8 +287,9 @@ const service = {
   submissionReceived: async (formId, submissionId, body, referer) => {
     try {
       const { configData, contexts } = await buildEmailTemplate(formId, submissionId, EmailTypes.SUBMISSION_RECEIVED, referer, { body });
-
-      return service._sendEmailTemplate(configData, contexts);
+      if (contexts[0].to.length > 0) {
```
jujaga, line 290: Use shorthand truthiness check:
```js
if (contexts[0].to.length) {
```

app/src/forms/email/emailService.js
```diff
@@ -287,8 +287,9 @@ const service = {
...
+        return service._sendEmailTemplate(configData, contexts);
```
jujaga, line 291: The specification of this function needs to return the result of the operation. By if-blocking this out, this function now returns nothing; if the caller function depends on the returned object, we're introducing a potential bug. We should have an else return { ...something}.

jujaga (changes requested):

app/src/forms/email/emailService.js
loneil: I'm not sure this should be based on checking contexts in the email service. Aren't the contexts set by the supplied email address in the body, so shouldn't it check a level or 2 higher to even see if email service should be called at all?

So check body email address at the controller level. The code flow is really "do I need to do anything about an email here?" when controlling the submission POST

loneil: IE, controller.createSubmission line 126

app/src/forms/email/emailService.js
```diff
@@ -287,8 +287,11 @@ const service = {
   submissionReceived: async (formId, submissionId, body, referer) => {
     try {
       const { configData, contexts } = await buildEmailTemplate(formId, submissionId, EmailTypes.SUBMISSION_RECEIVED, referer, { body });
-
-      return service._sendEmailTemplate(configData, contexts);
+      if (contexts[0].to.length) {
+        return service._sendEmailTemplate(configData, contexts);
+      } else {
+        return {};
+      }
```
loneil, line 294: Really don't think this is the right place to be doing this? It's not the email service deciding based on the CHES contexts whether to send, it should be the CONTROLLER deciding to invoke email service based on the request
See other comment below:

I'm not sure this should be based on checking contexts in the email service. Aren't the contexts set by the supplied email address in the body, so shouldn't it check a level or 2 higher to even see if email service should be called at all?

So check body email address at the controller level. The code flow is really "do I need to do anything about an email here?" when controlling the submission POST

IE, controller.createSubmission line 126

loneil (commented):
@jujaga your thoughts?

loneil:
Actually checking it out the form detail is only known at the `buildEmailTemplate` point so my other suggestion might not be valid... though still a little uneasy about only doing this contexts check in one of the email handlers when it really would apply to all... but it should fix this particular issue at least I think. Might consider refactoring later to apply more globally?

loneil (approved):
This should work, discussed about considering that "is there a `to` address" check in other stuff but can consider later

TimCsaky (approved):

jujaga (changes requested):
Please do a proper rebase of this branch.

bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-280

jujaga (approved):

PR #281, opened 2022-03-14
Adding Comments to Status Changes and improving Notes Panel

MichaelChung123:
# Description
Restructuring how we handle statuses in accordance with Notes and Comments. Added functionality to add a comment to any Status update so that recipients can receive a custom message in their email.

Comments and Notes both get added into the Notes list for internal records purposes.

## Types of changes
- Made changes to Notes panel to have a proper icon instead of text
- Added option to add comments to all status updates
- Added appropriate content to tests

## Checklist

### DIFF
--- a/app/frontend/src/components/forms/submission/StatusPanel.vue
+++ b/app/frontend/src/components/forms/submission/StatusPanel.vue
@@ -375,7 +375,7 @@ export default {
             const noteBody = {
               submissionId: this.submissionId,
               submissionStatusId: submissionStatusId,
-              notes: formattedComment,
+              note: formattedComment,
               userId: user.data.id,
             };
             const response = await formService.addNote(
--- a/app/frontend/src/components/forms/submission/StatusPanel.vue
+++ b/app/frontend/src/components/forms/submission/StatusPanel.vue
@@ -375,7 +375,7 @@ export default {
             const noteBody = {
               submissionId: this.submissionId,
               submissionStatusId: submissionStatusId,
-              emailComment: formattedComment,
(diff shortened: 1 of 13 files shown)


app/frontend/src/components/forms/submission/StatusPanel.vue
```diff
@@ -369,14 +361,21 @@ export default {
...
+            }
+
             const submissionStatusId =
               statusResponse.data[0].submissionStatusId;
             const user = await rbacService.getCurrentUser();
             const noteBody = {
               submissionId: this.submissionId,
               submissionStatusId: submissionStatusId,
-              note: this.note,
+              emailComment: formattedComment,
```
jujaga, line 378: Should the note attribute be changing to emailComment? FormService.addNote focuses on notes, of which formattedComment just happens to be for email context; but it's still a note.

app/src/forms/submission/service.js
```diff
@@ -164,9 +164,9 @@ const service = {
         id: uuidv4(),
         submissionId: submissionId,
         submissionStatusId: data.submissionStatusId,
-        note: data.note,
+        note: data.note ? data.note : data.emailComment,
```
jujaga, line 167: We shouldn't be needing to tweak this file at all - a note is a note, no matter what context it comes from. Changing the API shape like this is bad practice.

jujaga (commented):
Avoid changing the API spec shape - there should be only a few changes in the backend at most to support this new feature.

kamorel (approved):

jujaga (approved):

PR #283, opened 2022-04-19
Add history of actions in designer and ability to undo/redo actions

jasonchung1871:
# Description

When the form is mounted or when a form is fetched, we cache the original schema. When performing an action, we add that action to history of actions in the FormDesigner. These actions include importing a schema, adding, removing, moving, and updating a component. Added an undo button and redo button to the FormDesigner toolbar, both buttons have a counter displaying how many times the user can perform an undo or redo operation. When pressing undo,  we go back one action and generate a form schema from the original schema with all patches up until the previous action and set the form schema to the generated one. When pressing redo,  we go forward one action and generate a form schema from the original schema with all patches up until the next action and set the form schema to the generated one. When adding an action to the history, if we're not at the end of the history, we remove all actions after the current action.

## Types of changes

- Undo button, reverts to previous action in history
- Redo button, reverts to next action in history

## Checklist

## Further comments

Something that I could do is add tests to prove that the feature works.

### DIFF
--- a/app/frontend/src/components/designer/FormDesigner.vue
+++ b/app/frontend/src/components/designer/FormDesigner.vue
@@ -11,6 +11,21 @@
           <template #activator="{ on, attrs }">
             <v-btn
               class="mx-md-1 mx-0"
+              @click="onRefresh"
+              color="primary"
+              icon
+              v-bind="attrs"
+              v-on="on"
+            >
+              <v-icon>refresh</v-icon>
+            </v-btn>
+          </template>
+          <span>Refresh</span>
+        </v-tooltip>
+        <v-tooltip bottom>
+          <template #activator="{ on, attrs }">
+            <v-btn
+              class="mx-md-1 mx-0"
               @click="submitFormSchema"
               color="primary"
               icon
@@ -393,7 +408,15 @@ export default {
           this.formSchema = JSON.parse(fileReader.result);
           // Key-changing to force a re-render of the formio component when we want to load a new schema after the page is already in
           this.reRenderFormIo += 1;
-          this.originalSchema = JSON.parse(JSON.stringify(_.cloneDeep(this.formSchema)));
+          this.addPatchToHistory();
+          this.patchHistoryUndo = false;
+          this.patchHistoryRedo = false;
+          this.componentAdded = false;
+          this.componentRemoved = false;
+          this.componentUpdated = false;
+          this.componentMoved = false;
+          this.undoEnabled = this.canUndoPatch();
+          this.redoEnabled = this.canRedoPatch();
         });
         fileReader.readAsText(file);
       } catch (error) {
@@ -588,6 +611,9 @@ export default {
     },
     async onRedoClick() {
       this.redoPatchFromHistory();
+    },
+    async onRefresh() {
+      this.reRenderFormIo += 1;
     },
     async schemaCreateNew() {
       const emailList =
--- a/app/frontend/src/components/designer/FormDesigner.vue
+++ b/app/frontend/src/components/designer/FormDesigner.vue
@@ -181,15 +181,15 @@
       @change="onChangeMethod"
       @render="onRenderMethod"
       @initialized="init"
-      @addComponent="onAddComponentMethod"
-      @removeComponent="onRemoveComponentMethod"
-      @updateComponent="onUpdateComponentMethod"
+      @addComponent="onAddSchemaComponent"
+      @removeComponent="onRemoveSchemaComponent"
       class="form-designer"
     />
   </div>
 </template>
 
 <script>
+import { compare, applyPatch, deepClone } from 'fast-json-patch';
 import { mapActions, mapGetters } from 'vuex';
 import { FormBuilder } from 'vue-formio';
 import { mapFields } from 'vuex-map-fields';
@@ -198,9 +198,6 @@ import templateExtensions from '@/plugins/templateExtensions';
 import { formService } from '@/services';
 import { IdentityMode, NotificationTypes } from '@/utils/constants';
 import { generateIdps } from '@/utils/transformUtils';
-
-var deepFreeze = require('deep-freeze');
-import { compare, applyPatch, deepClone } from 'fast-json-patch';
 
 export default {
   name: 'FormDesigner',
@@ -216,7 +213,6 @@ export default {
     },
     versionId: String,
   },
-  originalSchema: null,
   data() {
     return {
       advancedItems: [
@@ -232,18 +228,17 @@ export default {
       displayVersion: 1,
       reRenderFormIo: 0,
       saving: false,
-      patchHistory: [],
-      maxPatches: 10,
-      patchIndex: -1,
-      patchHistoryUndo: false,
-      patchHistoryRedo: false,
-      patchObserver: null,
-      undoEnabled: false,
-      redoEnabled: false,
-      componentAdded: null,
-      componentRemoved: false,
-      componentUpdated: false,
-      componentMoved: false,
+      patch: {
+        componentAddedStart: false,
+        componentRemovedStart: false,
+        componentMovedStart: false,
+        history: [],
+        index: -1,
+        MAX_COUNT: 10,
+        redoClicked: false,
+        undoClicked: false,
+        originalSchema: null,
+      },
     };
   },
   computed: {
@@ -355,10 +350,16 @@ export default {
       };
     },
     undoCount() {
-      return (this.patchHistory.length > 0) ? (this.patchIndex + 1) : 0;
+      return this.patch.history.length > 0 ? this.patch.index + 1 : 0;
     },
     redoCount() {
-      return ((this.patchHistory.length > 0 ? this.patchHistory.length - this.patchIndex - 1 : 0));
+      return this.patch.history.length > 0 ? this.patch.history.length - this.patch.index - 1 : 0;
+    },
+    undoEnabled() {
+      return this.canUndoPatch();
+    },
+    redoEnabled() {
+      return this.canRedoPatch();
     },
   },
   methods: {
@@ -376,11 +377,10 @@ export default {
           res = await formService.readDraft(this.formId, this.draftId);
         }
         this.formSchema = { ...this.formSchema, ...res.data.schema };
-        if (this.patchHistory.length === 0) {
+        if (this.patch.history.length === 0) {
           // We are fetching an existing form, so we get the original schema here because
           // using the original schema in the mount will give you the default schema
-          this.$options.originalSchema = JSON.parse(JSON.stringify(deepFreeze(deepClone(this.formSchema))));
-          deepFreeze(this.$options.originalSchema);
+          this.patch.originalSchema = deepClone(this.formSchema);
         }
       } catch (error) {
         this.addNotification({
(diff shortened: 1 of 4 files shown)


bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-283

app/frontend/src/components/designer/FormDesigner.vue
```diff
@@ -369,15 +443,135 @@ export default {
     // ---------------------------------------------------------------------------------------------------
     init() {
       // Since change is triggered during loading
-      this.setDirtyFlag(false);
     },
-    onChangeMethod() {
+    onChangeMethod(_changed, _flags, _modified) {
```
jujaga, line 447: `onChangeMethod` is a bit ambiguous as to what event it is handling, and for what object/context. I would propose creating another method called something like `onSchemaChange` as it makes it more clear the intent of this method (do something with the schema when it changes), and  then calling that function within the `onChangeMethod` function. IIRC `onChangeMethod` is supposed to mirror/override the internal form.io event hooks, so it would be safer to functionalize out our history changes into its own invokable method instead.

app/frontend/src/components/designer/FormDesigner.vue
```diff
@@ -369,15 +443,135 @@ export default {
...
+        } else {
+          if (this.componentRemoved) {
+            // Component was removed..
+            this.addPatchToHistory();
+          } else if (this.componentMoved) {
+            this.addPatchToHistory();
+          } else {
+            this.resetHistoryFlags();
+          }
+        }
```
jujaga, line 470: - Double check the new variables such as componentAdded - do we need to be initalizing it as null, or can we keep it as just a simpler boolean? If we can do the latter, it can simplify the logic needed.
- This entire section of logic seems like it doesn't need to be branched because in most situations, you're either just calling `addPatchToHistory` or `resetHistoryFlags` methods when something changes (or not). You could potentially simplify this entire block to the following, ignoring what kind of change it was, and ignoring the `_flags` and `_modified` too (pseudocode... don't know if there's other side effects to worry about?):

```js
if (this.componentAdded || this.componentMoved || this.componentRemoved) {
  this.addPatchToHistory();
} else {
  this.resetHistoryFlags();
}
```

app/frontend/src/components/designer/FormDesigner.vue
```diff
@@ -369,15 +443,135 @@ export default {
...
+      var form = this.getPatch(++this.patchIndex);
+      var patch = compare(form, this.formSchema);
+      // Add the patch to the history
+      this.patchHistory.push(patch);
+
+      this.resetHistoryFlags();
+    },
+    getPatch(_idx) {
+      // Generate the form from the original schema
+      var form = JSON.parse(JSON.stringify(this.$options.originalSchema));
```
jujaga, line 528: - Do not use var. Declare with either const or let only.
- If I understand this correctly, you can use deepClone to achieve the same thing instead of stringify/parse chaining it.

app/frontend/src/components/designer/FormDesigner.vue
```diff
@@ -369,15 +443,135 @@ export default {
...
+    addPatchToHistory() {
+      // Remove any actions past the action we were on
+      if (this.patchHistory.length > 0) {
+        this.patchHistory.length = this.patchIndex + 1;
+      }
+
+      // Get the differences between the last patch 
+      // and the current form
+      var form = this.getPatch(++this.patchIndex);
+      var patch = compare(form, this.formSchema);
```
jujaga, line 520: Do not use var. Declare with either const or let only.

app/frontend/src/components/designer/FormDesigner.vue
```diff
@@ -369,15 +443,135 @@ export default {
...
+      this.patchHistory.push(patch);
+
+      this.resetHistoryFlags();
+    },
+    getPatch(_idx) {
+      // Generate the form from the original schema
+      var form = JSON.parse(JSON.stringify(this.$options.originalSchema));
+      if (this.patchIndex > -1 && this.patchHistory.length > 0) {
+        // Apply all patches until we reach the requested patch
+        for (var i = -1; i < _idx; i++) {
```
jujaga, line 531: Do not use var. Declare with either const or let only.

app/frontend/src/components/designer/FormDesigner.vue
```diff
@@ -369,15 +443,135 @@ export default {
...
+      // Get the differences between the last patch 
+      // and the current form
+      var form = this.getPatch(++this.patchIndex);
+      var patch = compare(form, this.formSchema);
+      // Add the patch to the history
+      this.patchHistory.push(patch);
+
+      this.resetHistoryFlags();
+    },
+    getPatch(_idx) {
```
jujaga, line 526: `_idx` shouldn't be underscored as it is actually being used in the body, that should not be prefixed with an underscore as that is contrary to the purposefully "unused" argument variable declaration. It should just be `idx` instead.

app/frontend/src/components/designer/FormDesigner.vue
```diff
@@ -369,15 +443,135 @@ export default {
...
+        this.patchHistoryRedo = true;
+        this.patchIndex++;
+        this.formSchema = this.getPatch(this.patchIndex);
+      }
+    },
+    canUndoPatch() {
+      if (this.patchHistory.length == 0) return false;
+      if (this.patchIndex > this.patchHistory.length) return false;
+      if (this.patchIndex < 0) return false;
+      return true;
```
jujaga, line 563:
```js
return this.patchHistory.length && this.patchIndex >= 0 && this.patchIndex < this.patchHistory.length;
```

app/frontend/src/components/designer/FormDesigner.vue
```diff
@@ -334,6 +399,15 @@ export default {
           this.formSchema = JSON.parse(fileReader.result);
           // Key-changing to force a re-render of the formio component when we want to load a new schema after the page is already in
           this.reRenderFormIo += 1;
+          this.addPatchToHistory();
+          this.patchHistoryUndo = false;
+          this.patchHistoryRedo = false;
+          this.componentAdded = false;
+          this.componentRemoved = false;
+          this.componentUpdated = false;
+          this.componentMoved = false;
```
jujaga, line 408: You already have a `resetHistoryFlags` method defined. You should be using reusing that.

```js
this.resetHistoryFlags();
```

app/frontend/src/components/designer/FormDesigner.vue
```diff
@@ -413,6 +607,12 @@ export default {
         this.saving = false;
       }
     },
+    async onUndoClick() {
+      this.undoPatchFromHistory();
+    },
+    async onRedoClick() {
+      this.redoPatchFromHistory();
+    },
```
jujaga, line 615: Not all methods need to be declared as async functions. For these ones, you can make these just normal functions. In general, you only want to make these methods async if and only if:

1. The method body has at least one `await`
2. The return result is expected to be of Promise type.
3. You know the methods invoked are asynchronous in nature and cannot behave synchronously.

Double check the rest of your new methods as well.

app/frontend/src/components/designer/FormDesigner.vue
```diff
@@ -477,11 +677,18 @@ export default {
       this.fetchForm(this.formId);
     }
   },
+  mounted() {
+    if (!this.formId) {
+      // We are creating a new form, so we obtain the original schema here.
+      this.$options.originalSchema = JSON.parse(JSON.stringify(deepFreeze(deepClone(this.formSchema))));
+      deepFreeze(this.$options.originalSchema);
```
jujaga, line 684: If we are already deepCloning an object, there is no need to do a JSON stringify/parse chain as it is already doing the exact same thing. If I understand the intent of code, this can be a one-liner:

```js
this.$options.originalSchema = deepFreeze(deepClone(this.formSchema));
```

jujaga (changes requested):
This is quite an interesting feature addition. Fantastic work! 👍 
That being said, there are a good handful of code style and considerations I'd like for you to take a look at. Please take a look at the detailed comments below and address them when you get a chance.

app/frontend/src/components/designer/FormDesigner.vue
jasonchung1871: Ah, the reason for this was because something kept making the originalSchema observable and changing it even though it should be read only. I think I found that it was occurring when generating the final schema in the getPatch method.

app/frontend/src/components/designer/FormDesigner.vue
```diff
@@ -176,6 +216,7 @@ export default {
     },
     versionId: String,
   },
+  originalSchema: null,
```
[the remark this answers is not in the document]
jasonchung1871: i could not for the life of me figure out why the originalSchema variable was being modified even though it was made to be read-only with deepFreeze, i've figured that out now though. i was trying out many different things to try and just prevent that variable from being modified after loading the schema.

app/frontend/src/components/designer/FormDesigner.vue
```diff
@@ -162,6 +199,9 @@ import { formService } from '@/services';
 import { IdentityMode, NotificationTypes } from '@/utils/constants';
 import { generateIdps } from '@/utils/transformUtils';
 
+var deepFreeze = require('deep-freeze');
+import { compare, applyPatch, deepClone } from 'fast-json-patch';
```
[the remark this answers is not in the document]
jasonchung1871: the last 3 imports are not mine, i just went with what was already there. i'll change them though

app/frontend/src/components/designer/FormDesigner.vue
```diff
@@ -317,6 +376,12 @@ export default {
           res = await formService.readDraft(this.formId, this.draftId);
         }
         this.formSchema = { ...this.formSchema, ...res.data.schema };
+        if (this.patchHistory.length === 0) {
+          // We are fetching an existing form, so we get the original schema here because
+          // using the original schema in the mount will give you the default schema
+          this.$options.originalSchema = JSON.parse(JSON.stringify(deepFreeze(deepClone(this.formSchema))));
+          deepFreeze(this.$options.originalSchema);
```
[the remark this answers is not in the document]
jasonchung1871: i've figured out what area was causing the original schema to be modified and have removed the deepFreeze as well!

app/frontend/src/components/designer/FormDesigner.vue
jasonchung1871: first off, sorry for the wall of text, but this is for anyone else that may run into a similar issue!

the first one was a mistake on my part, initially i was storing the component in the componentAdded variable so that i could store the new component in the history after the user clicks "save". i've reverted it back to a boolean and now it is much cleaner, simpler, and unpacks/repacks correctly since we're using fast-json-patch.

the reason why it isn't simplified is because i wasn't sure at first how complex it would become. formio calls the change method for absolutely every change even if the user doesn't click save or if they click cancel/remove. we unfortunately need the `flags` and `modified` parameters to verify that the user clicked save and actually added the component to the form. i also had to create my own flags for when we started an action such as add, remove, move, and update. those flags are set to true when the actions begin in their respective functions `onAddComponentMethod`, `onRemoveComponentMethod`, `onUpdateComponentMethod` and then set to false once an action is completed.

for example, the save process alone looks like:
drag and drop -> `onUpdateComponentMethod()` -> `onChangeMethod(changed, undefined, undefined)` -> `onAddComponent()`-> `onUpdateComponentMethod()` -> `onChangeMethod(changed, undefined, undefined)`
save -> `onChangeMethod(changed, flags, modified)` -> `onChangeMethod(changed, undefined, undefined)`
canceled after drag and drop -> `onRemoveComponentMethod()` -> `onChangeMethod(changed, undefined, undefined)`
removed after drag and drop -> `onRemoveComponentMethod()` -> `onChangeMethod(changed, undefined, undefined)`

after writing all this, i realized we don't really need fine-grained control over the actions since we're not writing the json comparator/patcher. i've removed the `onUpdateComponentMethod` method since an update is similar to an "add" with the `flags` and `modified` parameters defined. if we did want fine-grained control, then we could obtain the information for exactly what is being added/updated/removed and know if the action was saved or canceled/removed.

TimCsaky:
just tried in the browser..  looking good!
minor point, the 'redo' button is active after adding the first component to a blank form. might have something to do with the default data values?

jasonchung1871:
> just tried in the browser.. looking good! minor point, the 'redo' button is active after adding the first component to a blank form. possibly some incrementing issues. might have something to do with the default data values?

thank you! this was an issue on my part, i removed the increment to the history index when testing the limit on max actions in the history. i've added the increment back in, and all is well!

jujaga (approved):

PR #284, opened 2022-04-20
created custom component and hosted json data files

timisenco2015:
Created
    - FieldsAddress
    - Country
    - City
    - Province
 Created and hosted using json-server
     -  AB.json
     - BC.json
     - citiesandtowns
     - MB.json
     - NB.json
     - NF.json
     - NS.json
     - NT.json
     - NU.json
     - ON.json
     - PE.json
     - provinces.json
     - QC.json
     - SK.json
     - YT.json
    To have select options populated for all the components (FieldsAddress, City, Country, and Province), you will need to run the JSON server;
    - open the command prompt on windows
    - npm install JSON server 
    - cd to common-hosted-form-service\app\countrydataserver
    - enter node server.js
    - must be running on port 3000

# Description

## Checklist


jujaga (changes requested):
We cannot be introducing side-car programs to host static data as that is a complete anti-pattern to containerized application design; not to mention that side-car program would never be invoked in the container to begin with from a quick glance.

As discussed previously, while form.io custom components are definitely scalable and customizeable, the way they are currently managed and implemented in CHEFS does not allow this to be permutable per instance yet as it must be defined, imported and built during compile time.

Unfortunately I'm going to have to block this PR in the interim; aside from the above architectural issues, there are too many impactful structural changes, "hello world" artifact code, TODOs and unnecessary console.log statements that can potentially break existing form functionality for existing clients.

timisenco2015:
> We cannot be introducing side-car programs to host static data as that is a complete anti-pattern to containerized application design; not to mention that side-car program would never be invoked in the container to begin with from a quick glance.
> 
> As discussed previously, while form.io custom components are definitely scalable and customizeable, the way they are currently managed and implemented in CHEFS does not allow this to be permutable per instance yet as it must be defined, imported and built during compile time.
> 
> Unfortunately I'm going to have to block this PR in the interim; aside from the above architectural issues, there are too many impactful structural changes, "hello world" artifact code, TODOs and unnecessary console.log statements that can potentially break existing form functionality for existing clients.

I have removed the side-car. I thought that the best option was to have the static data in the folder with the source. I actually thought of adding json-server to docker.

I will forward this feedback to my team for review.

The "Hello world" artifact code, TODOs and console.log are from my source code.

jujaga:
Closing PR for now - can revisit these concepts down the line.

PR #286, opened 2022-05-05
Administrative section to add an owner

loneil:
# Description
https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-2616

Add a section in the [CHEFS Admin portal](https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-286/admin/) (for operational admins, needs the showcase team KC group to access) to add additional owners to a form.
Should only be used in the case where a form team person contacts us needing to add someone else to manage their forms when the original owner is no longer reachable. Otherwise they should use self-serve team management features if possible.

**Frontend**
From admin go into the form and at the bottom there's this form
image
Could be a user search but to keep it light just having the admin enter the direct user primary key. This feature is just being used by common service team admins and they can easily find the user id on the USER tab in the admin section for the user they need to add (assuming they've logged into CHEFS once).

**Backend**
New admin API call.
See spec:  https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-286/api/v1/docs#operation/adminAddUser

## Types of changes

## Checklist

### DIFF
--- a/app/frontend/src/components/admin/AddOwner.vue
+++ b/app/frontend/src/components/admin/AddOwner.vue
@@ -5,26 +5,31 @@
     current Owner or a Team Administrator for the form do this themselves.
     {{ formId }} {{ userGuid }}
 
-    <v-row class="mt-4">
-      <v-col cols="9" sm="6" md="6" lg="4">
-        <v-text-field
-          label="User ID (guid)"
-          v-model="userGuid"
-          hint="To find the User ID needed you can go to the 'USERS' tab in this Admin section and search for them."
-          persistent-hint
-        />
-      </v-col>
-      <v-col cols="3" md="2">
-        <v-btn color="primary">
-          <span>Add owner</span>
-        </v-btn>
-      </v-col>
-    </v-row>
+    <v-form ref="addUserForm" v-model="valid">
+      <v-row class="mt-4">
+        <v-col cols="9" sm="6" md="6" lg="4">
+          <v-text-field
+            label="User ID (guid)"
+            v-model="userGuid"
+            :rules="userGuidRules"
+            hint="To find the User ID needed you can go to the 'USERS' tab in the Admin portal and search for them."
+            persistent-hint
+          />
+        </v-col>
+        <v-col cols="3" md="2">
+          <v-btn color="primary" @click="addOwner" :disabled="!valid">
+            <span>Add owner</span>
+          </v-btn>
+        </v-col>
+      </v-row>
+    </v-form>
   </div>
 </template>
 
 <script>
 import { mapActions } from 'vuex';
+
+import { Regex } from '@/utils/constants';
 
 export default {
   name: 'AddOwner',
@@ -37,10 +42,21 @@ export default {
   data() {
     return {
       userGuid: '',
+      valid: false,
+
+      userGuidRules: [
+        v => !!v || 'User ID required',
+        v => new RegExp(Regex.GUID).test(v) || 'Enter a valid User ID GUID',
+      ],
     };
   },
   methods: {
     ...mapActions('admin', ['addOwner']),
+    async addOwner() {
+      if (this.$refs.addUserForm.validate()) {
+        alert('pass');
+      }
+    },
   },
 };
 </script>
--- a/app/frontend/src/components/admin/AddOwner.vue
+++ b/app/frontend/src/components/admin/AddOwner.vue
@@ -1,34 +1,35 @@
 <template>
-  <div>
-    This should only be done in the event that the current form owner is no
-    longer active or is out of contact in a priority event. Otherwise have the
-    current Owner or a Team Administrator for the form do this themselves.
+  <v-form ref="addUserForm" v-model="valid">
+    <p>
+      This should only be done in the event that the current form owner is no
+      longer active or is out of contact in a priority event. Otherwise have the
+      current Owner or a Team Administrator for the form do this themselves.
+    </p>
 
-    <v-form ref="addUserForm" v-model="valid">
-      <v-row class="mt-4">
-        <v-col cols="9" sm="6" md="6" lg="4">
-          <v-text-field
-            label="User ID (guid)"
-            v-model="userGuid"
-            :rules="userGuidRules"
-            hint="To find the User ID needed you can go to the 'USERS' tab in the Admin portal and search for them."
-            persistent-hint
-          />
-        </v-col>
-        <v-col cols="3" md="2">
-          <v-btn color="primary" @click="addOwner" :disabled="!valid">
-            <span>Add owner</span>
-          </v-btn>
-        </v-col>
-      </v-row>
-    </v-form>
-  </div>
+    <v-row class="mt-4">
+      <v-col cols="9" sm="6" md="6" lg="4">
+        <v-text-field
+          label="User ID (guid)"
+          v-model="userGuid"
+          :rules="userGuidRules"
+          hint="To find the User ID needed you can go to the 'USERS' tab in the Admin portal and search for them."
+          persistent-hint
+        />
+      </v-col>
+      <v-col cols="3" md="2">
+        <v-btn color="primary" @click="addOwner" :disabled="!valid">
+          <span>Add owner</span>
+        </v-btn>
+      </v-col>
+    </v-row>
+  </v-form>
 </template>
 
 <script>
 import { mapActions } from 'vuex';
 
-import { FormRoleCodes, Regex } from '@/utils/constants';
+import { FormRoleCodes } from '@/utils/constants';
+import { version as uuidVersion, validate as uuidValidate } from 'uuid';
 
 export default {
   name: 'AddOwner',
@@ -45,7 +46,9 @@ export default {
 
       userGuidRules: [
         (v) => !!v || 'User ID required',
-        (v) => new RegExp(Regex.GUID).test(v) || 'Enter a valid User ID GUID',
+        (v) =>
+          (uuidValidate(v) && uuidVersion(v) === 4) ||
+          'Enter a valid User ID GUID',
       ],
     };
   },
@@ -56,9 +59,8 @@ export default {
         await this.addFormUser({
           userId: this.userGuid,
           formId: this.formId,
-          role: FormRoleCodes.OWNER,
+          roles: [FormRoleCodes.OWNER]
         });
-        this.readRoles(this.formId);
       }
     },
   },
--- a/app/frontend/src/components/admin/AddOwner.vue
+++ b/app/frontend/src/components/admin/AddOwner.vue
@@ -1,0 +1,46 @@
+<template>
+  <div>
+    This should only be done in the event that the current form owner is no
+    longer active or is out of contact in a priority event. Otherwise have the
+    current Owner or a Team Administrator for the form do this themselves.
+    {{ formId }} {{ userGuid }}
+
+    <v-row class="mt-4">
+      <v-col cols="9" sm="6" md="6" lg="4">
+        <v-text-field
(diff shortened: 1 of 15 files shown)


app/frontend/src/components/admin/AddOwner.vue
```diff
@@ -0,0 +1,66 @@
+<template>
+  <div>
+    This should only be done in the event that the current form owner is no
+    longer active or is out of contact in a priority event. Otherwise have the
+    current Owner or a Team Administrator for the form do this themselves.
+
+    <v-form ref="addUserForm" v-model="valid">
```
jujaga, line 7: Raw text should be wrapped in either a <p> element or similar on its own.
In regards to template nesting, I think you may be able to get away without needing a wrapping div on line 2 for the template if you did the following:
```html
<v-form ...>
  <p>This should only...</p>
  <v-row>
    <v-col>
```

app/frontend/src/components/admin/AddOwner.vue
```diff
@@ -0,0 +1,66 @@
...
+  methods: {
+    ...mapActions('admin', ['addFormUser', 'readRoles']),
+    async addOwner() {
+      if (this.$refs.addUserForm.validate()) {
+        await this.addFormUser({
+          userId: this.userGuid,
+          formId: this.formId,
+          role: FormRoleCodes.OWNER,
+        });
+        this.readRoles(this.formId);
```
jujaga, line 61: Should this be wrapped in a try/catch? In the event addFormUser fails (likely network issue at that point), may be good to not invoke readRoles and add more network call strain to backend?

app/frontend/src/services/adminService.js
```diff
@@ -6,6 +6,18 @@ export default {
   // Form calls
   //
 
+  /**
+   * @function addFormUser
+   * Add a form user role (specific administrative task, for real form/user/role management see rbac service)
+   * @param {formId} formId The request body for the relationships
+   * @param {userId} userId The request body for the relationships
+   * @param {Object} [roles] The list of roles too add for this user
```
jujaga, line 14:
```js
* @param {String[]} [roles] The list of roles to add
```

app/frontend/src/services/adminService.js
```diff
@@ -6,6 +6,18 @@ export default {
...
+   * @returns {Promise} An axios response
+   */
+  addFormUser(userId, formId, role) {
+    return appAxios().put(`${ApiRoutes.ADMIN}${ApiRoutes.FORMS}/${formId}/addUser`, { userId: userId, formId: formId, role: role });
+  },
```
jujaga, line 19: Function signature, jsdoc description and implementation are somewhat mismatched and misleading. There are two way to think of this service call:
1) We are only ever going to allow one user/form pairing happen for each call - can make it have multiple roles if we want. In this case, the implementation needs to be properly handling multiple roles...
```js
  addFormUser(userId, formId, roles) {
    return appAxios().put(`${ApiRoutes.ADMIN}${ApiRoutes.FORMS}/${formId}/addUser`, roles.map(role => ({
      userId: userId,
      formId: formId,
      role: role
    })));
  },
```
2) This service call can handle arbitrary user/form pairings, in addition to the multiple roles. In this case, addFormUser should just be `addFormUser(permissions)` where permissions is an array of those triplets and passes through wholesale instead.
I'd personally recommend for now going with option 1. Option 2 allows for more future flexibility, but the logic onus would be on the caller instead (likely the vuex method).

app/frontend/src/components/admin/AddOwner.vue
```diff
@@ -0,0 +1,66 @@
...
+    };
+  },
+  methods: {
+    ...mapActions('admin', ['addFormUser', 'readRoles']),
+    async addOwner() {
+      if (this.$refs.addUserForm.validate()) {
+        await this.addFormUser({
+          userId: this.userGuid,
+          formId: this.formId,
+          role: FormRoleCodes.OWNER,
```
jujaga, line 59: This should be an array:
```js
roles: [FormRoleCodes.OWNER]
```

app/frontend/src/store/modules/admin.js
```diff
@@ -46,6 +46,21 @@ export default {
     //
     // Forms
     //
+    async addFormUser({ dispatch }, formUser) {
+      try {
+        const response = await adminService.addFormUser(formUser.userId, formUser.formId, formUser.role);
```
jujaga, line 51: Role -> Roles (should be array)
```js
const response = await adminService.addFormUser(formUser.userId, formUser.formId, formUser.roles);
```

app/frontend/src/components/admin/AddOwner.vue
```diff
@@ -0,0 +1,66 @@
...
+    },
+  },
+  data() {
+    return {
+      userGuid: '',
+      valid: false,
+
+      userGuidRules: [
+        (v) => !!v || 'User ID required',
+        (v) => new RegExp(Regex.GUID).test(v) || 'Enter a valid User ID GUID',
```
jujaga, line 48: Should we be introducing the GUID check via a regex? We could be leveraging the existing `uuidValidate` and `uuidVersion` methods from the `uuid` library to achieve the same result.
Ref: https://www.npmjs.com/package/uuid#uuidvalidatestr

app/frontend/src/utils/constants.js
```diff
@@ -99,5 +99,7 @@ export const NotificationTypes = Object.freeze({
 
 export const Regex = Object.freeze({
   // From ajv-format
-  EMAIL: '^[a-zA-Z0-9!#$%&\'*+/=?^_`{|}~-]+(?:\\.[a-zA-Z0-9!#$%&\'*+/=?^_`{|}~-]+)*@(?:[a-zA-Z0-9](?:[a-zA-Z0-9-]*[a-zA-Z0-9])?\\.)+[a-zA-Z0-9](?:[a-zA-Z0-9-]*[a-zA-Z0-9])?$'
+  EMAIL: '^[a-zA-Z0-9!#$%&\'*+/=?^_`{|}~-]+(?:\\.[a-zA-Z0-9!#$%&\'*+/=?^_`{|}~-]+)*@(?:[a-zA-Z0-9](?:[a-zA-Z0-9-]*[a-zA-Z0-9])?\\.)+[a-zA-Z0-9](?:[a-zA-Z0-9-]*[a-zA-Z0-9])?$',
+  // guid
+  GUID: '^[{]?[0-9a-fA-F]{8}-([0-9a-fA-F]{4}-){3}[0-9a-fA-F]{12}[}]?$'
```
jujaga, line 104: Should we be introducing the GUID check via a regex? We could be leveraging the existing `uuidValidate` and `uuidVersion` methods from the `uuid` library to achieve the same result.
Ref: https://www.npmjs.com/package/uuid#uuidvalidatestr

app/frontend/tests/unit/utils/constants.spec.js
```diff
@@ -100,7 +100,8 @@ describe('Constants', () => {
 
   it('Regex has the right values defined', () => {
     expect(constants.Regex).toEqual({
-      EMAIL: '^[a-zA-Z0-9!#$%&\'*+/=?^_`{|}~-]+(?:\\.[a-zA-Z0-9!#$%&\'*+/=?^_`{|}~-]+)*@(?:[a-zA-Z0-9](?:[a-zA-Z0-9-]*[a-zA-Z0-9])?\\.)+[a-zA-Z0-9](?:[a-zA-Z0-9-]*[a-zA-Z0-9])?$'
+      EMAIL: '^[a-zA-Z0-9!#$%&\'*+/=?^_`{|}~-]+(?:\\.[a-zA-Z0-9!#$%&\'*+/=?^_`{|}~-]+)*@(?:[a-zA-Z0-9](?:[a-zA-Z0-9-]*[a-zA-Z0-9])?\\.)+[a-zA-Z0-9](?:[a-zA-Z0-9-]*[a-zA-Z0-9])?$',
+      GUID: '^[{]?[0-9a-fA-F]{8}-([0-9a-fA-F]{4}-){3}[0-9a-fA-F]{12}[}]?$',
```
jujaga, line 104: Should we be introducing the GUID check via a regex? We could be leveraging the existing `uuidValidate` and `uuidVersion` methods from the `uuid` library to achieve the same result.
Ref: https://www.npmjs.com/package/uuid#uuidvalidatestr

jujaga (commented):
Looks pretty spiffy. Two things to address:
1) Should we be introducing the UUIDv4 check via regex or should we be leveraging the `uuid` library more?
2) Make sure the frontend service layer implementation matches the intent of the jsdocs w.r.t. roles plurality and propagate to the rest of the stack accordingly.

As well, I saw in a lot of the unit tests that we were just doing a bunch of `expect.any(Object)`. While this is fine, the tests would have more precision and value if we were more specific with those argument expectations whenever possible (so if we know the value, we test on that instead; if we expect it to be an array, do an `expect.any(Array)` instead, etc).

app/frontend/tests/unit/utils/constants.spec.js
loneil: Used uuid instead

app/frontend/src/utils/constants.js
loneil: Used uuid instead

app/frontend/src/components/admin/AddOwner.vue
loneil: Used uuid instead

app/frontend/src/components/admin/AddOwner.vue
loneil: Fixed

app/frontend/src/components/admin/AddOwner.vue
loneil: Moved call to dispatch in the same action instead (handled by that try catch)

app/frontend/src/services/adminService.js
loneil: fixed

app/frontend/src/services/adminService.js
loneil: I've modified it to support multiple roles as in 1

app/frontend/src/store/modules/admin.js
loneil: roles now

bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-286

jujaga (approved):

app/frontend/src/services/adminService.js
```diff
@@ -6,6 +6,22 @@ export default {
...
+   * @param {userId} userId The request body for the relationships
+   * @param {String[]} [roles] The list of roles to add
+   * @returns {Promise} An axios response
+   */
+  addFormUser(userId, formId, roles) {
+    return appAxios().put(`${ApiRoutes.ADMIN}${ApiRoutes.FORMS}/${formId}/addUser`, roles.map(role => ({
+      userId: userId,
+      formId: formId,
+      role: role
+    })), { params: { userId: userId } });
```
jujaga, line 22: For that new CHEFS endpoint, if the formId is required via path, and the query parameter is requiring a userId, then the body should just be an array of roles instead of the current array of formId, role and userId tuples.

jujaga: Discovered this limitation/behavior with query param being required came from the previous Objection 2.x to 3.x migration where querying with an undefined array would cause Objection to error out in the newer version. Decided for now to leave this the way its implemented in this PR and to update spec as needed to reflect current implemented behavior. Will revisit this at a later time.

jujaga (approved):

TimCsaky (approved):

PR #274, opened 2022-02-09
Remove submission list headers if their name was changed

kamorel:
# Description
Removes submission list headers if their name was changed in a newer form version.
[https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-2377](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-2377)
This will also remove the invalid column from the stored preferences the next time the user updates their prefs.

## Types of changes

Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] I have read the [CONTRIBUTING](CONTRIBUTING.md) doc
- [x] I have checked that unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)

PR #290, opened 2022-06-21
Can't populate submission table until form fields are known

loneil:
# Description
Fixes https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-2675

Refer to comments here https://github.com/bcgov/common-hosted-form-service/pull/274#pullrequestreview-878089635
The populate form part is actually synchronously dependent on the formFields list after the change from that PR, so it has to await on that. Might be a better way to lay that out than the 2 awaits but it works.

Also fix https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-2699
If computing something to determine a truthy value to use in the template, need to make it a computed value or invoke the method response. Can't do `v-if="methodName"` or that's truthy always.

## Types of changes

## Checklist

### DIFF
--- a/app/frontend/src/components/forms/SubmissionsTable.vue
+++ b/app/frontend/src/components/forms/SubmissionsTable.vue
@@ -175,9 +175,7 @@ export default {
         this.userFormPreferences.preferences.columnList
       ) {
         // Compare saved user prefs against the current form versions component names and remove any discrepancies
-        return this.userFormPreferences.preferences.columnList.filter(
-          (x) => this.formFields.indexOf(x) !== -1
-        );
+        return this.userFormPreferences.preferences.columnList.filter((x) => this.formFields.indexOf(x) !== -1);
       } else {
         return [];
       }
@@ -235,7 +233,7 @@ export default {
   },
 
   async mounted() {
-    // Get the form, latest version form fields, and fill the table
+    // Get the form, latest version form fields
     // These calls have to be waited on until the table can populate below
     await this.fetchForm(this.formId);
     await this.fetchFormFields({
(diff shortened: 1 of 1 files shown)


app/frontend/src/components/forms/SubmissionsTable.vue
```diff
@@ -232,13 +232,13 @@ export default {
...
-        formVersionId: this.form.versions[0].id,
-      });
+  async mounted() {
+    // Get the form, latest version form fields
+    // These calls have to be waited on until the table can populate below
+    await this.fetchForm(this.formId);
+    await this.fetchFormFields({
+      formId: this.formId,
+      formVersionId: this.form.versions[0].id,
     });
```
jujaga, line 242: It would be worthwhile to change this to the following promise-chain pattern in order to maximize parallelization:
```js
mounted() {
  Promise.all([
    this.getFormPermissionsForUser(this.formId),
    this.fetchForm(this.formId).then(() => {
      this.fetchFormFields({
        formId: this.formId,
        formVersionId: this.form.versions[0].id,
      });
    }),
  ]).then(() = {
    this.populateSubmissionsTable();
  });
}
```
To my understanding, `populateSubmissionsTable` needs all of the above previous things to have resolved first before proceeding. However, `fetchFormFields` depends on `fetchForm`, but NOT on `getFormPermissionsForUser`. So you want to make sure that `getFormPermissionsForUser` can run independently, and NOT block `fetchFormFields` nor `fetchForm` from running at the same time.

jujaga (approved):

bcgov-nr-csst:
PR Deployed at: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-290

TimCsaky (approved):

PR #311, opened 2022-10-20
Fix pr deployments

jasonchung1871:
# Description

When checking the actual PR deployment, we can see in the console an Unexpected token '<' error in the file app.e898c519.js . When trying to open that js file we find that inside contains some HTML instead of JS.

In the network tab, the requested URL is from https://chefs-dev.apps.silver.devops.gov.bc.ca/app/js/app.e898c519.js it should be https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-308/app/js/app.e898c519.js. It’s trying to load a file that does not exist.

The issue stopping us from using the path submit.digital.gov.bc.ca/pr-308 instead of submit.digital.gov.bc.ca/app is because we can't get that base/public path since it's happening before the config file is loaded. The best alternative I could find was to use an environment variable for this instance.

## Types of changes

## Checklist


vijaivir (approved):
Nice work. I think it's good to go

PR #327, opened 2022-10-25
Ayobami bcgeo address autocomplete

bcgov-citz-ccft:
Update FormDesigner.vue

# Description
Switch form.io address component from the Advanced Data Layout to BC Government Layout in FormDesigner.vue

## Types of changes

## Checklist

### DIFF
--- a/app/frontend/src/components/designer/FormDesigner.vue
+++ b/app/frontend/src/components/designer/FormDesigner.vue
@@ -332,11 +332,13 @@ export default {
           data: {
             title: 'Advanced Data',
             weight: 50,
+            address:false,
           },
           customControls: {
             title: 'BC Government',
             weight: 60,
             components: {
+              address:true,
               orgbook: true,
               simplefile: this.userType !== this.ID_MODE.PUBLIC,
             },


matthewhall78 (changes requested):
I don't think we should adding the default address component to the BC Government customControls.  I think we need to create a new custom component called something like "BC Address Autocomplete".   
Also I don't know why you would need to add the line 
```address:false,```
to the Advanced Data section. it seems to me that it is already excluded by default.

bcgov-citz-ccft:
> I don't think we should adding the default address component to the BC Government customControls. I think we need to create a new custom component called something like "BC Address Autocomplete". Also I don't know why you would need to add the line `address:false,` to the Advanced Data section. it seems to me that it is already excluded by default.



> I don't think we should adding the default address component to the BC Government customControls. I think we need to create a new custom component called something like "BC Address Autocomplete". Also I don't know why you would need to add the line `address:false,` to the Advanced Data section. it seems to me that it is already excluded by default.

I think it appears in the 'Advanced Data Section' by default.  For it not to appear in both the Advanced Data section and 'BC Government section' I set it to false there. Andy requested I switch the component to the 'BC Government section'

jasonchung1871 (approved):

PR #331, opened 2022-11-01
Feat forms 416

ptugger:
# Description
This PR Contains these features So far:
1. Allow Propagation Value - Checkbox 
- Added a checkbox to Advanced and Basic Fields (Under Validation tab) that allow form designer to choose a component used for Propagation value when Create new submition from the existing one.
- Story : https://bcdevex.atlassian.net/wiki/spaces/CCP/pages/1041432625/FORMS-321+Form+Designer+define+which+field+can+have+their+value+re-used+in+a+new+submission

2. Ability to create new submition from the existing one
- User can select previous submission to copy data from, And create a new submission
- Story : https://bcdevex.atlassian.net/wiki/spaces/CCP/pages/997949551/FORMS-214+Submitter+Create+Save+a+pre-populated+template+for+a+given+form

3. Form submission period setting
- This feature allow form designer to setup a form for specific period to be published.
- Story : https://bcdevex.atlassian.net/wiki/spaces/CCP/pages/1026195457/FORMS-289+Form+Designer+Setup+Form+Submission+Period+Setting

## Types of changes

## Checklist


app/frontend/src/utils/transformUtils.js
```diff
@@ -64,3 +65,203 @@ export function attachAttributesToLinks(formSchemaComponents) {
...
+                formSchedule.keepOpenForTerm,
+                formSchedule.keepOpenForInterval,
+                startDate,
+                formSchedule.repeatSubmission.everyTerm,
+                formSchedule.repeatSubmission.everyIntervalType,
+                formSchedule.allowLateSubmissions.enabled ? formSchedule.allowLateSubmissions.forNext.term : 0,
+                formSchedule.allowLateSubmissions.forNext.intervalType,
+                formSchedule.repeatSubmission.repeatUntil
+              );
+              for (let i = 0; i < availableDates.length; i++) {
```
jasonchung1871, line 122: not really an issue since it passed the lint, but the spacing is not consistent before your open/close brackets

app/src/db/migrations/20220822235439_025-alter-form-table-to-add-schedule-col.js
```diff
@@ -0,0 +1,15 @@
+exports.up = function(knex) {
+  return Promise.resolve()
+  // create new collumn to store Schedule (Period) setting
+    .then(() => knex.schema.alterTable('form', table => {
+      table.jsonb('schedule').comment('Form level Schedule settings.');
```
jasonchung1871, line 5: is there a reason why you decided to use json for the schedule field instead of regular fields?

jasonchung1871 (commented):
Please update the description with documentation so we know what we're trying to review

app/src/db/migrations/20220822235439_025-alter-form-table-to-add-schedule-col.js
ptugger: Because Schedule Object will be deep and big so It's better to save it in a JSON. Do you recommend something else?

app/frontend/src/utils/transformUtils.js
ptugger: Not sure Why is that showing there but when I did commit it was look good. More over on my local as well did not see any Lint Issues

jasonchung1871 (approved):

app/src/db/migrations/20220822235439_025-alter-form-table-to-add-schedule-col.js
timisenco2015: Can we have a different table for the schedule?

ptugger: Like just for the Schedule Settings? If Yes, Then I will have to take a look then can answer.

timisenco2015: I would suggest that this table.jsonb('schedule').comment('Form level Schedule settings.'); should be in a different table since it is a 1:Many relationship.

components/src/components/Common/Constants.ts
```diff
@@ -1,3 +1,4 @@
 export abstract class Constants {
   static readonly DEFAULT_HELP_LINK: string = 'https://github.com/bcgov/common-hosted-form-service/wiki';
+  static readonly ADV: string = '';
```
timisenco2015, line 3: I noticed that Constants.ADV was used in all the custom components but is set to an empty string here. Is there any reason for this?

PR #422, opened 2022-12-05
 an alternative option of generating pdf to cdog 

timisenco2015:
# Description
Form reviewers can use this feature to generate submissions into pdf. This is an alternative option to CDOG, and it does not require a template

## Types of changes

## Checklist


app/src/forms/submission/service.js
```diff
@@ -10,6 +9,7 @@ const {
   SubmissionAudit,
   SubmissionMetadata
 } = require('../common/models');
+const {reArrangeFormSubmissionJSon} = require('../common/utils');
```
jasonchung1871, line 12: The imported value should have spacing around its name to keep things consistent with everything else. const { ... } instead of {...}

PR #620, opened 2023-03-27
remove formId from multi submissions delete API

timisenco2015:
# Description
- remove formId from multi submissions delete API
- rename canDeleteAndRestoreMultiSubmissions middleware to filterMultipleSubmission
- rewrote unit test for multiple submissions delete

## Types of changes

## Checklist

### DIFF
--- a/app/frontend/src/store/modules/form.js
+++ b/app/frontend/src/store/modules/form.js
@@ -350,9 +350,9 @@ export default {
     },
 
 
-    async deleteMultiSubmissions ({ dispatch }, submissionIds) {
+    async deleteMultiSubmissions ({ dispatch },{formId, submissionIds}) {
       try {
-        await formService.deleteMultipleSubmissions(submissionIds[0], { data: {submissionIds:submissionIds} });
+        await formService.deleteMultipleSubmissions(submissionIds[0], formId, { data: {submissionIds:submissionIds} });
         dispatch('notifications/addNotification', {
           message: 'Submissions deleted successfully.',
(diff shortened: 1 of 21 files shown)


app/frontend/src/store/modules/form.js
```diff
@@ -350,9 +350,9 @@ export default {
     },
 
 
-    async deleteMultiSubmissions ({ dispatch },{formId, submissionIds}) {
+    async deleteMultiSubmissions ({ dispatch }, {submissionIds}) {
```
bcvesalink, line 353: I think it's still ok to pass `formId` to check if user have access to form in first place, we should investigate why `canDeleteAndRestoreMultiSubmissions` https://github.com/bcgov/common-hosted-form-service/pull/620/files#diff-c94e18e8dacd854755475f141756a44eb0e9567aabf5df871d930c30a4c4d78eL139 doesn't work as needed...

timisenco2015: I added the canDeleteAndRestoreMultiSubmissions after the bug was raised. I renamed that this morning to filterMultipleSubmission. I will add formId back

bcvesalink:
@timisenco2015 @jasonchung1871 do you guys know why we checking `Basic` auth here https://github.com/bcgov/common-hosted-form-service/blob/master/app/src/forms/auth/middleware/apiAccess.js#L9 shouldn't we check `Bearer` so we deny any access to API if no token provided in a header before going even further?

PR #668, opened 2023-04-17
Ability to select submission fields to export to csv 

timisenco2015:
# Description
This feature allows users to select the submission fields they want to export to CSV

## Types of changes

## Checklist


app/frontend/src/views/form/Export.vue
```diff
@@ -0,0 +1,25 @@
...
+<script>
+import ExportSubmissions from '@/components/forms/ExportSubmissions.vue';
+import { IdentityProviders } from '@/utils/constants';
+export default {
+  name: 'Export',
+  components: {
+    ExportSubmissions,
+  },
+  props: {
+    f: {
```
bcvesalink, line 16: can you change the prop name to be more meaningful, too short

timisenco2015: Okay. I was following the pattern in the codes

app/frontend/src/router/index.js
```diff
@@ -127,6 +127,20 @@ export default function getRouter(basePath = '/') {
...
+          {
+            path: 'export',
+            name: 'SubmissionsExport',
+            component: () =>
+              import(
+                /* webpackChunkName: "export" */ '@/views/form/Export.vue'
+              ),
+            meta: {
+              breadcrumbTitle: 'Submissions Export',
+              requiresAuth: IdentityProviders.IDIR,
```
jasonchung1871, line 139: IDIR only for submissions export? Shouldn't it be just requiresAuth: true?

app/frontend/src/router/index.js
```diff
@@ -109,7 +109,7 @@ export default function getRouter(basePath = '/') {
               ),
             meta: {
               breadcrumbTitle: 'Form Designer',
-              requiresAuth: IdentityProviders.IDIR,
+              requiresAuth: true,
```
jasonchung1871, line 112: This one should still be IdentityProviders.IDIR as we only want IDIR users to use the form designer

jasonchung1871 (changes requested):

app/frontend/src/router/index.js
timisenco2015: Thanks for this. I mixed Form Designer up with the Submission Export

app/frontend/src/router/index.js
timisenco2015: Thanks for this. I mixed Form Designer with the Submission Export.

jasonchung1871: The FormDesigner here should still be IDIR

app/frontend/src/router/index.js
jasonchung1871: The SubmissionsExport here should be requiresAuth: true

jasonchung1871 (changes requested):

app/frontend/src/router/index.js
timisenco2015: This has been fixed

app/frontend/src/router/index.js
timisenco2015: This has been fixed

app/frontend/src/components/forms/ExportSubmissions.vue
```diff
@@ -275,9 +316,25 @@ export default {
...
       startDateMenu: false,
-      versionSelected: 1,
-      csvTemplates: 'flattenedWithBlankOut',
+      versionSelected: 0,
+      versionSelectedId: '',
+      csvFormats: 'flattenedWithBlankOut',
       versions: [],
+      allDataFields: true,
+      inputFilter: '',
+      select: ['Vuetify', 'Programming'],
```
jasonchung1871, line 325: What is this for?

jasonchung1871 (changes requested):

app/frontend/src/components/forms/ExportSubmissions.vue
```diff
@@ -1,269 +1,309 @@
...
-                      readonly
+  <div>
+    <v-row class="mt-6" no-gutters>
+      <v-col>
+        <v-row>
+          <v-col cols="12" sm="6" order="2" order-sm="1">
+            <h1>Export Submissions to File</h1>
+          </v-col>
+          <v-col class="text-right" cols="12" sm="6" order="1" order-sm="2">
+            <span v-if="canViewSubmissions">
```
jasonchung1871, line 10: I think this conditional is unnecessary as this page should only be accessible if you were a submission reviewer.

app/frontend/src/components/forms/ExportSubmissions.vue
```diff
@@ -1,269 +1,309 @@
...
+                            placeholder="yyyy-mm-dd"
+                            append-icon="event"
+                            v-on:click:append="startDateMenu = true"
+                            readonly
+                            v-on="on"
+                            dense
+                            outlined
+                          ></v-text-field>
+                        </template>
+                        <v-date-picker
```
jasonchung1871, line 163: Why did we go with two separate date pickers with custom logic vs using the built in vuetify date picker with a range?

https://v2.vuetifyjs.com/en/components/date-pickers/#range

app/frontend/src/components/forms/ExportSubmissions.vue
```diff
@@ -371,25 +465,59 @@ export default {
         });
       }
     },
-    updateVersions() {
+    canViewSubmissions() {
```
jasonchung1871, line 468: This should be a computed function instead of a method function. Also, this shouldn't be necessary as you need the role of submission viewer to access this page.

app/frontend/src/services/formService.js
```diff
@@ -407,7 +433,7 @@ export default {
     );
   },
 
-  /**
+  /**a
```
jasonchung1871, line 436: Remove this `a`

jasonchung1871 (changes requested):

app/frontend/src/components/forms/ExportSubmissions.vue
timisenco2015: This was inherited from the showcase team. We can create a separate ticket for this.

PR #690, opened 2023-04-26
FORMS-605 - Add message queue to file upload service

bcvesalink:
# Description
As part of the back end solution for the bulk upload/multi upload we need to ensure we have resources in place to avoid overload and scaling challenges. a queue system of sorts is needed to ensure if many actors attempt to multi upload at one time, the system can queue up requests accordingly

## Types of changes

## Checklist

### DIFF
--- a/app/src/forms/file/service.js
+++ b/app/src/forms/file/service.js
@@ -1,39 +1,69 @@
 const config = require('config');
 const { v4: uuidv4 } = require('uuid');
 
+const Queue = require('bull');
+
 const { FileStorage } = require('../common/models');
 const storageService = require('./storage/storageService');
 
 const PERMANENT_STORAGE = config.get('files.permanent');
 
+// create our job queue
+const queue = new Queue('Upload files queue', { redis: { password: 'redispass' } });
+
 const service = {
   create: async (data, currentUser) => {
-    let trx;
-    try {
-      trx = await FileStorage.startTransaction();
+    const job = await queue.add('Upload files job', {
+      fileData: data,
+      user: currentUser,
+    });
 
-      const obj = {};
-      obj.id = uuidv4();
-      obj.storage = 'uploads';
-      obj.originalName = data.originalname;
-      obj.mimeType = data.mimetype;
-      obj.size = data.size;
-      obj.path = data.path;
-      obj.createdBy = currentUser.usernameIdp;
+    queue
+      .on('completed', function () {
+        console.log('Upload file Job completed...');
+      })
+      .on('failed', function () {
+        console.log('Upload file Job failed');
+      })
+      .on('progress', function (job, progress) {
+        console.log('\rUpload file job #' + job.id + ' ' + progress + '% complete');
+      });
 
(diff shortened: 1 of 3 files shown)


app/src/forms/file/service.js
```diff
@@ -1,39 +1,69 @@
...
+    // process upload file jobs, 1 at a time.
+    queue.process('Upload files job', 1, async (job, done) => {
+      const data = job.data.fileData;
+      const currentUser = job.data.user;
 
-      await FileStorage.query(trx).insert(obj);
+      job.progress(0);
+      let trx;
+      try {
+        trx = await FileStorage.startTransaction();
```
jasonchung1871, line 40: Shouldn't the FileStorage insertion only happen when the file is successfully uploaded? What happens when it fails?

app/src/forms/file/service.js
```diff
@@ -1,39 +1,69 @@
...
-      obj.id = uuidv4();
-      obj.storage = 'uploads';
-      obj.originalName = data.originalname;
-      obj.mimeType = data.mimetype;
-      obj.size = data.size;
-      obj.path = data.path;
-      obj.createdBy = currentUser.usernameIdp;
+    queue
+      .on('completed', function () {
+        console.log('Upload file Job completed...');
```
jasonchung1871, line 23: Most of these console.log statements need to be removed

jasonchung1871 (changes requested):

app/src/forms/file/service.js
bcvesalink: agree for success completion but still I think need leave if it fails for logging

app/src/forms/file/service.js
bcvesalink: if fails it would've caught by try/catch and rollback on https://github.com/bcgov/common-hosted-form-service/pull/690/files/8a2557266d9637b39d029a9edb3c4c17275a1b7c#diff-018ee37873b8124ab6eeaf78c0fd49b3ae9f27d08999fb2eb2e9d33a894d060eR62

app/src/forms/file/service.js
jasonchung1871: We should look into using this log.
https://github.com/bcgov/common-hosted-form-service/blob/master/app/src/components/log.js

app/src/forms/file/service.js
jasonchung1871: Isn't the on progress event potentially executed multiple times?
https://github.com/OptimalBits/bull/blob/develop/REFERENCE.md#events
Shouldn't we just try to insert it once the job is completed?

bcvesalink: it executes every time when we do `job.progress(number)` inside the `progress` method, so it basically just to inform/log the queue job status, in our case it's just two breakpoints `0` and `100`, probably even redundant those two, or we can add `50%`  somewhere in a middle

WalterMoar:
Closing this - it's quite old and isn't a priority problem at the moment.

PR #760, opened 2023-05-31
perf: FORMS-695 increase memory for patroni pods

bcgov-citz-ccft:
# Description

Background: the production Patroni database is crashing due to CHEFS using `user_form_access_vw` (possibly only for a subset of users, those with large numbers of forms?). The database is offline for around 250ms, and since it drops all connections the CHEFS pods have to re-establish them. 

To improve performance, increase the default memory `request` and `limit` for the Patroni database pods. This will allow more of the data to be kept in memory, making for faster queries.

## Types of changes

Performance (non-breaking change which fixes an issue)

## Checklist


openshift/patroni.dc.yaml
```diff
@@ -368,12 +368,12 @@ parameters:
...
     displayName: Memory Request
     required: true
-    value: 256Mi
+    value: 384Mi
   - name: MEMORY_LIMIT
     description: Maximum amount of memory the container can use.
     displayName: Memory Limit
     required: true
-    value: 512Mi
+    value: 768Mi
```
bcgov-citz-ccft, line 376: Note that this will be applied to dev, test, and prod. This isn't necessarily a bad thing, _unless_ dev or test don't use the `request` amount. In that case we can do something like a per-environment parameter file for these settings.

WalterMoar:
> Release [874feef](https://github.com/bcgov/common-hosted-form-service/commit/874feefbaf950c1b88e99ed2923bdd756b6df970) deployed at https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-760

Note that although this is applied as a PR, the PR deployments use ephemeral Patroni, so the settings in this PR will not be used.

jasonchung1871 (approved):

PR #805, opened 2023-06-21
Do not Merge - For review only - Add in form subscribe

jimmyPasta:
# Description
Add in a form subscribe table/view and add in seed data
The 999 seed had issues in our environment running due to it expecting: 
formId: ID.forms[0] 
to be 
formId: ID.forms[0].id
This may be an environmental issue but both Aurelio and I ran into it


jimmyPasta:
Early look at creating pull request

PR #833, opened 2023-07-04
 Right to left (RTL) direction support

timisenco2015:
# Description
Some languages, such as Persian, Hindi, and Arabic, support writing from right to left instead of the traditional left to right. This feature will enable CHEFS to support languages that write from right to left.

## Types of changes

## Checklist


app/frontend/src/components/admin/AdminPage.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <v-tabs>
+  <v-tabs :class="{ 'dir-rtl': isRTL }">
```
WalterMoar, line 2: There is a small problem with the tabs: when you change languages, the line under the current tab is not redrawn. For example, if you go from English (FORMS) to a language with a longer word, like Portuguese (FORMULARIOS), the line isn't long enough.

image

This isn't such a big deal, but if you switch to a RtoL language, the line is on the wrong side of the screen.

image

If you slightly resize the browser window to cause a redraw, then it is fine for either case.

image


Can we force a redraw of the v-tab when the language changes?

app/frontend/src/components/designer/FormViewerActions.vue
```diff
@@ -1,11 +1,11 @@
 <template>
-  <v-row class="d-print-none">
+  <v-row class="d-print-none" :class="{ 'dir-rtl': isRTL }">
     <v-col v-if="formId">
       <v-btn outlined @click="goToAllSubmissionOrDraft">
-        <span>{{ $t('trans.formViewerActions.viewAllSubmissions') }}</span>
+        <span>{{ $t('trans.formViewerActions.viewAllSubmissions') }} s </span>
```
WalterMoar, line 5: This extra "s" is causing problems.

image

timisenco2015: I will remove it. I added that when I was trying to confirm if I'm on the right page and I forgot to remove it.

app/frontend/src/store/modules/form.js
```diff
@@ -882,6 +887,15 @@ export default {
     },
     async setMultiLanguage({ commit }, multiLanguage) {
       commit('SET_MULTI_LANGUAGE', multiLanguage);
+      if (
+        multiLanguage === 'ar' ||
+        multiLanguage === 'fa' ||
+        multiLanguage === 'pa'
```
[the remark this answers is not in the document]
timisenco2015: I will add the Shahmukhi script and leave Gurmukhi to be ltr.

app/frontend/src/components/admin/AdminPage.vue
timisenco2015: We should be able to do that. I will work on it

WalterMoar:
I noticed a RTL bug that isn't part of the PR but maybe it's an easy fix? On the Admin page Forms table, the created date is displaying "DATE TIME AM/PM - USER" for LTR but for RTL it's "AM/PM - USER TIME DATE":

image

Not sure how this is supposed to display in RTL! :) I'm thinking just reversing the order of the items probably isn't correct?

app/frontend/src/components/admin/AdminPage.vue
```diff
@@ -30,5 +31,14 @@ export default {
...
+  computed: {
+    ...mapGetters('form', ['isRTL']),
+  },
 };
 </script>
+<style lang="css" scoped>
+.dir-rtl {
+  direction: rtl !important;
+  text-align: right;
+}
```
WalterMoar, line 43: Would it be better to have `.dir-rtl` as a global style instead of scoped here?

app/frontend/src/components/admin/AdminFormsTable.vue
```diff
@@ -1,24 +1,29 @@
...
+          :class="isRTL ? 'float-left' : 'float-right'"
+        >
           <v-text-field
             v-model="search"
             append-icon="mdi-magnify"
             :label="$t('trans.adminFormsTable.search')"
             single-line
             hide-details
             class="pb-5"
+            :class="[{ 'dir-rtl': isRTL }, isRTL ? 'label' : null]"
```
WalterMoar, line 26: I'm just curious, is:

> [{ 'dir-rtl': isRTL }, isRTL ? 'label' : null]

the same as:

> { 'dir-rtl': isRTL, 'label': isRTL }

timisenco2015: This should work. I did not think that way. Thanks

app/frontend/src/components/base/BaseFilter.vue
```diff
@@ -63,6 +64,7 @@
       <v-btn
         @click="cancelFilterData"
         class="mt-3 ml-3 primary--text"
+        :class="isRTL ? 'mr-3' : 'ml-3'"
```
WalterMoar, line 67: Does the `ml-3` need to be removed from the `class` on the top line?

app/frontend/src/components/designer/FormDesigner.vue
```diff
@@ -1,65 +1,68 @@
...
+            <span>{{ $t('trans.formDesigner.importDesign') }}</span>
+          </v-tooltip>
+        </v-col>
+        <!-- form name -->
+      </v-row>
+    </div>
+    <BaseInfoCard class="my-6" :class="{ 'dir-rtl': isRTL }">
       <h4 class="primary--text">
-        <v-icon class="mr-1" color="primary">info</v-icon
+        <v-icon :class="isRTL ? 'ml-1' : mr - 1" color="primary">info</v-icon
```
WalterMoar, line 65: It looks like the `mr-1` wasn't quoted and prettier added spaces around the '-'.

app/frontend/src/store/modules/form.js
```diff
@@ -126,6 +128,9 @@ export default {
     SET_FORM(state, form) {
       state.form = form;
     },
+    SET_IS_RTl(state, isRTL) {
```
WalterMoar, line 131: Can you uppercase the `l` in `SET_IS_RTl`, for consistency?

app/frontend/src/components/admin/AdminFormsTable.vue
WalterMoar: What would you think about replacing all the ternary operators that have "null" as the second argument with the "object" style? I think it's a bit easier to read, and more consistent?

app/frontend/src/internationalization/trans/chefs/en/en.json
```diff
@@ -823,8 +837,8 @@
       "updCurrUserFormPrefConsErrMsg": "Error updating user form prefs using formID {formId}, and prefs {preferences}: {error}",
       "getCurrUserFormPrefErrMsg": "An error occurred while fetching your preferences for this form.",
       "getCurrUserFormPrefConsErrMsg": "Error getting user form prefs using formID {formId}: {error}",
-      "delCurrformNotiMsg": "Form \"{state.form.name}\" has been deleted successfully.",
-      "delCurrformErrMsg": "An error occurred while attempting to delete \"${state.form.name}\".",
+      "delCurrformNotiMsg": "Form {name} has been deleted successfully.",
+      "delCurrformErrMsg": "An error occurred while attempting to delete {state.form.name}.",
```
WalterMoar, line 841: Hmm, it looks like `delCurrformErrMsg` is never used.

timisenco2015: It probably needed to be removed and the translation updated. I will remove it

app/frontend/src/components/admin/AdminFormsTable.vue
```diff
@@ -1,24 +1,29 @@
...
         <v-checkbox
           class="pl-3"
+          :class="isRTL ? 'float-right' : 'float-left'"
           v-model="activeOnly"
           :label="$t('trans.adminFormsTable.showDeletedForms')"
           @click="refeshForms"
         />
       </v-col>
-      <v-col cols="12" sm="4">
+      <v-col cols="6">
```
WalterMoar, line 13: I'm curious about the change here of the `cols` and `sm` values. Is this to change the layout on mobile devices so that the deleted forms checkbox and the search field are always on the same line?

app/frontend/src/components/admin/AdminPage.vue
WalterMoar: Sorry - to clarify: could the `.dir-rtl` style be defined in `styles.scss`, so that it doesn't have to be defined on every page that uses it?

app/frontend/src/components/admin/AdminPage.vue
```diff
@@ -27,8 +28,18 @@ export default {
   },
   data() {
     return {
+      index: 2,
```
WalterMoar, line 31: Is `index` needed here?

app/frontend/src/components/admin/AdminPage.vue
```diff
@@ -27,8 +28,18 @@ export default {
...
       adminDashboardUrl: this.$config.adminDashboardUrl,
     };
   },
+  computed: {
+    ...mapGetters('form', ['isRTL', 'multiLanguage']),
```
WalterMoar, line 36: What is `multiLanguage` used for?

app/frontend/src/components/admin/AdminPage.vue
timisenco2015: I forgot to remove it. I was trying to work on the tab issue

app/frontend/src/components/admin/AdminPage.vue
timisenco2015: I forgot to remove it. I was trying to work on the tab issue

app/frontend/src/components/admin/AdminPage.vue
WalterMoar: Oh, OK - maybe I should wait until you're done!

app/frontend/src/components/admin/AdminFormsTable.vue
WalterMoar: This is the old:
image
and the new:
image

app/frontend/src/components/designer/FormDesigner.vue
```diff
@@ -1,65 +1,68 @@
...
+            <span>{{ $t('trans.formDesigner.importDesign') }}</span>
+          </v-tooltip>
+        </v-col>
+        <!-- form name -->
+      </v-row>
+    </div>
+    <BaseInfoCard class="my-6" :class="{ 'dir-rtl': isRTL }">
       <h4 class="primary--text">
-        <v-icon class="mr-1" color="primary">info</v-icon
+        <v-icon :class="isRTL ? 'ml-1' : 'mr - 1'" color="primary">info</v-icon
```
WalterMoar, line 65: I think this should be `mr-1` without the spaces. It's really hard to see the difference, though, since it's only a few pixels!

app/frontend/src/components/forms/manage/ManageVersions.vue
```diff
@@ -57,18 +57,24 @@
       <!-- Status  -->
       <template #[`item.status`]="{ item }">
         <v-switch
+          :class="{ 'dir-rtl': isRTL }"
```
WalterMoar, line 60: Hmm, something broke the switch when viewing in RTL. It looks OK in dev, but in this PR the round circle isn't in the right place:

image

timisenco2015: This is not from us. It is from Vuetify. I will raise the issue with them but I have tried to see what can be done on our end but nothing is working

app/frontend/src/components/admin/AdminPage.vue
timisenco2015: I tried it but did not work well

app/frontend/src/components/admin/AdminPage.vue
```diff
@@ -27,8 +28,17 @@ export default {
...
+      tab: null,
       adminDashboardUrl: this.$config.adminDashboardUrl,
     };
   },
+  computed: {
+    ...mapGetters('form', ['isRTL']),
+  },
+  watch: {
+    isRTL() {
+      this.tab = null;
```
WalterMoar, line 40: This does have the effect of changing which tab is selected when switching between LTR and RTL languages. Since it's the admin page and it's a pretty weird use case, shall we leave it for now and fix later?

app/frontend/src/internationalization/trans/chefs/pu/pu.json
```diff
@@ -0,0 +1,898 @@
...
+    "401UnAuthorized": "401: انئدھکارت۔ :",
+    "401UnAuthorizedErrMsg": "تہانوں اس پنے تکّ پہنچ کرن دی اجازت نہیں ہے۔"
+  },
+  "formDesigner": {
+    "formDesign": "فارم ڈزائین",
+    "exportDesign": "نریات ڈزائین",
+    "importDesign": "ڈزائین آیات کرو",
+    "important": "مہتوپورن",
+    "formDesignInfoA": "جدوں تسیں اس فارم نوں بناؤنا پورا کر لیندے ہو تاں <�strong>سیو ڈزائین<�/strong> بٹن دی ورتوں کرو۔",
+    "formDesignInfoB": "تہاڈے اپبھوگتا نوں اہ فارم جمھاں کراؤن لئی <�strong>SUBMIT بٹن دتا گیا ہے اتے اسنوں سرکھئت کرن توں بائد کریاشیل ہو جاویگا۔",
```
WalterMoar, line 297: It looks like there is some kind of special character here, and it's not formatting as STRONG:
image

app/frontend/src/components/designer/FormSettings.vue
```diff
@@ -160,7 +198,7 @@
             :disabled="userType === ID_MODE.PUBLIC"
           >
             <template #label>
-              <span>
+              <span :class="{ 'mr-2': isRTL }">
                 Allow <strong> multiple draft</strong> upload
```
WalterMoar, line 202: Looks like we're missing all the translations for "Allow **multiple draft** upload". Should we fix in this PR or do it in another one?

timisenco2015: I will fix it

app/frontend/src/components/designer/FormSettings.vue
```diff
@@ -24,6 +25,7 @@
             data-test="text-description"
             v-model="description"
             :rules="descriptionRules"
+            :class="[{ 'dir-rtl': isRTL, label: isRTL }]"
```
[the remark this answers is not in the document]
timisenco2015: What page is this? We cannot control the legend.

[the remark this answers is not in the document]
WalterMoar: It's the "Manage Form" > "Form Settings":
image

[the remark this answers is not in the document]
timisenco2015: We cannot control the legend.

app/frontend/src/components/designer/FloatButton.vue
```diff
@@ -1,5 +1,6 @@
 <template>
   <div
+    :class="{ 'dir-rtl': true }"
```
WalterMoar, line 3: Is this supposed to always be true, or is it based on isRTL? (if always true, then no need to add isRTL to the computed)

app/frontend/src/components/designer/FormDesigner.vue
```diff
@@ -1,65 +1,68 @@
...
-          >{{ $t('trans.formDesigner.version') }} :
-          {{ this.displayVersion }}</em
-        >
-      </v-col>
-    </v-row>
-    <BaseInfoCard class="my-6">
+    <div :class="{ 'dir-rtl': isRTL }">
+      <v-row class="mt-6" no-gutters>
+        <!-- page title -->
+        <v-col cols="11">
```
WalterMoar, line 6: Is there a reason to switch to using 11 columns for the title? Or should it be 12, like before?

app/frontend/src/components/designer/FormDesigner.vue
```diff
@@ -1,65 +1,68 @@
...
+          <v-row>
+            <v-col cols="12">
```
WalterMoar, line 8: Can this inner v-row and v-column be removed? Or is it needed for something?

WalterMoar:
I found another bug unrelated to the RTL work, but maybe this is a good place to fix it? If I go to the create a new form page and select then unselect the disclaimer checkbox, I get the wrong error message. It's displaying `confirmPageNav` instead of `agreementErrMsg` in Create.vue:100

image

app/frontend/src/components/designer/FormDisclaimer.vue
```diff
@@ -1,17 +1,23 @@
...
+        </span>
       </li>
       <li>
-        {{ $t('trans.formDisclaimer.consent') }}
+        <span :class="{ 'dir-rtl': isRTL }">
+          {{ $t('trans.formDisclaimer.consent') }}
+        </span>
       </li>
       <li>
         {{ $t('trans.formDisclaimer.formIntention') }}
```
WalterMoar, line 23: I'm confused here - the formIntention doesn't have the RTL but it seems to be working? Maybe you don't need it on every <li> because the <ul> does it for all of the <li> items?

WalterMoar: Oh, nevermind.... the `ul` doesn't set `dir-rtl`. So why is this working for `formIntention`?

WalterMoar: Is it working because the Create.vue page that is using the FormDisclaimer already has RTL set?

My concern is that we have these `dir-rtl` classes in so many places, but maybe they aren't needed? Maybe it's just on the outer containers that we need it set?

app/frontend/src/internationalization/trans/chefs/ar/ar.json
```diff
@@ -770,8 +775,15 @@
     "copyToClipboard": "نسخ إلى الحافظة",
     "errCopyToClipboard": "خطأ في محاولة النسخ إلى الحافظة."
   },
-  "root": {
-    "admin": "مسؤل"
+  "admin": {
+    "root": "مسؤل"
+  },
```
WalterMoar, line 780: I think this is supposed to be admin.root.admin

app/frontend/src/internationalization/trans/chefs/en/en.json
```diff
@@ -740,6 +752,7 @@
   "addTeamMember": {
     "cantFindChefsUsers": "Can't find someone? They may not have logged into CHEFS.<br /> Kindly send them a link to CHEFS and ask them to log in.",
     "cancel": "Cancel",
+    "add": "Add",
```
WalterMoar, line 755: This `addTeamMember.add` is missing from the ar.json and de.json files.

app/frontend/src/components/designer/FormDisclaimer.vue
timisenco2015: We are using Vuetify components, and I have to add dir-rtl to each Vuetify component so it works correctly.

WalterMoar: But for this page being just HTML - if you put `:class="{ 'dir-rtl': isRTL }"` _only_ on the `div` on line 2, will it still work correctly for all the tags under it? (just curious!)

WalterMoar: Actually, no - if you remove _all_ the `:class="{ 'dir-rtl': isRTL }"` from this file, will it still work because it's included in another page? (again just curious - I'm hoping we can make things as simple as possible, so that there is less code to maintain in the future).

timisenco2015: I have fixed that. I am also scanning through other components too.

timisenco2015: There are some pages/components that I only apply dir-RTL to the outer container while some I avoided the outer container because it will affect the layout for example in FormDesigner.vue

app/frontend/src/components/designer/FormViewer.vue
```diff
@@ -62,40 +67,47 @@
           "
           transition="scale-transition"
         >
-          <div v-if="saving">
+          <div v-if="saving" :class="[isRTL ? 'mr-1' : '']">
             <v-progress-linear indeterminate />
-            {{ $t('trans.formViewer.saving') }}
+            {{ $t('trans.formVewer.saving') }}
```
WalterMoar, line 72: There's a type here with `formVewer`.

app/frontend/src/components/designer/FormViewer.vue
```diff
@@ -62,40 +67,47 @@
           "
           transition="scale-transition"
         >
-          <div v-if="saving">
+          <div v-if="saving" :class="[isRTL ? 'mr-1' : '']">
```
WalterMoar, line 70: For consistency do we also want to do these as `:class="{ 'mr-1': isRTL }"`? I ask because when I see this I wonder if `:class` treats the empty string as no class to add, or what it does...

app/frontend/src/components/designer/FormViewer.vue
```diff
@@ -133,26 +145,43 @@
           @render="onFormRender"
           :language="multiLanguage"
         />
-        <p v-if="version" class="text-right">
+        <p
+          v-if="version"
+          :class="isRTL ? 'text-left' : 'text-right'"
+          class="mt-9"
```
WalterMoar, line 151: Just curious - did you mean to add this `mt-9` margin?

app/frontend/src/components/forms/manage/AddTeamMember.vue
```diff
@@ -110,22 +111,22 @@
             <!-- buttons -->
             <v-btn
               color="primary"
-              class="ml-2"
+              class="isRTL ? mr-3 : ml-2"
```
WalterMoar, line 114: Do we want `mr-2` to match the `ml-2`, or does it need the extra space?

app/frontend/src/components/designer/FormViewer.vue
timisenco2015: Correct

app/frontend/src/components/forms/submission/ManageSubmissionUsers.vue
```diff
@@ -15,26 +15,27 @@
...
     </v-tooltip>
     <v-dialog v-model="dialog" width="600">
-      <v-card>
+      <v-card :class="{ 'dir-rtl': isRTL }">
         <v-card-title class="text-h5 pb-0">
           {{ $t('trans.manageSubmissionUsers.manageTeamMembers') }}
-          <v-radio-group v-model="selectedIdp" row>
+        </v-card-title>
+        <v-card-subtitle class="mt-1">
+          <v-radio-group v-model="yselectedIdp" row>
```
WalterMoar, line 23: I think there's an extra `y` here before `selectedIdp`.

app/frontend/src/views/form/Preview.vue
```diff
@@ -1,16 +1,15 @@
 <template>
   <BaseSecure :idp="[IDP.IDIR]">
     <h1>
-      PREVIEW
+      {{ $t('trans.preview.preview') }}
```
WalterMoar, line 4: Does this need to be on the right side of the screen? Does the `?` need to be on the other side of the text?

app/frontend/src/views/form/Success.vue
```diff
@@ -3,12 +3,12 @@
...
+          <h1 class="mb-5" :class="{ 'dir-rtl': isRTL }">
             <v-icon large color="success">check_circle</v-icon>
             {{ $t('trans.sucess.sucessFormSubmissn') }}
           </h1>
           <div v-if="form.showSubmissionConfirmation">
-            <h3>
+            <h3 :class="{ 'dir-rtl': isRTL }">
               <span class="d-print-none">
                 {{ $t('trans.sucess.keepRecord') }}
               </span>
```
WalterMoar, line 14: Minor bug - on line 16 the `trans.sucess.confirmationId` translations all have a `:` at the end, so we end up with two `::`

image

app/frontend/src/views/user/Root.vue
```diff
@@ -2,15 +2,15 @@
   <BaseSecure>
```
WalterMoar, line 2: Is this view still used? I can't figure out how to navigate to it, and if I go to it directly I get a 401.

timisenco2015: Not sure

app/frontend/src/views/About.vue
```diff
@@ -36,8 +44,10 @@
 
     <v-row justify="center" class="example-text">
       <v-col cols="12" lg="4">
-        <h2>{{ $t('trans.homePage.chefsHowToTitle') }}</h2>
-        <p>
+        <h2 :class="{ 'dir-rtl': isRTL }">
+          {{ $t('trans.homePage.chefsHowToTitle') }}
+        </h2>
+        <p :class="{ 'dir-rtl': isRTL }">
           {{ $t('trans.homePage.chefsHowToSub') }}
```
WalterMoar, line 51: It looks like the `chefsHowToSub` translations all contain "Get Started" at the end, but there is another `getStarted` text piece below that repeats it.

WalterMoar: (not related to this PR!)

app/frontend/src/components/designer/FormDisclaimer.vue
```diff
@@ -1,17 +1,23 @@
...
     <h3 class="mb-5">
       {{ $t('trans.formDisclaimer.disclaimerAndStatement') }}
     </h3>
-    <ul>
+    <ul :class="isRTL ? 'mr-6' : null">
       <li>
-        {{ $t('trans.formDisclaimer.privacyLaw') }}
+        <span>
+          {{ $t('trans.formDisclaimer.privacyLaw') }}
+        </span>
```
WalterMoar, line 10: Do we still need these `span` tags? It would be good to make as few changes as possible. Ideally it would be great if everything change in the "file diff" view of the PR was for a reason.

app/frontend/src/components/admin/AdminPage.vue
timisenco2015: I could not find any alternative solution.

app/frontend/src/components/admin/AdminPage.vue
jasonchung1871: image

It's still doing this one, but the line is moving to the right spot when switching to RtoL.

timisenco2015: Vuetify uses the left position, and there is a way they calculate that which is not exposed. I can create new ticket. What do you think?

app/frontend/src/store/modules/form.js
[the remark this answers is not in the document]
jasonchung1871: This got changed to `pu` but what is `pu`? I don't see it in the ISO 639-1 or the ISO 639-2 codes.

app/frontend/src/components/forms/manage/ManageVersions.vue
jasonchung1871: I've unresolved this conversation as this can be resolved by removing the RTL on the switch itself. 
image
image

app/frontend/src/components/admin/AdminPage.vue
jasonchung1871: Why is it necessary to set the tab to null when changing to a RtoL language?

app/frontend/src/components/admin/AdminPage.vue
jasonchung1871: I think it's fine, this is only visible in the admin page

jasonchung1871 (approved):

PR #838, opened 2023-07-05
Forms-735_auto_gen_api_doc_with_swagger

timisenco2015:
# Description

## Types of changes

Documentation (non-breaking change with enhancements to documentation) -->

## Checklist


app/src/forms/admin/index.js
```diff
@@ -2,6 +2,14 @@ const dataErrors = require('../common/middleware').dataErrors;
 const routes = require('./routes');
 const setupMount = require('../common/utils').setupMount;
 
+const swaggerSpec = require('../../swagger/swaggerui');
+
+swaggerSpec.tags.push({
+  name: 'Admin',
+  description:
+    'These API endpoints handle the input data provided by a user that completes the form, for example, writing the form data to the database, or exporting form submissions created during a defined time period.',
+});
```
WalterMoar, line 11: This is the description of the `submissions` endpoint in the previous docs. Do we want it to be what the `admin` endpoint used to be?

> These API endpoints are used for elevated permission administrative calls. All calls are secured by a role that operational team members will be granted and allow fetching some details about forms without needing a user->form permission.

app/src/forms/admin/routes.js
```diff
@@ -18,22 +18,160 @@ routes.use(currentUser);
...
 // Forms
 //
+
+/**
+ * @openapi
+ * /admin/forms:
+ *  get:
+ *    tags:
+ *      - Admin
+ *    description: This endpoint will fetch all forms under the admin access.
```
WalterMoar, line 28: It looks like if there's no "title" defined, then the description is truncated to 50 characters and used as the title. Since this doesn't really offer any new information, would it be better to set the "title" for each endpoint to what it was in the old docs?

**Old:**
image

**New:**
image

WalterMoar: (In other words, would it be better to do:
```
*   title: List all forms in the system
*   description: This endpoint will fetch all forms under the admin access.
```

app/src/forms/admin/routes.js
```diff
@@ -18,22 +18,160 @@ routes.use(currentUser);
...
+ *      '200':
+ *        description: Success
+ *        content:
+ *          application/json:
+ *            schema:
+ *              $ref: '#/components/responses/responseBody/AdminReadFormEx'
+ *      '403':
+ *        $ref: '#/components/responses/Error/AccessDenied'
+ *      '5XX':
+ *        $ref: '#/components/responses/Error/UnExpected'
```
WalterMoar, line 75: If an endpoint takes parameters, it should also respond with a 404 if the parameter(s) are wrong. Can the 404 response be added to all these endpoints?

app/src/forms/admin/routes.js
WalterMoar: (what's the difference between "title" and "summary" in the OpenAPI spec? Maybe it should be summary instead of title?)

app/src/swagger/swagger_api_doc.yaml
```diff
@@ -0,0 +1,4449 @@
...
+              schema:
+                $ref: '#/components/schemas/respError/UnExpectedError'
+        UserNotFound:
+          description: User Not Found
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/respError/UserNotFoundError'
+        AccessDenied:
+          description: Acess Denied
```
WalterMoar, line 72: Typo here: "Acess" instead of "Access"

app/src/forms/file/routes.js
```diff
@@ -9,14 +9,130 @@ const { currentUser } = require('../auth/middleware/userAccess');
...
+ *        name: fileId
+ *        schema:
+ *          type: string
+ *          format: uuid
+ *        description: ID of the file.
+ *        required: true
+ *        example: 3cb9acc7-cfd8-4491-b091-1277bc0ec303
+ *    responses:
+ *      '200':
+ *        description: Sucess
```
WalterMoar, line 132: Typo here: "Sucess" instead of "Success".

app/src/forms/rbac/routes.js
```diff
@@ -9,42 +9,578 @@ const { currentUser, hasFormPermissions, hasSubmissionPermissions, hasFormRoles,
...
+ *  get:
+ *    tags:
+ *      - RBAC
+ *    description: This endpoint will get list of a list of identity providers.
+ *    security:
+ *      - bearerAuth: []
+ *      - openId: []
+ *    responses:
+ *      '200':
+ *        description: Sucess
```
WalterMoar, line 77: Typo here: "Sucess" instead of "Success".

app/src/forms/rbac/routes.js
```diff
@@ -9,42 +9,578 @@ const { currentUser, hasFormPermissions, hasSubmissionPermissions, hasFormRoles,
...
+ *    requestBody:
+ *      required: true
+ *      content:
+ *        application/json:
+ *          schema:
+ *            type: array
+ *            example: ["7cc11a0f-a7c9-4cbe-bd92-a3a22781ac8c", "3ddb68cb-b936-49fe-9ba4-48b04859eaf7" ]
+ *    responses:
+ *      '200':
+ *        description: Sucess
```
WalterMoar, line 568: Typo here: "Sucess" instead of "Success".

app/src/swagger/swagger_api_doc.yaml
```diff
@@ -0,0 +1,4449 @@
...
+              formName:
+                type: string
+                example: BC Gov Form
+                description: Name of the form.
+              label:
+                type: string
+                example: ''
+              identityProviders:
+                type: string
+                description: List of identity providers that can be used to acess this form.
```
WalterMoar, line 221: Typo here: "acess" instead of "access".

app/src/swagger/swagger_api_doc.yaml
```diff
@@ -0,0 +1,4449 @@
...
+              example: false
+              default: false
+            allowSubmitterToUploadFile:
+              type: boolean
+              description: If enabled, Submitters can upload when submitting submissions
+              example: false
+              default: false
+            identityProviders:
+              type: string
+              description: List of identity providers that can be used to acess the form
```
WalterMoar, line 864: Typo here: "acess" instead of "access".

app/src/swagger/swagger_api_doc.yaml
```diff
@@ -0,0 +1,4449 @@
...
+                  encrypted: false
+                  properties: {}
+                  mask: false
+                  inputType: text
+                  inputFormat: plain
+                  displayMask: ''
+                  truncateMultipleSpaces: false
+          identityProviders:
+            type: string
+            description: List of identity providers that can be used to acess the form
```
WalterMoar, line 2988: Typo here: "acess" instead of "access".

app/src/swagger/swagger_api_doc.yaml
```diff
@@ -0,0 +1,4449 @@
...
+            type: string
+            description: Name of the form.
+            example: test
+          description:
+            type: string
+            description: Short details about the form.
+            example: ''
+          identityProviders:
+            type: string
+            description: List of identity providers that can be used to acess this form
```
WalterMoar, line 3039: Typo here: "acess" instead of "access".

app/src/forms/admin/routes.js
```diff
@@ -18,22 +18,160 @@ routes.use(currentUser);
...
 
+/**
+ * @openapi
+ * /admin/forms/{formId}/apiKey:
+ *  delete:
+ *    tags:
+ *      - Admin
+ *    description: This endpoint will delete the form API key.
+ *    security:
+ *      - openId: []
```
WalterMoar, line 89: In the old documentation this also had `BearerAuth` - were the old docs wrong?

app/src/forms/admin/routes.js
```diff
@@ -18,22 +18,160 @@ routes.use(currentUser);
...
 
+/**
+ * @openapi
+ * /admin/forms/{formId}/apiKey:
+ *  get:
+ *    tags:
+ *      - Admin
+ *    description: This endpoint will fetch the API key details for this form.
+ *    security:
+ *      - openId: []
```
WalterMoar, line 119: In the old documentation this also had `BearerAuth` - were the old docs wrong?

app/src/forms/admin/routes.js
```diff
@@ -65,18 +367,141 @@ routes.get('/users/:userId', async (req, res, next) => {
 //Form componets help info
 //
 
+/**
+ * @openapi
+ * /admin/formcomponents/proactivehelp/object:
```
WalterMoar, line 372: Oh, it's great that you found endpoints that weren't documented before. This will be very helpful in the future!

app/src/forms/admin/index.js
timisenco2015: Fix

app/src/forms/admin/routes.js
timisenco2015: Fix

app/src/forms/admin/routes.js
timisenco2015: Fix

app/src/forms/docu/param/formId.json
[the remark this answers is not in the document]
timisenco2015: Fix

app/src/swagger/swagger_api_doc.yaml
timisenco2015: Fix

app/src/forms/file/routes.js
timisenco2015: Fixed

app/src/forms/rbac/routes.js
```diff
@@ -9,42 +9,578 @@ const { currentUser, hasFormPermissions, hasSubmissionPermissions, hasFormRoles,
...
+ *  get:
+ *    tags:
+ *      - RBAC
+ *    description: This endpoint will get list of users and roles for a form
+ *    security:
+ *      - bearerAuth: []
+ *      - openId: []
+ *    responses:
+ *      '200':
+ *        description: Sucess
```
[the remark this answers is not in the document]
timisenco2015: Fixed

app/src/forms/rbac/routes.js
timisenco2015: Fixed

app/src/forms/admin/routes.js
timisenco2015: Fixed

app/src/forms/admin/routes.js
timisenco2015: Fixed

app/src/swagger/swagger_api_doc.yaml
timisenco2015: Fixed

app/src/swagger/swagger_api_doc.yaml
timisenco2015: Fixed

app/src/swagger/swagger_api_doc.yaml
timisenco2015: Fixed

app/src/swagger/swagger_api_doc.yaml
timisenco2015: Fixed

app/src/forms/rbac/routes.js
timisenco2015: Fixed

app/src/forms/rbac/routes.js
timisenco2015: Fixed

app/src/forms/admin/routes.js
```diff
@@ -18,22 +18,175 @@ routes.use(currentUser);
 //
 // Forms
 //
+
+/**
+ * @openapi
+ * /admin/forms:
+ *  get:
```
WalterMoar, line 25: The old docs covered the `active` parameter, and it needs to be added here.

app/src/forms/admin/routes.js
```diff
@@ -18,22 +18,175 @@ routes.use(currentUser);
...
+ *        content:
+ *          application/json:
+ *            schema:
+ *              $ref: '#/components/responses/responseBody/AdminReadFormEx'
+ *      '403':
+ *        $ref: '#/components/responses/Error/AccessDenied'
+ *      '5XX':
+ *        $ref: '#/components/responses/Error/UnExpected'
+ *      '404':
+ *        $ref: '#/components/responses/Error/ResourceNotFound'
```
WalterMoar, line 79: This ref for '#/components/responses/Error/ResourceNotFound' has a ref for '#/components/schemas/respError/ResourceNotFoundError', which has a ref for '#/components/schemas/Problem'. I think this is true for all of the Error responses, they all display the same Problem information. I don't think this is very useful for the developers to have the same general information for every HTTP status code.

Although this is the same as the old docs, can we fix it now so that we display the custom values? For this case it would be what's in '#/components/schemas/respError/ResourceNotFoundError'?

app/src/forms/admin/routes.js
```diff
@@ -18,22 +18,175 @@ routes.use(currentUser);
...
+ *      '200':
+ *        description: Success
+ *        content:
+ *          application/json:
+ *            schema:
+ *              $ref: '#/components/responses/responseBody/AdminsFormGetEx'
+ *      '403':
+ *        $ref: '#/components/responses/Error/AccessDenied'
+ *      '5XX':
+ *        $ref: '#/components/responses/Error/UnExpected'
```
WalterMoar, line 42: Do we want to call this 500 instead of 5XX? There are many 5XX codes that are not "Unexpected". It's really only 500 that's unexpected because it's a server error. New developers might also be confused that there's actually a 5XX error?

We'll also have things like 503 when the OpenShift cluster is really busy, but we don't need to document those... although they should be "expected".

app/src/swagger/swagger_api_doc.yaml
```diff
@@ -0,0 +1,4449 @@
...
+                detail:
+                  example: Unexpected Error.
+                status:
+                  example: 5XX
+                error:
+                  example: {}
+                title:
+                  example: Unexpected
+                type:
+                  example: https://httpstatuses.com/401
```
WalterMoar, line 4391: This should be 5XX (or 500 if we change to 500) rather than 401.

app/src/forms/admin/routes.js
WalterMoar: Or do we want to leave it as 5XX so devs handle all 5XX errors? @jasonchung1871?

app/src/forms/admin/routes.js
```diff
@@ -18,22 +18,175 @@ routes.use(currentUser);
...
+ *        name: formId
+ *        schema:
+ *          type: string
+ *          format: uuid
+ *        description: ID of the form.
+ *        required: true
+ *        example: c6455376-382c-439d-a811-0381a012d696
+ *    responses:
+ *      '200':
+ *        description: Success
```
WalterMoar, line 107: The old docs say that it returns a 204. Were the old docs incorrect? We definitely want to match what's being returned.

app/src/forms/admin/routes.js
```diff
@@ -18,22 +18,175 @@ routes.use(currentUser);
...
+ * @openapi
+ * /admin/forms/{formId}/apiKey:
+ *  delete:
+ *    tags:
+ *      - Admin
+ *    summary: Delete Form API Key
+ *    description: This endpoint will delete the form API key.
+ *    security:
+ *      - openId: []
+ *      - bearerAuth: []
```
WalterMoar, line 95: I can't find the "security" property in the OpenAPI spec, so I don't know how this works - but in the old docs it says "
(BearerAuth and OpenID)" and in the new ones it says "openId or bearerAuth". I assume the new docs are more correct, yes? That we accept either one or the other?

app/src/forms/admin/routes.js
timisenco2015: Okay

app/src/forms/admin/routes.js
timisenco2015: The 5XX is for swagger is used to denote unknown or unexpected error.

app/src/forms/admin/routes.js
```diff
@@ -42,21 +195,193 @@ routes.get('/forms/:formId/versions/:formVersionId', async (req, res, next) => {
...
+ *        $ref: '#/components/responses/Error/ResourceNotFound'
+ *    '422':
+ *      description: 'Unprocessable Entity'
+ *      content:
+ *        application/json:
+ *          schema:
+ *            oneOf:
+ *              - $ref: '#/components/schemas/respError/UnprocessableEntityDBError'
+ *              - $ref: '#/components/schemas/respError/RoleRequiredValidationError'
+ *              - $ref: '#/components/schemas/respError/UserAndFormIdequiredError'
```
WalterMoar, line 315: In the UI it looks like this:

image

but the three "Object" buttons all display the same "Problem" default values. And in the right hand side Response Samples, only the last one (UserAndFormIdequiredError) is shown.

Is there any value in sorting this out? The developer might want to know that there are other errors possible, and not just UserAndFormIdequiredError.

app/src/forms/admin/routes.js
```diff
@@ -42,21 +195,193 @@ routes.get('/forms/:formId/versions/:formVersionId', async (req, res, next) => {
...
+ *        $ref: '#/components/responses/Error/ResourceNotFound'
+ * */
 routes.get('/forms/:formId/formUsers', async (req, res, next) => {
   await controller.getFormUserRoles(req, res, next);
 });
 
+/**
+ * @openapi
+ * /admin/forms/{formId}/addUser:
+ *  put:
```
WalterMoar, line 238: I'm a bit confused here: is the formId actually "required" both as a path parameter and as a query parameter? And should "role" be listed as a non-required query parameter?

app/src/forms/admin/routes.js
```diff
@@ -42,21 +195,193 @@ routes.get('/forms/:formId/versions/:formVersionId', async (req, res, next) => {
...
+ *      - openId: []
+ *    parameters:
+ *      - in: path
+ *        name: formId
+ *        schema:
+ *          type: string
+ *          format: uuid
+ *        description: ID of the form.
+ *        required: true
+ *        example: c6455376-382c-439d-a811-0381a012d696
```
WalterMoar, line 369: This should be the example for userId, yes? With the userId `description` and and `example` values?

app/src/forms/admin/routes.js
timisenco2015: Screenshot 2023-07-05 161912
This is how it is displaying on my end. I will switch to swagger ui to see why it is not showing other errors

app/src/forms/admin/routes.js
timisenco2015: I was too but that was from the showcase team. It is part of what we need clean up or fix

app/src/forms/admin/routes.js
```diff
@@ -65,18 +390,149 @@ routes.get('/users/:userId', async (req, res, next) => {
 //Form componets help info
 //
 
+/**
+ * @openapi
+ * /admin/formcomponents/proactivehelp/object:
+ *  post:
+ *   tags:
+ *    - Admin
+ *   summary: create or update proactive help details form.io componen
```
WalterMoar, line 399: Little typo here for "componen"

app/src/forms/admin/routes.js
```diff
@@ -65,18 +390,149 @@ routes.get('/users/:userId', async (req, res, next) => {
...
+ *      content:
+ *        application/json:
+ *          schema:
+ *            $ref: '#/components/responses/responseBody/AdminGetProactiveHelpObject'
+ *    '403':
+ *      $ref: '#/components/responses/Error/AccessDenied'
+ *    '5XX':
+ *      $ref: '#/components/responses/Error/UnExpected'
+ *    '422':
+ *      $ref: '#/components/responses/Error/UnprocessableEntityDB'
```
WalterMoar, line 421: For some reason this is showing up in the docs as "422 Unauthorized". It should be "403 Access Denied"?

app/src/forms/admin/routes.js
```diff
@@ -65,18 +390,149 @@ routes.get('/users/:userId', async (req, res, next) => {
...
+ *      content:
+ *        application/json:
+ *          schema:
+ *            $ref: '#/components/responses/responseBody/AdminProactiveHelpPublishStatus'
+ *    '403':
+ *      $ref: '#/components/responses/Error/AccessDenied'
+ *    '5XX':
+ *      $ref: '#/components/responses/Error/UnExpected'
+ *    '422':
+ *      $ref: '#/components/responses/Error/UnprocessableEntityDB'
```
WalterMoar, line 464: For some reason this is showing up in the docs as "422 Unauthorized". It should be "403 Access Denied"?

app/src/forms/admin/routes.js
```diff
@@ -65,18 +390,149 @@ routes.get('/users/:userId', async (req, res, next) => {
...
+ *            properties:
+ *              url:
+ *                type: string
+ *                format: binary
+ *    '403':
+ *      $ref: '#/components/responses/Error/AccessDenied'
+ *    '5XX':
+ *      $ref: '#/components/responses/Error/UnExpected'
+ *    '422':
+ *      $ref: '#/components/responses/Error/UnprocessableEntityDB'
```
WalterMoar, line 506: Is this an actual response?

app/src/forms/admin/routes.js
timisenco2015: 403 access denied is if the token has expired, but showcase used 422.

app/src/forms/admin/routes.js
timisenco2015: Yes

WalterMoar: Yikes! We definitely have some cleanup todo - thanks!

app/src/forms/admin/routes.js
WalterMoar: We definitely have some cleanup to do!

app/src/forms/admin/routes.js
WalterMoar: Thank you!

PR #845, opened 2023-07-06
Unable to select all versions for export 

timisenco2015:
# Description
Added “All” option to drop-down menu in submission export page

## Types of changes

## Checklist

### DIFF
--- a/app/frontend/src/components/forms/ExportSubmissions.vue
+++ b/app/frontend/src/components/forms/ExportSubmissions.vue
@@ -425,7 +425,7 @@ export default {
           type: 'submissions',
           draft: false,
           deleted: false,
-          version: version,
+          version: version === 'All' ? undefined : version,
         });
         (this.allDataFields = true), this.selected.push(...this.FILTER_HEADERS);
       }
@@ -477,7 +477,9 @@ export default {
           this.form.id,
           this.exportFormat,
           this.csvFormats,
-          this.exportFormat === 'csv' ? this.versionSelected : undefined,
+          this.exportFormat === 'csv' && this.versionSelected !== 'All'
+            ? this.versionSelected
+            : undefined,
           {
             minDate: from,
             maxDate: to,
@@ -533,7 +535,8 @@ export default {
           versions.sort((a, b) => b.published - a.published);
         }
         this.versions.push(
-          ...this.form.versions.map((version) => version.version)
+          ...this.form.versions.map((version) => version.version),
+          'All'
         );
         this.versionSelected = this.versions[0];
         await this.refreshFormFields(this.versionSelected);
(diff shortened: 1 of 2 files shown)


WalterMoar:
When selecting "All" versions it isn't exporting the data if form fields change.

image

1. Published a new form with text field `simpletextfield`
2. Created a submission
3. Published a new version of the form, but renamed `simpletextfield` to `simpletextfield2`
4. Created a submission
5. Exported "All" versions

I don't know if it is possible for us to export "all" versions to a single file. The fields could be completely different between versions of the form.

timisenco2015:
> When selecting "All" versions it isn't exporting the data if form fields change.
> 
> image
> 
> 1. Published a new form with text field `simpletextfield`
> 2. Created a submission
> 3. Published a new version of the form, but renamed `simpletextfield` to `simpletextfield2`
> 4. Created a submission
> 5. Exported "All" versions
> 
> I don't know if it is possible for us to export "all" versions to a single file. The fields could be completely different between versions of the form.

Good morning. I saw your comment, Walter. Yes, that was why we did not add the "All" initially, but since users can use the API directly by not specifying the form version number, it might not be a bad idea to add it to the front end. That is why I did make it "all" a default value

timisenco2015:
> When selecting "All" versions it isn't exporting the data if form fields change.
> 
> image
> 
> 1. Published a new form with text field `simpletextfield`
> 2. Created a submission
> 3. Published a new version of the form, but renamed `simpletextfield` to `simpletextfield2`
> 4. Created a submission
> 5. Exported "All" versions
> 
> I don't know if it is possible for us to export "all" versions to a single file. The fields could be completely different between versions of the form.

Yes, that was why we did not add the "All" initially, but since users can use the API directly by not specifying the form version number, it might not be a bad idea to add it to the front end. That is why I did make it "all" a default value.

WalterMoar:
> > When selecting "All" versions it isn't exporting the data if form fields change.
> > image
> > 
> > 1. Published a new form with text field `simpletextfield`
> > 2. Created a submission
> > 3. Published a new version of the form, but renamed `simpletextfield` to `simpletextfield2`
> > 4. Created a submission
> > 5. Exported "All" versions
> > 
> > I don't know if it is possible for us to export "all" versions to a single file. The fields could be completely different between versions of the form.
> 
> Yes, that was why we did not add the "All" initially, but since users can use the API directly by not specifying the form version number, it might not be a bad idea to add it to the front end. That is why I did make it "all" a default value.

Is there a way to include all fields from all versions? I tried another scenario:

1. Published a new form with text field `field1`
2. Created a submission
3. Published a new version of the form, adding `field2`
4. Created a submission
5. Exported "All" versions

image

It's only including `field`. So it looks like any new fields added to new versions of the form will never be exported.

I think we're going to get a lot of bug reports about missing fields from the exports.

Issue #536, opened 2023-02-08
CSV Export places 


MonicaG:
#### Describe the bug

Data is not placed in the correct columns when exporting submissions to CSV.  The columns affected are those that were conditionally displayed. 

#### To Reproduce

1. Create a form that contains fields that use conditional logic. 
2. Create form submissions where there are submissions that use the conditional fields and submissions that don't
3. Export the data as CSV
4. Open the CSV file. The conditional field data will be placed in the wrong column

#### Expected behavior

Data should appear in the correct columns

#### Screenshots

If you need example data I can share the exported CSV file.

#### Desktop (please complete the following information)

- OS: MacOS Ventura 13.1
- Browser: Safari
- Version: 16.2

#### Additional context

I was able to get around the problem by exporting the data as json. I then used the [jq command line tool](https://stedolan.github.io/jq/) to convert it to CSV via the following command:

```bash
cat submissions.json | jq -r '(. | map(leaf_paths) | unique) as $cols | map (. as $row | ($cols | map(. as $col | $row | getpath($col)))) as $rows | ([($cols | map(. | map(tostring) | join(".")))] + $rows) | map(@csv) | .[]' > output.csv
```
[Reference](https://stackoverflow.com/a/69230819/4704303)

I wanted to share this work around incase anyone else needs it.

WalterMoar:
Hello @MonicaG - This problem was fixed recently in #892

PR #892, opened 2023-07-14
FORMS-784: Fix for fields filtering for single row export option 3

bcvesalink:
# Description
The fix which adding the filtering fields upon the export for Option #3 `Single row export`

## Types of changes

## Checklist


app/src/forms/form/exportService.js
```diff
@@ -89,7 +90,7 @@ const service = {
     return await flattenComponents(schema.components);
   },
 
-  _buildCsvHeaders: async (form, data, version, fields) => {
+  _buildCsvHeaders: async (form, data, version, fields, singleRow = false) => {
     /**
      * get column order to match field order in form design
      * object key order is not preserved when submission JSON is saved to jsonb field type in postgres.
```
WalterMoar, line 96: Hi Sergey - I noticed this comment about matching the field order to the form design. However, when I export for "WALTER - TEST 1" (added you as owner) the columns are not in the same order as the form. Do you think this will matter to the users?

WalterMoar:
I found a small problem with datagrids: the number of columns for a datagrid only depends on the last submission. To reproduce:

1. Have a form with a datagrid containing a single text field

2. Submit with three rows in the datagrid, then export
image

3. Submit with one row in the datagrid, then export
image

Is it only looking at the final submission for the fields? I think it will need to look at all submissions, which may also fix https://github.com/bcgov/common-hosted-form-service/issues/536

app/src/forms/form/exportService.js
bcvesalink: hi Walter, good point, I'll see if npm package I used handles sorting

WalterMoar: Or maybe it's the Object.keys() that is losing the order? Tricky!

WalterMoar (approved):
Looks good and fixes GitHub Issue 536 too!

WalterMoar (approved):

PR #928, opened 2023-08-05
Fix form columns missing from selection

timisenco2015:
# Description
If users deselect default columns  (Submission Date, Submitter, Status) but select other columns, then click on save; on returning to the select columns dialog box, the default columns no longer appear in the columns list. Clients reported this bug, and this PR will fix the bug.

## Types of changes

## Checklist


app/frontend/src/components/base/BaseFilter.vue
jasonchung1871, line 1: If you look at the Vue3 changes for this file, you'll see that we don't watch the selectedData and modify the selected data. Is there any reason in particular to add custom logic to the filter here when it can be added to the table that needs it?

jasonchung1871: What I mean is, is this something that needs to be done in the BaseFilter? The data should be sorted prior to being sent to the BaseFilter. Does there need to be sorting in all the pages? (Forms/My Submissions/Export Data)

timisenco2015: Yes.

timisenco2015: Screenshot 2023-08-08 140528

timisenco2015: When a user selects a field, it will move to the top. That is what the sorting is for.

jasonchung1871: I don't think there's a way to turn it off except by removing it for every other page since it's inside of the BaseFilter. Why not pass in the custom sort as a prop function, and use that instead of a method inside of the BaseFilter. That way if it's set to null outside of the BaseFilter then it will use the default Vuetify sort. It does create some code duplication unless we have a utility function from another file.

PR #948, opened 2023-08-15
bug: FORMS-831 fix multiselect submission copying

bcgov-citz-ccft:
# Description

When a form has a multiselect and the user is allowed to copy from an existing submission, a select field with multiple values is not copying correctly.

Expected Behaviour:
image

Current Behaviour:
image

It was also noticed during troubleshooting that all multiselects are being copied, even when they shouldn’t be - will also fix.

Expected Behaviour:
image

Current Behaviour:
image

## Types of changes

## Checklist

### DIFF
--- a/app/frontend/src/components/designer/FormViewer.vue
+++ b/app/frontend/src/components/designer/FormViewer.vue
@@ -331,9 +331,12 @@ export default {
         for (let property in obj) {
           const innerObject = obj[property];
 
-          // When the form contains a Data Grid there will be an array that
-          // needs to be checked, and an array of properties to be unset.
-          if (Array.isArray(innerObject)) {
+          if (propNeeded === property) {
+            fields = fields + stack + '.' + property;
+            return fields.replace(/^\./, '');
+          } else if (Array.isArray(innerObject)) {
+            // When the form contains a Data Grid there will be an array that
+            // needs to be checked, and an array of properties to be unset.
             const fieldsArray = [];
             for (let i = 0; i < innerObject.length; i++) {
               const next = iterate(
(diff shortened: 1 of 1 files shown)


app/frontend/src/components/designer/FormViewer.vue
```diff
@@ -331,9 +331,12 @@ export default {
         for (let property in obj) {
           const innerObject = obj[property];
 
-          // When the form contains a Data Grid there will be an array that
-          // needs to be checked, and an array of properties to be unset.
-          if (Array.isArray(innerObject)) {
+          if (propNeeded === property) {
+            fields = fields + stack + '.' + property;
+            return fields.replace(/^\./, '');
```
bcgov-citz-ccft, line 336: The problem where the multiselects always being copied is because both datagrids and multiselects appear as Arrays. The fix was to move the `propNeeded === property` clause to come before the Array check for datagrids.

WalterMoar, line 336: The problem where the multiselects always being copied is because both datagrids and multiselects appear as Arrays. The fix was to move the `propNeeded === property` clause to come before the Array check for datagrids.

app/frontend/src/components/designer/FormViewer.vue
```diff
@@ -1181,9 +1181,10 @@ export default {
   },
   async created() {
     if (this.submissionId && this.isDuplicate) {
-      //Run when make new submission from existing one called.
+      // Run when make new submission from existing one called. Get the
+      // published version of form, and then get the submission data.
+      await this.getFormSchema();
       await this.getFormData();
-      await this.getFormSchema(); //We need this to be called as well, because we need latest version of form
```
WalterMoar, line 1186: The previous code loaded the data and then the form - for some reason this was causing the last item in a multiselect to be `undefined`.

jasonchung1871 (approved):

PR #966, opened 2023-08-21
  Forms-764_server_side_pagination_for_data_tables

timisenco2015:
# Description
This feature replaces the front-side (front-end) pagination with the server-side (backend) pagination. Instead of loading all the data (e.g. submissions, which can be thousands of submissions) to the front-end and using Vuetify v-data-table built pagination functionality to handle pagination, the backend now handles the pagination (range of data is loaded to the frontend per request).

## Checklist


bcvesalink:
@timisenco2015 I think API versioning was also part of this story, we need to make it so we still keep old api call without changes

timisenco2015:
> @timisenco2015 I think API versioning was also part of this story, we need to make it so we still keep old api call without changes

User can still this API without the Pagination. I put a check that check if using pagination or as default

WalterMoar (changes requested):
I think one of the really hard problems is that we don't want to remove the ability to sort the data in the table. So the API is going to have to know how to sort based on things within the form. And it will also have to paginate on those as well. I think we'll have unhappy users if we remove this.
image

bcvesalink:
> I think one of the really hard problems is that we don't want to remove the ability to sort the data in the table. So the API is going to have to know how to sort based on things within the form. And it will also have to paginate on those as well. I think we'll have unhappy users if we remove this. image

yeah, we definitely should keep sorting functionality

WalterMoar:
The data is looking good! However, it looks like there is something slightly wrong when sorting by submission date:
image

WalterMoar:
I found a problem where if there are no submissions then clicking on some of the header items will produce 500 errors in the backend.
image

WalterMoar:
When I click any of the column headers, I do find it surprising that the page goes back to the first page (if I have more than 10 submissions, that is):

1. Create a form and 11 submissions
2. In the submissions click the `Confirmation ID`
3. Go to the second page of results
4. Click the `Confirmation ID`
5. Goes back to the first 10 results

However, in most cases it doesn't make sense to be part-way through the submissions and then change the sort order! I'm not really sure what would be best here.

timisenco2015:
> When I click any of the column headers, I do find it surprising that the page goes back to the first page (if I have more than 10 submissions, that is):
> 
> 1. Create a form and 11 submissions
> 2. In the submissions click the `Confirmation ID`
> 3. Go to the second page of results
> 4. Click the `Confirmation ID`
> 5. Goes back to the first 10 results
> 
> However, in most cases it doesn't make sense to be part-way through the submissions and then change the sort order! I'm not really sure what would be best here.

It sorts the whole submissions at the backend, except we want to go for the page sorting and not the entire submissions sorting

WalterMoar (approved):

PR #995, opened 2023-09-08
FORMS-852 - CSV option 3 export is missing data

bcvesalink:
# Description
A user reported that when they export using CSV option 3, some of their data is missing. One thing of note with this form is that the missing data is select lists that are populated with raw JSON, and the value of the select is the object for the selection.

This PR will solve the issue when we had discrepancy in fields names and array for multi and single row options so thus export field filter was working incorrectly

## Types of changes

## Checklist


timisenco2015 (commented):
If I may suggest that we should move "CSV Format" above "Data Fields" since any selection in the "CSV Format" changes the fields in the "Data Fields.". We can discuss this with the UI team.
Screenshot 2023-09-10 214154

app/tests/unit/forms/form/exportService.spec.js
```diff
@@ -247,6 +247,94 @@ describe('_buildCsvHeaders', () => {
...
+
+    expect(result).toHaveLength(20);
+    expect(result).toEqual(
+      expect.arrayContaining([
+        'form.confirmationId',
+        'oneRowPerLake.0.closestTown',
+        'oneRowPerLake.0.dataGrid.0.fishType',
+        'oneRowPerLake.0.dataGrid.1.fishType',
+        'oneRowPerLake.0.dataGrid.0.numberKept',
+        'oneRowPerLake.0.dataGrid.1.fishType',
```
WalterMoar, line 280: Looks like `'oneRowPerLake.0.dataGrid.1.fishType'` is repeated, is this correct? Or is it

```
        'oneRowPerLake.0.dataGrid.1.numberKept',
```

WalterMoar (approved):
Looks good! @timisenco2015 has a really good point that since the CSV type changes the fields list, the CSV type should go above the fields list. But with the Vue 3 upgrade coming so soon, maybe let's not change the UI too much!

PR #1007, opened 2023-09-14
FORMS-844: Add ability to subset submissions for export

bcvesalink:
# Description
Requested By (and initially reported as a [UX bug ](https://bcdevex.atlassian.net/browse/FORMS-838)via)
As a CHEFS stakeholder, I need to be able to

-  define a subset of submissions for a given form version
-  export that subset of submissions to a CSV-formatted file

## Types of changes

## Checklist


app/frontend/src/store/modules/form.js
```diff
@@ -125,6 +126,7 @@ export default {
     isRTL: (state) => state.isRTL,
     subscriptionData: (state) => state.subscriptionData,
     totalSubmissions: (state) => state.totalSubmissions,
+    selectedSubmissions: (state) => state.selectedSubmissions,
```
WalterMoar, line 129: I think we need to start improving the code coverage of the unit tests. Can you check the coverage reports in the actions (or run the tests locally and check app/frontend/coverage/lcov-report/index.html and app/coverage/lcov-report/index.html) to make sure there are tests for anything that is added?

(Although a lot of the tests are fairly simple and don't add a lot of value, it'll make it easier to identify missing coverage if we follow what the original team did for tests)

WalterMoar:
I think one of the problems we're going to have is that if I have v1 and v2 submissions, I can't tell from the submissions table that they're different versions. So if I select submissions from different versions, and then choose a CSV export, not all of the submissions I selected will be exported.

WalterMoar:
> I think one of the problems we're going to have is that if I have v1 and v2 submissions, I can't tell from the submissions table that they're different versions. So if I select submissions from different versions, and then choose a CSV export, not all of the submissions I selected will be exported.

Maybe a question for Jordan and Manadir, but maybe a new feature would be to remove the "Version" dropdown (or allow it to be multiselect) and download a .zip file of CSVs if there are multiple versions? Just a wild idea!

bcvesalink:
> > I think one of the problems we're going to have is that if I have v1 and v2 submissions, I can't tell from the submissions table that they're different versions. So if I select submissions from different versions, and then choose a CSV export, not all of the submissions I selected will be exported.
> 
> Maybe a question for Jordan and Manadir, but maybe a new feature would be to remove the "Version" dropdown (or allow it to be multiselect) and download a .zip file of CSVs if there are multiple versions? Just a wild idea!

agreed, Manadir and Jordan should be involved here

bcvesalink:
> > I think one of the problems we're going to have is that if I have v1 and v2 submissions, I can't tell from the submissions table that they're different versions. So if I select submissions from different versions, and then choose a CSV export, not all of the submissions I selected will be exported.
> 
> Maybe a question for Jordan and Manadir, but maybe a new feature would be to remove the "Version" dropdown (or allow it to be multiselect) and download a .zip file of CSVs if there are multiple versions? Just a wild idea!

@WalterMoar actually, I thought it might be a good idea to not even filter by version if selected submission ids specified, so it'd more convenient for users, they select needed submissions from the list and expect all of them to be imported no matter what version they belong to, what do you think?

WalterMoar:
Closing this - I think we need to look at redesigning the whole CSV export.

PR #1030, opened 2023-09-20
fix: experimental tooltips timing delays

timisenco2015:
# Description
Tooltip timing problems occur when the user transitions away from the tooltip content, but the tooltip itself remains visible for an additional 4 seconds

## Types of changes

## Checklist


WalterMoar:
It looks like the reason for those `close-delay` settings was to make it easier to click the links in the tooltips. In the deploy for this PR, it's fairly difficult to click the links because the tooltips now disappear when you try to move your mouse into them.

Is there any way to have the `close-delay` but have all other tooltips disappear when a new one is shown? If that's too difficult, then maybe a shorter delay would work?

PR #1033, opened 2023-09-21
FORMS-867: Export CSV submission columns issue (new validation for 'form' keyword in fieldnames)

bcvesalink:
# Description

This PR is to fix use cases when user would name one of the fields as `form` which is system keyword, so we added additional validation so users can't use this API name for fields

## Types of changes

## Checklist


app/frontend/src/internationalization/trans/chefs/en/en.json
```diff
@@ -325,7 +325,8 @@
     "version": "Version",
     "save": "Save",
     "saving": "Saving",
-    "notSaved": "Not Saved"
+    "notSaved": "Not Saved",
+    "fieldname_error": "You CAN NOT use 'form' keyword as {label} fieldname"
```
WalterMoar, line 329: This is fairly minor, but so far we've never used an underscore in a translation key. For consistency can you change this to `fieldnameError`?

app/frontend/src/components/designer/FormDesigner.vue
```diff
@@ -529,6 +532,18 @@ export default {
               this.addPatchToHistory();
             }
           }
+          this.canSave = true;
+          modified?.components?.map((comp) => {
+            if (comp.key === 'form') {
```
WalterMoar, line 537: My concern here is that it's a breaking change. A form that was fine before this PR will produce an error when edited and saved after this PR, even if there are no changes to the form.

Maybe this isn't a big deal - how hard would it be to identify how many forms already in production use "form" as one of the field names? Maybe we could work with those form owners to fix their forms before the change.

WalterMoar (changes requested):

app/frontend/src/internationalization/trans/chefs/en/en.json
bcvesalink: @WalterMoar addressed

app/frontend/src/components/designer/FormDesigner.vue
bcvesalink: @WalterMoar there are only 2 published forms which have `form` as a field name, should we just notify them?

WalterMoar (approved):

PR #1048, opened 2023-09-27
FORMS-881 - Not allow to save drafts if form validation fails

bcvesalink:
# Description
This fix is regarding reported bug https://bcdevex.atlassian.net/wiki/spaces/CCP/pages/1218969620/Invalid+data+remains+in+the+component+without+any+error+message+after+saving+and+then+reopening+a+draft

## Types of changes

## Checklist

### DIFF
--- a/app/frontend/src/components/designer/FormViewer.vue
+++ b/app/frontend/src/components/designer/FormViewer.vue
@@ -37,7 +37,6 @@
           <FormViewerActions
             :block="block"
             :draftEnabled="form.enableSubmitterDraft"
-            :formIsValid="formIsValid"
             :formId="form.id"
             :isDraft="submissionRecord.draft"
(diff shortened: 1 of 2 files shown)


bcvesalink:
> This might cause problems because it forces the user to complete the form before saving. For example, if the form has 100 required fields, I might want to fill in the first 20 fields and then save it as a draft to be completed later when I have time. With this change, though, I would be forced to fill in all 100 fields, so I might as well submit the form instead of saving it as a draft.
> 
> I'm guessing there's no easy way to only prevent the saving of draft if there are validation errors in fields that have been filled in, and ignoring empty fields?

Good point @WalterMoar I'll see what I can do here...

WalterMoar (approved):
Looks great!!! :)

PR #1064, opened 2023-10-06
Fix/data table no search results

WalterMoar:
# Description

1. Fixed the `noMachingRecordText` translation keys to be `noMatchingRecordText`.

***

2. Added the missing Russian `mySubmissionsTable.noMatchingRecordText` translation.

***

3. Fixed the MySubmissionsTable to display the correct message when a search is done but there are no matching results. Before:
image
After:
image

***

4. Fixed the TeamManagement to display the correct message when a search is done but there are no matching results.
Before:
image
After:
image

## Types of changes

## Checklist

### DIFF
--- a/app/frontend/src/internationalization/trans/chefs/en/en.json
+++ b/app/frontend/src/internationalization/trans/chefs/en/en.json
@@ -665,6 +665,7 @@
     "updatedBy": "Modified By"
   },
   "teamManagement": {
+    "noMatchingRecordText": "No matching records found",
     "teamManagement": "Team Management",
     "selectColumns": "Select Columns",
     "manageForm": "Manage Form",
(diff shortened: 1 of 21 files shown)


app/frontend/src/internationalization/trans/chefs/en/en.json
```diff
@@ -391,6 +391,7 @@
     "emailRequired": "Email is required"
   },
   "submissionsTable": {
+    "noMatchingRecordText": "No matching records found",
```
WalterMoar, line 394: I re-ordered the items in the en translation so that it's the same order as the other files. This makes it easier to compare files to check for inconsistencies.

I'm thinking, though, that we should do some sort of automation here, like alphabetizing the keys. That way it would always be easy to compare files for missing/extra/different keys.

jasonchung1871 (approved):

app/frontend/src/internationalization/trans/chefs/en/en.json
jasonchung1871: I think this should be done inside the main branch, but yes, I agree with you. Would make it much easier to search through the localization files.

PR #1076, opened 2023-10-11
fix: API repeated call when searching in MySubmission page

timisenco2015:
# Description

## Types of changes

## Checklist


app/frontend/src/components/forms/SubmissionsTable.vue
```diff
@@ -41,7 +41,6 @@ export default {
           key: 'event',
         },
       ],
-      forceTableRefresh: 0,
```
jasonchung1871, line 44: Could this be returned to it's original position? The variable names should be sorted in alphabetical order for consistency.

app/frontend/src/components/forms/SubmissionsTable.vue
```diff
@@ -562,8 +561,15 @@ export default {
       });
       await this.populateSubmissionsTable();
     },
-    handleSearch(value) {
+    async handleSearch(value) {
+      this.searchEnabled = true;
       this.search = value;
+      if (value === '') {
+        this.searchEnabled = false;
```
jasonchung1871, line 568: This searchEnabled is not needed since we know it's disabled if the length of the value is greater than 0.

app/frontend/src/components/forms/SubmissionsTable.vue
jasonchung1871, line 1: I've tried adding this to my local environment and it'll make many calls instead of just one since debounce is no longer being used to prevent multiple calls. handleSearch calls refreshSubmissions for any changes to the input.

jasonchung1871 (changes requested):

app/frontend/src/components/forms/SubmissionsTable.vue
timisenco2015: We are using it at the backend.

app/frontend/src/components/forms/SubmissionsTable.vue
timisenco2015: Okay

app/frontend/src/components/forms/SubmissionsTable.vue
timisenco2015: We are still using the debounce in the handleSearch. The debounce is only used for the search. The handleSearch calls the refreshSubmissions if the search is empty. It is a way to reset the submission list to default when the user clears the search input.

app/frontend/src/components/forms/SubmissionsTable.vue
jasonchung1871: What I meant is that we don't need to create a variable to check whether the search is enabled because we can figure out if search is enabled just by checking if the search string is not empty.

app/frontend/src/components/forms/SubmissionsTable.vue
jasonchung1871: I understand but in my local environment, with these changes, when you type in ASDF very quickly it will make a search on the back end for A and then a search for AS and then a search for ASD and then a search for ASDF.

timisenco2015: I have found the fix for that. I will push the fix now

app/frontend/src/components/forms/SubmissionsTable.vue
timisenco2015: Okay👌

jasonchung1871 (approved):

PR #1083, opened 2023-10-12
feat: FORMS-882 add submissionId to exports

bcgov-citz-ccft:
# Description

We have a Fider request for the submission ID to be added to the JSON exports. We are already exporting the confirmation ID, which is 8 characters long and taken from the submission ID. The requested change is because the submission ID is essentially unique, but the confirmation ID is much more likely to have a collision. Users who are doing ETL with the exports would greatly prefer to have the more-unique key for the submission.

Since we’re adding the submission ID to the JSON export, it should also be added to the CSV export.

https://chefs-fider.apps.silver.devops.gov.bc.ca/posts/112/data-export-missing-formsubmissionid

## Types of changes

## Checklist

## Further comments

Luckily this is already in the view, so it's just a matter of adding it to the columns pulled by the model.

### DIFF
--- a/app/src/forms/form/exportService.js
+++ b/app/src/forms/form/exportService.js
@@ -200,7 +200,7 @@ const service = {
 
   _submissionsColumns: (form, params) => {
     // Custom columns not defined - return default column selection behavior
-    let columns = ['confirmationId', 'formName', 'version', 'createdAt', 'fullName', 'username', 'email'];
+    let columns = ['submissionId', 'confirmationId', 'formName', 'version', 'createdAt', 'fullName', 'username', 'email'];
     // if form has 'status updates' enabled in the form settings include these in export
     if (form.enableStatusUpdates) {
       columns = columns.concat(['status', 'assignee', 'assigneeEmail']);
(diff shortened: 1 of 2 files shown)


WalterMoar:
Thoughts on how this could break things for the users?

I suppose one possibility is that people have code to parse CSV files and they depend on column position - not a very robust way of doing things! I think adding this column would break any sort of fragile CSV-parsing process.

WalterMoar:
> Thoughts on how this could break things for the users?
> 
> I suppose one possibility is that people have code to parse CSV files and they depend on column position - not a very robust way of doing things! I think adding this column would break any sort of fragile CSV-parsing process.

I posted in both the Teams and rocket.chat channels about the upcoming change and there were no responses. I think that this possibly-breaking-change is an unlikely case and we should go ahead with this as-is.

jasonchung1871 (approved):

WalterMoar:
Today's my Friday so I'll merge this first thing Monday morning.

timisenco2015 (approved):

PR #1091, opened 2023-10-17
fix: database view performance

bcgov-citz-ccft:
# Description

Very slow performance of the database indicates that the view `user_form_access_vw` is running very slowly. This view is composed of multiple other views, two of which combine every form versus every user. There are a few things that can be done to improve the performance (and readability) of these views.

1. Remove the join of "role" in user_form_permissions_vw.
2. Remove the join of "role" in user_form_roles_vw.
3. Simplify the EXISTS SELECT in user_form_permissions_vw.
4. Simplify the EXISTS SELECT in user_form_roles_vw.
5. Remove the sorting in the user_form_access_vw.

## Types of changes

## Checklist

### DIFF
--- a/app/src/db/migrations/20231017192656_037-user-form-permissions.js
+++ b/app/src/db/migrations/20231017192656_037-user-form-permissions.js
@@ -1,9 +1,9 @@
-// Having performance problems and these views seem to be a cause of it. The
-// EXPLAIN for user_form_access_vw says that the role table in
-// user_form_permissions_vw is the bottleneck. Since it isn't needed, just get
-// rid of it rather than try to optimize it.
-//
-// Note that we have to remove and recreate the dependent user_form_access_vw.
+// Having performance problems and these views seem to be a cause of it.
+// 1. Remove the join of "role" in user_form_permissions_vw.
+// 2. Remove the join of "role" in user_form_roles_vw.
+// 3. Simplify the EXISTS SELECT in user_form_permissions_vw.
+// 4. Simplify the EXISTS SELECT in user_form_roles_vw.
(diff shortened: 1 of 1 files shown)


bcvesalink:
@WalterMoar do you think we should remove expensive sorting in auth api as well? Here  https://github.com/bcgov/common-hosted-form-service/blob/master/app/src/forms/auth/service.js#L136 and here https://github.com/bcgov/common-hosted-form-service/blob/master/app/src/forms/auth/service.js#L141 ?

WalterMoar:
> @WalterMoar do you think we should remove expensive sorting in auth api as well? Here https://github.com/bcgov/common-hosted-form-service/blob/master/app/src/forms/auth/service.js#L136 and here https://github.com/bcgov/common-hosted-form-service/blob/master/app/src/forms/auth/service.js#L141 ?

We should definitely think about it. It won't affect the database load, though - only the load on the app pods. I think we have a lot of room for improvement, so let's keep that in mind!

bcvesalink:
> > @WalterMoar do you think we should remove expensive sorting in auth api as well? Here https://github.com/bcgov/common-hosted-form-service/blob/master/app/src/forms/auth/service.js#L136 and here https://github.com/bcgov/common-hosted-form-service/blob/master/app/src/forms/auth/service.js#L141 ?
> 
> We should definitely think about it. It won't affect the database load, though - only the load on the app pods. I think we have a lot of room for improvement, so let's keep that in mind!

I mean, we removed sorting in view, why would we need sorting in API?

jasonchung1871 (approved):

PR #1111, opened 2023-10-24
fix: FORMS-903 choicesjs select visibility

bcgov-citz-ccft:
# Description

There’s a bug with the Select component that if the default Widget Type of ChoicesJS is used, then the items in the list are not visible when used in components like Data Grids:

image

The workaround has been to switch the Widget Type to HTML 5. There is a different bug with that widget though: https://github.com/formio/formio.js/issues/5390

Implement the fix described in: https://github.com/formio/formio.js/issues/5216#issuecomment-1591282850 

and leave a comment that it should be re-evaluated when formio.js is updated.

After:
image

## Types of changes

## Checklist

### DIFF
--- a/app/frontend/src/assets/scss/style.scss
+++ b/app/frontend/src/assets/scss/style.scss
@@ -444,6 +444,18 @@ a,
     label.field-required.control-label--hidden::after {
       content: '';
     }
+
+    // We can't use HTML 5 selects due to:
+    // https://github.com/formio/formio.js/issues/5390
+    //
+    // So use the fix described in:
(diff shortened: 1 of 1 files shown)


timisenco2015 (approved):

WalterMoar:
This is going to remove the horizontal scroll bar from really wide data grids. Their fix for that broke the select lists, but I think we need to find a solution for both problems. Changing this to draft for now.

jasonchung1871 (approved):

WalterMoar:
Closing this - need to revisit at some point, but not now.

PR #1131, opened 2023-11-02
FORMS 897 - GET /files API Access + File Storage Bug

vijaivir:
# Description

This PR introduces the functionality for GET requests on the /files endpoint to be accessible using an API Key. The feature was requested by JEDI and has been tested in this PR environment to be working as expected. 

Bug: This PR also addresses an issue that was causing some files to have a null submission id value, resulting in these files not being added to the submission folder.  

###  Code changes
`apiAccess.js`: Added an else if statement that checks weather the request parameter matches `id`, then retrieves the submissionId and formId respectively. 

`filePermissions.js`: added support for apiUser, granting `hasFileCreate` and `hasFilePermissions` to API users in the middleware. 

`routes.js`: mounted API Access on the GET /files route and added rate limiting. 

`service.js`: The `_findFileIds` function only checked the top level components for the files, but these can be nested within a form. I've added a recursive function to handle such nested cases as well. 

This change allows API users to directly download attached files in submissions using their API key instead of having to manually download each file from each submission. Additionally, it addresses a bug which is preventing files from being stored in the submissions folder in our database. 

https://bcdevex.atlassian.net/jira/software/c/projects/FORMS/boards/10?selectedIssue=FORMS-897
https://chefs-fider.apps.silver.devops.gov.bc.ca/posts/42/allow-attachments-to-be-downloaded-over-an-api

## Types of changes

## Checklist


jimmyPasta:
Thanks for developing this fix. I am glad that file upload in containers now properly retrieves file ids when in a container.

app/src/forms/file/routes.js
```diff
@@ -13,7 +14,7 @@ routes.post('/', middleware.publicRateLimiter, hasFileCreate, fileUpload.upload,
   await controller.create(req, res, next);
 });
 
-routes.get('/:id', currentFileRecord, hasFilePermissions(P.SUBMISSION_READ), async (req, res, next) => {
+routes.get('/:id', middleware.publicRateLimiter, apiAccess, currentFileRecord, hasFilePermissions(P.SUBMISSION_READ), async (req, res, next) => {
```
WalterMoar, line 17: I'm afraid that my latest commit is going to break this - can you update your branch and use the apiKeyRateLimiter here?

WalterMoar: In case you don't have it: the spec is deployed at `http://localhost:8080/app/api/v1/docs`

vijaivir: Thank you, I was wondering about the new rate limiting so this was really helpful! I have now updated the code and the api docs.

app/src/forms/form/service.js
```diff
@@ -30,16 +30,23 @@ const Rolenames = [Roles.OWNER, Roles.TEAM_MANAGER, Roles.FORM_DESIGNER, Roles.S
...
-      schema.components
-        // Get the file controls
-        .filter((x) => x.type === 'simplefile')
-        // for the file controls, get their respective data element (skip if it's not in data)
-        // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap#for_adding_and_removing_items_during_a_map
-        .flatMap((x) => (data.submission.data[x.key] ? data.submission.data[x.key] : []))
-        // get the id from the data
-        .map((x) => x.data.id)
-    );
+    const findFiles = (components) => {
```
WalterMoar, line 33: Curious about if you have any ideas how we can test this sort of thing. Not that the tests are needed now, because we don't really have a good plan / philosophy about how we want to do testing, but something to think about! We're not able to test everything that someone could do with a form, but I wonder what the "good enough" approach will be!

app/src/docs/v1.api-spec.yaml
```diff
@@ -1421,6 +1421,10 @@ paths:
     get:
       summary: Get a file
       operationId: fileGet
+      security:
+        - BasicAuth: []
+        - BearerAuth: []
+          OpenID: []
```
WalterMoar, line 1427: Great! Can you also put in the HTTP 429 error? There are other examples in the file. Basically any endpoints that allow API Key access should have the rate limiting applied, and also document the 429 responses so that people building apps will _hopefully_ handle them!

app/src/forms/form/service.js
vijaivir: I agree that we should have a testing plan, I can take on this task if you want. In this case though the function is being tested pretty well in app/tests/unit/forms/form/service.spec.js.

WalterMoar: No need to worry about it now - we'll get there eventually, and we'll probably be focused on frontend tests for awhile.

app/src/forms/file/middleware/filePermissions.js
```diff
@@ -13,7 +13,8 @@ const currentFileRecord = async (req, res, next) => {
   let fileRecord = undefined;
   try {
     // Check if authed, can expand for API key access if needed
-    if (req.params.id && req.currentUser) {
+    if (req.params.id && (req.currentUser || req.apiUser)) {
+      // expanded for api user
```
WalterMoar, line 17: Can you please add tests in `app/tests/unit/forms/file/middleware/filePermissions.spec.js`? After running the tests you can check coverage in `app/coverage/lcov-report/index.html`

app/src/forms/auth/middleware/apiAccess.js
```diff
@@ -21,6 +22,15 @@ module.exports = async (req, res, next) => {
...
         formId = result?.form?.id;
+      } else if (params.id && uuidValidate(params.id)) {
+        // get submissionID from request body
+        const sid = await fileService.read(params.id);
+        //check to see that an associated submissionId exists
+        if (!sid || !sid.formSubmissionId) {
+          throw new Error('Submission ID not found in file storage.');
+        }
+        const result = await submissionService.read(sid.formSubmissionId);
+        formId = result?.form?.id;
```
WalterMoar, line 33: Please add tests for this to `app/tests/unit/forms/auth/middleware/apiAccess.spec.js`.

app/src/forms/file/routes.js
```diff
@@ -1,18 +1,20 @@
...
 routes.use(currentUser);
 
 routes.post('/', hasFileCreate, fileUpload.upload, async (req, res, next) => {
   await controller.create(req, res, next);
 });
 
-routes.get('/:id', currentFileRecord, hasFilePermissions(P.SUBMISSION_READ), async (req, res, next) => {
+routes.get('/:id', rateLimiter, apiAccess, currentFileRecord, hasFilePermissions(P.SUBMISSION_READ), async (req, res, next) => {
   await controller.read(req, res, next);
 });
```
WalterMoar, line 19: Don't worry about tests here - the route tests are a mess and I'll cover these ones eventually :)

WalterMoar (changes requested):

app/src/forms/auth/middleware/apiAccess.js
vijaivir: I've added some tests now!

app/src/forms/file/middleware/filePermissions.js
vijaivir: I've added new tests for these. `filePermissions.js` and `apiAccess.js` are fully covered in the coverage report!

WalterMoar (approved):
Looks great!

PR #1143, opened 2023-11-08
fix: FORMS-912 increase rate limit and restore logging

bcgov-citz-ccft:
# Description

JEDI is hitting the rate limit, so bump it up from 20 to 60.

Also, by moving the rate limiting to the app as a whole, it had two effects:
1. The static code analysis (CodeQL and GitHub Security) can’t detect that the rate limiting is being applied
2. The 429 errors are not logged - missed this in the testing, but it’s important

So moved it back to the routes from app.js

## Types of changes

## Checklist

## Further comments

Bad news is that moving the rate limiting to the routes will cause the rate limiting to be used by the tests - this was breaking the tests once they started returning 429 instead of the expected response. Made the tests for endpoints that allow API Keys to only use Bearer tokens.

Good news - now I know how to test the rate limiting. No time for it now, but will do ASAP as tech debt.

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -8,7 +8,6 @@ const querystring = require('querystring');
 const keycloak = require('./src/components/keycloak');
 const log = require('./src/components/log')(module.filename);
 const httpLogger = require('./src/components/log').httpLogger;
-const rateLimiter = require('./src/forms/common/middleware');
 const v1Router = require('./src/routes/v1');
 
 const DataConnection = require('./src/db/dataConnection');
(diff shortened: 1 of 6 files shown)


WalterMoar:
> CodeQL found more than 10 potential problems in the proposed changes. Check the [Files changed](/bcgov/common-hosted-form-service/pull/1143/files) tab for more details.

It doesn't like my fake "Bearer: abc123" tokens - I'll try to find a way around this in a future PR.

WalterMoar:
Can be tested with the following, but remember that there are two API pods, so it should more-or-less round-robin them.

```bash
for i in {1..130}; do echo "Request #$i"; curl -i "https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-1143/api/v1/forms/FORMID/submissions" -u "FORMID:APIKEY"; echo; done;
```

PR #1160, opened 2023-11-23
fix: Forms-924 'send my team a notification' still sending a notification even if unselected

bcgov-citz-ccft:
# Description

This pull request aims to fix the "Send my team a notification" functionality. The problem with this feature was that if this setting gets turned on and the associated field is populated with an email address, and later the option gets unticked, emails would still get sent. This is due to the fact that the field associated with the entered email does not get cleared after unticking the checkbox. 

This fix involves adding a new column into the "Form" table to keep track of whether the form owner wants to receive emails to their team members after submissions or not. 

## Types of changes

## Checklist

### DIFF
--- a/app/frontend/src/components/designer/FormDesigner.vue
+++ b/app/frontend/src/components/designer/FormDesigner.vue
@@ -622,6 +622,7 @@ export default {
           idps: this.form.idps,
           userType: this.form.userType,
         }),
+        sendSubReceivedEmail: this.form.sendSubReceivedEmail,
         enableSubmitterDraft: this.form.enableSubmitterDraft,
         enableCopyExistingSubmission: this.form.enableCopyExistingSubmission,
         enableStatusUpdates: this.form.enableStatusUpdates,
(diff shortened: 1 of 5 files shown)


WalterMoar (changes requested):
I think that with there now being both email address(es) and the "enabled" flag, we can also have the behaviour where:
1. I set up my form with emails enabled and going to a team of 10 users.
2. For whatever reason I disable the emails (maybe form launch and expecting mega submissions)
3. Later on, when I re-enable emails, I would like to not have to fill in the 10 email addresses again.

Since we have the new flag, would it be much effort to do this? I think you'd just have to not clear the emails in the form and these rest should magically work?

---

Tests would be good to add! 

---

The API docs (app/src/docs/v1.api-spec.yaml) should also be updated to include the new field. The `submissionReceivedEmails` only exists once in the file so should be a small change.

WalterMoar (approved):
Looks great!

PR #1169, opened 2023-12-01
fix: FORMS-927 admin show deleted forms checkbox

bcgov-citz-ccft:
# Description

The “Show Deleted” checkbox on the Admin page’s list of forms is not working. It’s being assigned a value of false but the results being displayed are for true. So then when it is checked, the results are for true again (but page looks like false) and then when it’s unchecked the results are for false (but page looks like true). So the first click nothing changes in the data, but after that the logic for it is backwards.

## Types of changes

## Checklist

## Further comments

The actual fix was that the `@click` event was firing before the model was updated. We shouldn't really be doing two things when the state of the component changes - instead just set the data and put a watcher on that data.

The name of the field in the component is “activeOnly” while the text on the page for it is “Show Deleted”. This flipped logic is a little confusing, so I changed the data field to be “showDeleted” (with the same initial state, which was incorrect before!) and then flipped the logic everywhere else.

Changed the mounted code to call the same code we have for refreshing - let's just maintain one call to `getForms`.

### DIFF
--- a/app/frontend/src/components/admin/AdminFormsTable.vue
+++ b/app/frontend/src/components/admin/AdminFormsTable.vue
@@ -8,8 +8,8 @@ import { useFormStore } from '~/store/form';
 export default {
   data() {
     return {
-      activeOnly: false,
-      loading: true,
+      showDeleted: false,
+      loading: false,
       search: '',
     };
   },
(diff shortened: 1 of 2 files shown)


vijaivir (approved):
Tested the PR deployment, its working as expected!

vijaivir (approved):

PR #1172, opened 2023-12-05
fix: FORMS-932 validate submission id for read

bcgov-citz-ccft:
# Description

In the API if the formSubmissionId is an invalid value such as “undefined”, or “{{sumissionId}}” (sic) or a UUID with a trailing `.`, then the value is used in a query against the database. This query fails and produced an HTTP 403 response.

By doing the validation in the controller, we:

- Take load off the database
- Have fewer errors in the database logs. This is important because we want the logs to contain errors about the database, not errors about the mistakes being made by our users. Once we prevent the user mistakes from hitting the database we can begin to alert on database errors
- The 403 response is not really correct - this should be a 400 response

Do the read method of the controller as a proof of concept. This is the endpoint that sees the most errors due to the calling systems containing bugs. If everything works out then we can think about rolling this out to the other endpoints.

## Types of changes

## Checklist

## Further comments

Had to update the route tests because they were using an invalid UUID.

### DIFF
--- a/app/package-lock.json
+++ b/app/package-lock.json
@@ -45,11 +45,12 @@
         "winston": "^3.8.2"
       },
       "devDependencies": {
+        "@jest-mock/express": "^2.0.2",
         "eslint": "^8.40.0",
         "eslint-config-prettier": "^8.8.0",
         "eslint-config-recommended": "^4.1.0",
         "eslint-plugin-prettier": "^4.2.1",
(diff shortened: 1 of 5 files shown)


WalterMoar:
Closing this - it's something we should do but not very high value, plus it's a potentially breaking change (although only for those sending in bad requests). We can do a big cleanup of all this later but focus on other stuff now.

PR #1176, opened 2023-12-06
prevents export with no submissions

jasonchung1871:
# Description
If a reviewer is trying to export submissions with no submissions, it gives an infinite spinner. This will return an error to the user indicating that there are no submissions if there are no submissions. A valid submission is a form that is not in the draft state or deleted state.

resolves: https://bcdevex.atlassian.net/browse/FORMS-857

## Checklist


app/src/forms/form/controller.js
```diff
@@ -9,10 +9,15 @@ module.exports = {
   export: async (req, res, next) => {
     try {
       const result = await exportService.export(req.params.formId, req.query, req.currentUser, req.headers.referer);
-      ['Content-Disposition', 'Content-Type'].forEach((h) => {
-        res.setHeader(h, result.headers[h.toLowerCase()]);
-      });
-      return res.send(result.data);
+      if (result.data.length === 0) {
+        res.setHeader('Content-Type', 'application/json');
+        res.status(400).send(JSON.stringify({ detail: 'There are no submissions to export.' }));
```
WalterMoar, line 14: I think that if we change the response status it's going to affect API Key users who are expecting a 200 response and an empty set of data. It seems like this isn't a 400 "Bad Request" but just that the request is good but there is no data to return? Thoughts?

I think Sergey said that even if a CSV has no rows of data, in the file storage table it should still have a CSV with the column names - it shouldn't be 0 bytes. Maybe this is what's missing? (my memory is hazy!)

PR #1179, opened 2023-12-06
 Fix: GET /files API working for Objects and Arrays

vijaivir:
# Description
A user reported that the GET /files API does not retrieve files that are uploaded in containers. Upon further investigation, this was also the case for many other components such as data grids, columns, wells, panels, and tables. The error was in checking the form schema's components array, which does not contain all these elements.

The fix now uses the form data instead and handles the case for both arrays and objects to successfully retrieve files from within any component.

## Types of changes

## Checklist


WalterMoar (approved):
Looks good - I also tried all sorts of combos and it was working fine!

PR #1146, opened 2023-11-14
Add in publish form event subscription

jimmyPasta:
When a form is published the form id and form version id are sent to the subscribed endpoint.

image

Add in callouts to the existing endpoint subscription

Telling external applications when a form version is published informs the endpoint that changes occurred.
Notifation of this event when subscribed lets the external application synchronize with new fields or be able to take appropriate actions.

## Types of changes

New event eventFormPublished

https://bcdevex.atlassian.net/wiki/spaces/CCP/pages/1178304551/Event+Subscription+Endpoints

## Checklist

- [x] I have read the [CONTRIBUTING](CONTRIBUTING.md) doc
- [x] I have checked that unit tests pass locally with my changes
- [x] I have run the npm script lint on the frontend and backend
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)
- [x] I have approval from the product owner for the contribution in this pull request

PR #1180, opened 2023-12-06
Add in publish form event subscription (#1146)

bcgov-citz-ccft:
# Description
When a form is published the form id and form version id are sent to the subscribed endpoint.

image

Add in callouts to the existing endpoint subscription

Telling external applications when a form version is published informs the endpoint that changes occurred.
Notifation of this event when subscribed lets the external application synchronize with new fields or be able to take appropriate actions.

## Types of changes

## Checklist

## Further comments

https://bcdevex.atlassian.net/wiki/spaces/CCP/pages/1178304551/Event+Subscription+Endpoints


app/src/forms/form/service.js
```diff
@@ -369,6 +369,16 @@ const service = {
...
 
+      const { subscribe } = await service.readForm(formId);
+      if (subscribe && subscribe.enabled) {
+        const subscribeConfig = await service.readFormSubscriptionDetails(formId);
+        const config = Object.assign({}, subscribe, subscribeConfig);
+        const formVersion = new FormVersion();
+        formVersion.id = formVersionId;
+        formVersion.formId = formId;
+        service.postSubscriptionEvent(config, formVersion, null, SubscriptionEvent.FORM_PUBLISHED);
+      }
```
WalterMoar, line 380: I'm getting a "publish" event when a version is unpublished. The publish boolean created on line 353 could be used to restrict this to only the publishing of new versions. Can you wrap this whole chunk of new code in an `if (publish) { ... }`?

WalterMoar (changes requested):
Apparently there are two ways of publishing a form version: the service methods publishVersion (the one that has the new code) and publishDraft. This second method will also need to produce the event. It is called when you:
- create a new form
- add a control to it, save, and publish
- go to Form Management and at the bottom click the `+` to create a new version
- change the form and save - do not publish yet
- go to Form Management and click the toggle to publish the new version

app/src/forms/form/service.js
WalterMoar: This is all good with the code fixes.

WalterMoar (approved):

WalterMoar (approved):
I just want to leave some notes here:
- Looks great, and good to go!
- There's a chicken/egg issue where you can't configure the subscription endpoint until the form is published. But then you wouldn't get the publish event. So we should probably remove the restriction and allow it to be set up any time.
- If I have two versions and one is published: if I unpublish one and publish the other, I get unpublish/publish events. However, if one is published and I publish the other, I only get the publish event (even though the other version gets unpublished, there is no event). Not a biggee.

This is an experimental feature so it doesn't have to be completely polished - we can do that after more functionality is completed.

PR #1190, opened 2023-12-15
feat: FORMS-937 allow basicauth for template rendering

bcgov-citz-ccft:
# Description

In the Teams channel (2023-12-01 at 16:28) someone is asking if we can allow Basic Auth use of the endpoint `app/api/v1/submissions/{formSubmissionId}/template/render`. This would be so that people who are using the CHEFS API can also use it for their document generation, rather than having to also sign up for CDOGS.
- Update the endpoint with the rate limiting and apiKey middleware
- Update the API documentation Authorizations, plus add the rate limiting info to the 200 response, and the 429 response

## Types of changes

## Checklist

## Further comments

> Note: Can’t really add tests since we haven’t determined how we want to test the route middleware

### DIFF
--- a/app/src/docs/v1.api-spec.yaml
+++ b/app/src/docs/v1.api-spec.yaml
@@ -1396,11 +1396,26 @@ paths:
                 $ref: '#/components/schemas/Error'
   /submissions/{formSubmissionId}/template/render:
     post:
-      summary: Generate document from inline Template
+      summary: Generate a document from a document template
       description: >-
-        This endpoint accepts a document template and a set (or multiple sets)
-        of substitution variables and merges them into the document.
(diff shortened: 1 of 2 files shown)


app/src/docs/v1.api-spec.yaml
```diff
@@ -3081,7 +3102,7 @@ components:
             showSubmissionConfirmation:
               type: boolean
               example: true
-            sendSubmissionReceivedEmail: 
+            sendSubmissionReceivedEmail:
```
WalterMoar, line 3105: (I have the VSCode Prettier extension to format on save)

Sai-CITZ (approved):

vijaivir (approved):

PR #1193, opened 2023-12-18
feat: FORMS-926 added utils.getBaseUrl for notification urls

bcgov-citz-ccft:
# Description

If a form has “After Submission” team notification emails configured, they will fail if a submission is sent in via an API call. The `emailService._appUrl` method takes a `referer` parameter that is used to construct the protocol, domain name, and port portions of the URL to the resulting submission. When the request is an API call without the `referer`, it’s failing with an `Invalid URL` error. 

Since the API call will not be including a `referer`, another way must be found to construct the URLs. The fix should work for the various types of URLs that we have:
- http://localhost:5173/app/…
- https://chefs-dev.apps.silver.devops.gov.bc.ca/app/…
- https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-1111/…
- https://submit.digital.gov.bc.c.a/app/…

## Types of changes

## Checklist

## Further comments

Noticed that `reminderService._getReferer` has the fix needed for `emailService._appUrl`. Replaced both with a new utility function `getBaseUrl`.

### DIFF
--- a/app/src/forms/common/utils.js
+++ b/app/src/forms/common/utils.js
@@ -1,3 +1,4 @@
+const config = require('config');
 const falsey = require('falsey');
 const moment = require('moment');
 const clone = require('lodash/clone');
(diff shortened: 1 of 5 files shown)


Sai-CITZ (approved):
Great solution! Tested for email notif using API call & normal submission - everything looks great.

PR #1196, opened 2023-12-19
fix: FORMS-940 Validate UUIDs in the controller

bcgov-citz-ccft:
# Description

The API is currently set up so that bad data from API Key callers is typically caught in the database queries. This add a small amount of load to the database, but more importantly it causes errors in the database logs. We cannot alert on errors in the database logs until the errors due to API Key users are prevented.

In other words, errors in the database logs should be for _database_ errors, not _user_ errors. We can’t control the _user_ errors and they add noise to the database logs. Since we must investigate and correct the _database_ errors, we want the logs as clean as possible.

## Types of changes

## Checklist

## Further comments

> NOTE: this is sort of a proof-of-concept for the two endpoints where we commonly see invalid UUIDs. If it works well then it should eventually be done for all the UUIDs in all the endpoints :|

### DIFF
--- a/app/src/docs/v1.api-spec.yaml
+++ b/app/src/docs/v1.api-spec.yaml
@@ -493,6 +493,8 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/FormOptions'
+        '400':
+          $ref: '#/components/responses/BadRequest'
         default:
           description: Unexpected error
           content:
@@ -926,6 +928,8 @@ paths:
               $ref: '#/components/headers/RateLimit'
             RateLimit-Policy:
               $ref: '#/components/headers/RateLimit-Policy'
+        '400':
+          $ref: '#/components/responses/BadRequest'
         '401':
           $ref: '#/components/responses/UnauthorizedError'
         '403':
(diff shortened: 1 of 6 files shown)


app/src/docs/v1.api-spec.yaml
```diff
@@ -2608,7 +2612,7 @@ paths:
           content:
             application/json:
               schema:
-                  $ref: '#/components/schemas/Label'
+                $ref: '#/components/schemas/Label'
```
bcgov-citz-ccft, line 2615: This is just the VSCode Prettier extension doing its "Format on Save" thing.

WalterMoar, line 2615: This is just the VSCode Prettier extension doing its "Format on Save" thing.

app/tests/unit/routes/v1/form.spec.js
```diff
@@ -900,7 +901,7 @@ describe(`${basePath}/:formId/export/fields`, () => {
 });
 
 describe(`${basePath}/:formId/options`, () => {
-  const path = `${basePath}/:formId/options`;
+  const path = `${basePath}/${uuidv4()}/options`;
```
WalterMoar, line 904: This is a bit of a warning flag that the _route_ tests are actually testing the _controller_. That's a whole other topic of testing that needs to be dealt with!

Sai-CITZ (approved):

PR #1202, opened 2023-12-30
fix: FORMS-948 use dataErrors only one time

bcgov-citz-ccft:
# Description

The function `common/utils/setupMount` is doing `app.use(dataErrors)` but this is a global `use` that should only ever be done once. Because `setupMount` is called from multiple `index.js` files, when an error is thrown the error handling function is being called repeatedly. Using `dataErrors` once will simplify the code and the runtime will be faster and less complicated.

(Ran into a problem when trying to modify `dataErrors` in that the changes needed to handle being run multiple times - not ideal!)

## Types of changes

## Checklist

## Further comments

We need to set up some sort of integration tests (postman, etc) to regression test this sort of thing (unless someone can think of a way of unit testing it! I did add tests to cover `dataErrors`, but those don't test that it's integrated - and only once - with the routes).

### DIFF
--- a/app/src/forms/file/index.js
+++ b/app/src/forms/file/index.js
@@ -1,5 +1,4 @@
 const config = require('config');
-const dataErrors = require('../common/middleware').dataErrors;
 const routes = require('./routes');
 
 const _PATH = config.get('files.uploads.path') || 'files';
@@ -16,6 +15,5 @@ fileUpload.init({
 module.exports.mount = (app) => {
   const p = `/${_PATH}`;
   app.use(p, routes);
-  app.use(dataErrors);
   return p;
 };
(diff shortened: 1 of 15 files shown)


app/src/forms/file/index.js
```diff
@@ -14,8 +14,5 @@ fileUpload.init({
 });
 
 module.exports.mount = (app) => {
-  const p = `/${_PATH}`;
-  app.use(p, routes);
-  app.use(dataErrors);
-  return p;
+  return setupMount(_PATH, app, routes);
```
WalterMoar, line 17: Changing the file/index to use the same `setupMount` utility function that all the other index files use.

vijaivir (approved):

PR #1206, opened 2024-01-03
feat: FORMS-916 add Confirmation ID and Submission ID to data available in CDOGS

bcgov-citz-ccft:
# Description

A request has come in from the Teams channel (2023-11-16 11:24 a.m. - Jeremy K.) to be able to include the submission’s Confirmation ID in a CDOGS template. As this field is currently not available in the data sent to CDOGS, it cannot be used in the template.

A couple of catches!
- If we just add `confirmationId` and have it accessible in the CDOGS template as `d.confirmationId` there is the chance that someone could have a field on their form that is also called `confirmationId`. We have seen this problem in the exports. Maybe it would be less likely to have a problem if we were to use it on the form as `d.chefs.confirmationId`?
- If we make the Confirmation ID available, we should also make the Submission ID available. Follow whatever pattern is used for the Confirmation ID, such as `d.chefs.submissionId`. 

## Types of changes

## Checklist

### DIFF
--- a/app/tests/unit/forms/submission/controller.spec.js
+++ b/app/tests/unit/forms/submission/controller.spec.js
@@ -1,3 +1,5 @@
+const { getMockRes } = require('@jest-mock/express');
+
 const { Statuses } = require('../../../../src/forms/common/constants');
 const controller = require('../../../../src/forms/submission/controller');
 const emailService = require('../../../../src/forms/email/emailService');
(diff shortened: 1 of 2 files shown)


app/tests/unit/forms/submission/controller.spec.js
```diff
@@ -99,26 +111,48 @@ describe('templateUploadAndRender', () => {
...
+      id: submissionId,
       submission: {
-        submission: {
-          data: {
-            simpletextfield: 'fistName lastName',
-            submit: true,
-          },
+        data: {
+          simpletextfield: 'firstName lastName',
+          submit: true,
```
bcgov-citz-ccft, line 119: This was previously nested too deeply and not appearing in what was sent to the CDOGS service! The tests now check that the data returned from the "database" is properly sent to "cdogs".

WalterMoar, line 119: This was previously nested too deeply and not appearing in what was sent to the CDOGS service! The tests now check that the data returned from the "database" is properly sent to "cdogs".

app/tests/unit/forms/submission/controller.spec.js
```diff
@@ -99,26 +111,48 @@ describe('templateUploadAndRender', () => {
...
+    expect(cdogsService.templateUploadAndRender).toHaveBeenCalledTimes(1);
+    expect(cdogsService.templateUploadAndRender).toHaveBeenCalledWith(templateReq.body);
+    expect(res.set).toHaveBeenCalledWith(
+      expect.objectContaining({
+        'Content-Disposition': 'attachment',
+      })
+    );
+  });
+
+  it('should use the cdogs content disposition when it exists', async () => {
```
WalterMoar, line 141: Added a second test to cover the branch where the content disposition is set in the response.

WalterMoar:
image

vijaivir:
Would we need to update our documentation to inform users to use the `chefs` keyword for confirmation and submission ids?

vijaivir (approved):

WalterMoar:
> Would we need to update our documentation to inform users to use the `chefs` keyword for confirmation and submission ids?

Yes indeed! Unfortunately, though, the Wiki is in limbo while we move to Backstage. That's something to be sorted out next sprint, hopefully.

PR #1222, opened 2024-01-09
fix: FORMS-954 express proxy configuration

WalterMoar:
# Description

The new custom Rate Limiter middleware for Express is causing the following error message on the first basic auth call to the API:

> ValidationError: The 'X-Forwarded-For' header is set but the Express 'trust proxy' setting is false (default). This could indicate a misconfiguration which would prevent express-rate-limit from accurately identifying users. See https://express-rate-limit.github.io/ERR_ERL_UNEXPECTED_X_FORWARDED_FOR/ for more information.

This could indicate that we are rate limiting based on a proxy IP address rather than the original IP address of the requester. Need to double-check that this is not the case.

Also look at the IP address being logged - these do not seem to be IP addresses of the requesters, but rather something in between. The problem is probably that we’re logging the proxy IP addresses rather than the requester’s IP address.

## Types of changes

## Checklist

## Further comments

For localhost work Express seems to ignore this setting when there are no x-forwarded-for headers. So there was no need to make a special case to not set this for localhost.

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -27,6 +27,12 @@ app.use(compression());
 app.use(express.json({ limit: config.get('server.bodyLimit') }));
 app.use(express.urlencoded({ extended: true }));
 
+// Express needs to know about the OpenShift proxy. With this setting Express
+// pulls the IP address from the headers, rather than use the proxy IP address.
+// This gives the correct IP address in the logs and for the rate limiting.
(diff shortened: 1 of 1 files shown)


WalterMoar:
Much better :D
image

jasonchung1871 (approved):

PR #1225, opened 2024-01-10
feat: Frontend component for GET /files API

vijaivir:
# Description

This feature adds a checkbox to the API management section of a form allowing users to specify whether they want to allow any submitted files to be accessible via the API key. As the GET /files API now allows access to submitted files, this adds an extra level of security so that files cannot be accessed unless specified by the form owner.

Screenshot 2024-01-10 at 2 53 54 PM

A new table is added to the API Key schema called `filesAPIAccess` which stores a boolean value representing whether access to files is allowed via the API. To allow access to files, the checkbox must be selected before the API key is generated. Similarly, access can be revoked by deselecting the checkbox and generating a new key.

## Types of changes

## Checklist

## Further comments

Some consideration should be made for the JEDI folks that might already be using the /files API, since they will now have forms without the updated `filesAPIAccess` attribute. In this case we will need to manually add a `true` value to their api key or ask them to generate a new api key.

### DIFF
--- a/app/src/db/migrations/20210721210234_019-api-key-models.js
+++ b/app/src/db/migrations/20210721210234_019-api-key-models.js
@@ -89,6 +89,7 @@ exports.up = function (knex) {
       );
       table.uuid('formId').references('id').inTable('form').notNullable().index();
       table.uuid('secret').unique().notNullable();
+      table.boolean('filesAPIAccess').defaultTo(false);
       stamps(knex, table);
     }))
 
--- a/app/frontend/src/components/forms/manage/ApiKey.vue
+++ b/app/frontend/src/components/forms/manage/ApiKey.vue
@@ -17,6 +17,7 @@ export default {
       showConfirmationDialog: false,
       showDeleteDialog: false,
       showSecret: false,
+      filesAPIAccess: false,
     };
   },
   computed: {
@@ -58,7 +59,7 @@ export default {
     ]),
     async createKey() {
       this.loading = true;
-      await this.generateApiKey(this.form.id);
+      await this.generateApiKey(this.form.id, this.filesAPIAccess);
       this.showSecret = false;
       this.loading = false;
       this.showConfirmationDialog = false;
@@ -74,6 +75,7 @@ export default {
     async readKey() {
       this.loading = true;
       await this.readApiKey(this.form.id);
+      this.filesAPIAccess = this.apiKey?.filesAPIAccess;
       this.loading = false;
     },
     showHideKey() {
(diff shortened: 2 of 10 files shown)


app/src/db/migrations/20210721210234_019-api-key-models.js
```diff
@@ -89,6 +89,7 @@ exports.up = function (knex) {
       );
       table.uuid('formId').references('id').inTable('form').notNullable().index();
       table.uuid('secret').unique().notNullable();
+      table.boolean('filesAPIAccess').defaultTo(false);
```
WalterMoar, line 92: For the database migrations, instead of changing the old migrations we create a new migration with `npm run migrate:make [your-migration-name]`. I don't think this change to an old migration would get picked up when deploying in dev/test/prod.

In the new migration we would do an add column for the "up" and then remove the column for the "down".

WalterMoar:
I'm worried that some users are going to think that clicking the checkbox will change the `filesAPIAccess` setting. For example, if I create an API key and then enable the files API access, it doesn't get saved.

Is there a way that we can make the `filesAPIAccess` independent of creating the API key? Perhaps a new endpoint? (The use case I'm thinking of is that if I have an existing system using an API key, I don't want to have to change the API key to enable calling the file API)

Thoughts?

vijaivir:
> I'm worried that some users are going to think that clicking the checkbox will change the `filesAPIAccess` setting. For example, if I create an API key and then enable the files API access, it doesn't get saved.
> 

Good point, I could create a message that pops when the user clicks on the checkbox. It could inform them that they would need to create a new key to use the feature. Alternatively, a tooltip could work as well that informs users about what the checkbox is and how to use it. 

> Is there a way that we can make the `filesAPIAccess` independent of creating the API key? Perhaps a new endpoint? (The use case I'm thinking of is that if I have an existing system using an API key, I don't want to have to change the API key to enable calling the file API)
> 
> Thoughts?

Yes, this should be possible and sounds like a better solution, I could have the checkbox work independently of the api key. The checkbox would simply allow/deny access to files if an api key has been generated.

app/src/db/migrations/20210721210234_019-api-key-models.js
vijaivir: Ok let me try it out.

vijaivir: Hi Walter, I was having some issues with resolving the merge conflicts so I decided to create a new PR to keep it cleaner: https://github.com/bcgov/common-hosted-form-service/pull/1253

PR #1228, opened 2024-01-11
fix: FORMS-946 validate authorization uuids

WalterMoar:
# Description

We get a lot of API Key calls with invalid UUIDs in the path. If these are not valid UUIDs, return 400 and not 401.

The apiAccess code has also increased in code complexity and has duplication in getting the form ID from the submission ID. Refactor to clean this up.

## Types of changes

## Checklist

## Further comments

Notes:
- Refactored the part where we figure out the Form ID given the various parameters (form id, submission id, or file id)
- Failed early when there are errors in the parameters, rather than do it in checkCredentials
- The tests are a substantial rewrite. Made them much stricter in checking the HTTP response code
- The tests have to check either the status of the response or the next value - this isn't ideal, but the basic auth library does it one way, and we do it a "better" way

### DIFF
--- a/app/src/forms/auth/middleware/apiAccess.js
+++ b/app/src/forms/auth/middleware/apiAccess.js
@@ -6,6 +6,74 @@ const formService = require('../../form/service');
 const submissionService = require('../../submission/service');
 const fileService = require('../../file/service');
 
+const HTTP_401_DETAIL = 'Invalid authorization credentials.';
+
+/**
+ * Gets the Form ID from the request parameters. Handles the cases where instead
+ * of an explicit Form ID, the parameters include something like a Submission ID
+ * that then has to be used to find the Form ID.
+ *
+ * @param {*} params the request parameters.
+ * @returns a UUID for form that the requested data belongs to.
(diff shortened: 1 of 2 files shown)


app/src/forms/auth/middleware/apiAccess.js
```diff
@@ -6,40 +6,99 @@ const formService = require('../../form/service');
...
-        const result = await formService.readApiKey(formId);
-        secret = result && result.secret ? result.secret : '';
+      const apiKey = await formService.readApiKey(formId);
+      if (!apiKey || !apiKey.secret) {
+        throw new Problem(401, { detail: HTTP_401_DETAIL });
       }
 
+      // if (params.id && apiKey.filesApiAccess === false) {
+      //   throw new Problem(401, { detail: HTTP_401_DETAIL });
+      // }
```
WalterMoar, line 98: Code for Jai to duplicate what he has in PR1225 - although this should be a 403 Problem like in PR1225 (oops)!

jasonchung1871 (approved):

PR #1246, opened 2024-01-16
fix: FORMS-959 optimize the /submissions routes

WalterMoar:
# Description

The API is set up to load the active forms and deleted forms for all API calls. For example, to get a form the API will get the permissions for all active forms and all deleted forms, and then it will pull the one form out of active or deleted to check permissions. We cannot afford these expensive database queries on every API call.

Change the API so that it only fetches the one form that is required to check permissions. Only do the active if the deleted fail.

Since this is a big scary change, only do one set of endpoints at a time. There will need to be a lot of manual testing to check this out.

Next up is the /submissions endpoints.

## Types of changes

## Checklist


jasonchung1871 (approved):

usingtechnology (commented):
i like what you are doing here, defer making any huge reads until needed.

PR #1251, opened 2024-01-18
fix: FORMS-959 optimize all routes and do cleanup

WalterMoar:
# Description

The API is set up to load the active forms and deleted forms for all API calls. For example, to get a form the API will get the permissions for all active forms and all deleted forms, and then it will pull the one form out of active or deleted to check permissions. We cannot afford these expensive database queries on every API call.

Change the API so that it only fetches the one form that is required to check permissions. Only do the active if the deleted fail.

Since this is a big scary change, only do one set of endpoints at a time. There will need to be a lot of manual testing to check this out.

This is the final chunk of work: get rid of the "temp" function and then do cleanup. This change will then cover all routes.

## Types of changes

## Checklist

### DIFF
--- a/app/src/forms/auth/middleware/userAccess.js
+++ b/app/src/forms/auth/middleware/userAccess.js
@@ -18,11 +18,7 @@ const getToken = (req) => {
 const setUser = async (req, _res, next) => {
   try {
     const token = getToken(req);
-    // we can limit the form list from query string or url params.  Url params override query params
-    // ex. /forms/:formId=ABC/version?formId=123
-    // the ABC in the url will be used... so don't do that.
-    const params = { ...req.query, ...req.params };
-    req.currentUser = await service.login(token, params, req.chefsLoadForms);
+    req.currentUser = await service.login(token);
     next();
   } catch (error) {
     next(error);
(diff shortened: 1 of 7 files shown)


app/src/forms/auth/middleware/userAccess.js
```diff
@@ -43,54 +39,12 @@ const currentUser = async (req, res, next) => {
...
 const _getForm = async (currentUser, formId) => {
-  let forms;
-  if (currentUser.forms) {
-    forms = currentUser.forms;
-  } else {
-    forms = await service.getUserForms(currentUser, { active: true, formId: formId });
-  }
-
+  const forms = await service.getUserForms(currentUser, { active: true, formId: formId });
   let form = forms.find((f) => f.formId === formId);
```
WalterMoar, line 44: There's a few places now where we get the forms from the service and then filter out the one we want. This second step is _probably_ not needed, but we'd need additional checks around the response from the service. I'm going to leave it for now so that it's an easier review, but it's pretty ugly code so I'll clean it up once I put some tests in place.

app/src/forms/auth/service.js
```diff
@@ -201,19 +201,11 @@ const service = {
...
-    if (!loadForms) {
-      return { ...user };
-    }
-
-    const forms = await service.getUserForms(user, params); // get forms for user (filtered by params)...
-    params.active = false;
-    const deletedForms = await service.getUserForms(user, params); // get forms for user (filtered by params)...
-
-    return { ...user, forms: forms, deletedForms: deletedForms };
+    return { ...user };
```
WalterMoar, line 208: I wasn't sure about this duplication - we were duplicating the `user` object before to include it in the result. I don't think we make any downstream changes to the object that would cause problems with the original? We probably don't need the copy now? I think this could just be:

```suggestion
    return user;
```

app/src/forms/form/routes.js
```diff
@@ -1,14 +1,14 @@
...
-const { currentUserTemp, hasFormPermissions } = require('../auth/middleware/userAccess');
+const { currentUser, hasFormPermissions } = require('../auth/middleware/userAccess');
 const P = require('../common/constants').Permissions;
 const rateLimiter = require('../common/middleware').apiKeyRateLimiter;
 
 const keycloak = require('../../components/keycloak');
 const controller = require('./controller');
 
-routes.use(currentUserTemp);
+routes.use(currentUser);
```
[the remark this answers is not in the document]
WalterMoar: Rate limiting is on our todo list, and hopefully that will be coming sometime soon.

PR #1253, opened 2024-01-18
feat: Frontend component for GET /files API

vijaivir:
# Description

This PR introduces a new API endpoint `\filesApiAccess` that allows user submitted files to be accessible via the api key. 
The frontend adds a checkbox to the API management section of a form allowing users to specify whether or not they want to allow this behaviour. As the GET /files API now allows access to submitted files, this adds an extra level of security so that files cannot be accessed unless specified by the form owner.

Screenshot 2024-01-18 at 9 55 00 AM

A new table is added to the API Key schema called `filesApiAccess` which stores a boolean value representing whether access to files is allowed via the API. To allow access to files, the checkbox must be selected after the API key is generated (it is otherwise disabled). Similarly, access can be revoked by deselecting the checkbox or deleting the key. 

## Types of changes

## Checklist


vijaivir:
Still need to add the API documentation.

app/src/forms/form/service.js
```diff
@@ -722,6 +724,30 @@ const service = {
     }
   },
 
+  // toggle filesApiAccess for the api key
```
WalterMoar, line 727: I'm wondering if instead of toggling the value, that we might want to pass in the desired true/false value? What I'm wondering is that if I want to set it to false, I have to first get the value to make sure that it is true, and then call this - but if I could just pass in "false" as the desired value, then it will be false even if it was false before?

What do you think? I don't think it's a big deal since all the API Key stuff is internal to CHEFS and will probably never be made available to basic auth users. Just a thought!

app/src/db/migrations/20240115201832_files-api-access.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = function(knex) {
+  return Promise.resolve()
+    .then(() => knex.schema.alterTable('form_api_key', table => {
+      table.boolean('filesApiAccess').defaultTo(false).comment('Keeps track of whether files can be accessed using the API key');
+    }));
+};
```
WalterMoar, line 10: We do have at least one form that is calling the files route with the API Key. How do we deploy this change without breaking things for them? After the column is added, should we maybe set it to true for everyone, and they can disable if they want? Tricky!

This is the Kibana query I used to find the callers, where the existence of "formId" means that it's an API Key user, and that it passed authn/z.
```
kubernetes.namespace_labels.environment:prod AND kubernetes.container_name.raw:app AND message:"app/api/v1/files" AND message:"formId"
```

app/src/forms/form/service.js
vijaivir: I'm a little confused. Were you thinking in terms of limiting the API calls? If so, you are correct, I can reduce the call in this function by passing in the variable from the frontend call. But since I have set this up as a toggle, there will never be a case where "false" will be passed if the value is already false. 

Also, isn't the API key already being used for basic auth users?

app/src/db/migrations/20240115201832_files-api-access.js
vijaivir: I think setting it true for everyone would defeat the purpose of this feature, as it was added to create an additional check to limit access to files. Could we set it to true just for that one form? 

I haven't looked into the database yet but I believe it was only James who is using this feature so far, so I could let him know about this change to ensure that he turns the feature on manually in the required forms.

app/src/forms/form/service.js
WalterMoar: Sorry for the confusion - what I'm thinking is that we want the API call to set a value (true or false), not toggle the value (change whatever it might be). For example, if I don't know what the value is now but I want to set it to false, I'd have to do (pseudo code):
```
state = GET filesApiAccess
if (state == true) {
 PUT filesApiAccess
}
```
where what I want to do is just:
```
PUT filesApiAccess false
```
I think it will be a lot simpler down the road if we set an explicit true/false value instead of toggling the existing value.

app/src/db/migrations/20240115201832_files-api-access.js
WalterMoar: I'm really worried about breaking existing functionality for our users. Even if someone has never made an API call with their key, we don't know if maybe they're creating a system that relies on the functionality. I think the safest thing would be to set all existing API Keys to allow the file access, and then ask users to turn it off if they don't need it. For all future API Keys that are created, agreed, the file access should be disabled by default.

The other approach is that we could contact every form owner who has an API Key and ask them if they use the file download or not. Perhaps that would be better? Maybe it's a Duncan/Bhuvan question?

app/src/forms/form/service.js
vijaivir: Hey Walter, I have made the changes you requested. I am now sending the value of true/false from the frontend and explicitly setting it. Let me know if this works.

app/src/db/migrations/20240115201832_files-api-access.js
vijaivir: It's tough to say. I thought James was the only one who even knew about this feature but you are right, we don't know who's read the updated docs and what kind of systems they're creating. I like your plan, so we can merge this and as soon as it's deployed we manually set the filesApiAccess value to true for everyone.

app/src/forms/form/service.js
WalterMoar: Looks great!

WalterMoar:
The backend looks good! There might (?) be an issue if someone passes in a garbage value instead of true/false but since this is only used by the CHEFS app I don't think we need to worry about that now - we'll get there when the test coverage gets more complete.

I'll wait for @jasonchung1871 to review the frontend, but looks good to me.

vijaivir:
> The backend looks good! There might (?) be an issue if someone passes in a garbage value instead of true/false but since this is only used by the CHEFS app I don't think we need to worry about that now - we'll get there when the test coverage gets more complete.
> 
> I'll wait for @jasonchung1871 to review the frontend, but looks good to me.

Interesting point I hadn't thought of that. I can add a simple check for this but should we even document this endpoint if it's meant to be used internally by CHEFS?

WalterMoar:
> > The backend looks good! There might (?) be an issue if someone passes in a garbage value instead of true/false but since this is only used by the CHEFS app I don't think we need to worry about that now - we'll get there when the test coverage gets more complete.
> > I'll wait for @jasonchung1871 to review the frontend, but looks good to me.
> 
> Interesting point I hadn't thought of that. I can add a simple check for this but should we even document this endpoint if it's meant to be used internally by CHEFS?

That's a really good question! I think we want to completely redo the API docs at some point. But I think the approach so far is to document every endpoint - it's helpful for the CHEFS devs too! Eventually we might want to split them up into internal (Bearer-only) and external (Basic + Bearer) sets of documentation. Although a lot of the internal endpoints are useful to the basicauth folks and we end up allowing them access to them (although I don't think that this endpoint is something we'd ever allow).

vijaivir:
> > > The backend looks good! There might (?) be an issue if someone passes in a garbage value instead of true/false but since this is only used by the CHEFS app I don't think we need to worry about that now - we'll get there when the test coverage gets more complete.
> > > I'll wait for @jasonchung1871 to review the frontend, but looks good to me.
> > 
> > 
> > Interesting point I hadn't thought of that. I can add a simple check for this but should we even document this endpoint if it's meant to be used internally by CHEFS?
> 
> That's a really good question! I think we want to completely redo the API docs at some point. But I think the approach so far is to document every endpoint - it's helpful for the CHEFS devs too! Eventually we might want to split them up into internal (Bearer-only) and external (Basic + Bearer) sets of documentation. Although a lot of the internal endpoints are useful to the basicauth folks and we end up allowing them access to them (although I don't think that this endpoint is something we'd ever allow).

That makes sense, I have added a boolean validation for now!

app/src/forms/form/service.js
```diff
@@ -747,6 +749,33 @@ const service = {
     }
   },
 
+  // Set the filesApiAccess boolean for the api key
+  filesApiKeyAccess: async (formId, filesApiAccess) => {
+    let trx;
+    try {
+      if (typeof filesApiAccess !== 'boolean') {
+        throw new Problem(422, `filesApiAccess must be a boolean`);
```
WalterMoar, line 757: Awesome that you're adding the check to make sure that the value is valid. And love that you're throwing a Problem, some of the other code is doing weird and wild things but throwing something is super clean.

One request, though: can you change this to a 400 instead of a 422? The API is using 422 in a bunch of places that it shouldn't be used, and we've slowly been replacing them with 400. It would be great to get this done before it gets deployed, so we don't have to change it later.

vijaivir: Sounds good. It's done!

WalterMoar: Looks great!

WalterMoar:
The backend looks great!! I think the frontend is good too, but I'd love to get Jason C's approval on that.

jasonchung1871 (approved):

PR #1258, opened 2024-01-25
Keycloak Custom to Standard, Identity Provider refactoring

usingtechnology:
# Description

This PR has evolved into a long-running branch.
It incorporates changes to remove hardcoded Identity Provider (IDP) and move from a custom Keycloak realm to the BC Gov Standard realm.

Please note that not all hardcoded BC Gov logic is removed. There are still some IDIM things, backing out all of BC Gov quirks/nuances will be a separate project. This PR removes logic that explicitly used hardcoded values for our current IDPs.

Identity Provider logic is/was deeply embedded in CHEFS code. Which IDP a user signs in with determines levels of access to UX screens and functionality. The IDPs were hardcoded in the app. This does not make it easy to add new IDP, remove IDPs or change the access permissions for an IDP.

The move from a custom Keycloak realm (that the CHEFs team managed) to the standard BC Gov realm reduces support time for CHEFs developers and puts support onto the SSO team (experts in the field). More importantly, by using tokens from the standard realm, we open up CHEFs to use single sign-on for other users/apps within that realm. *NOTE*: we are currently validating that the token is for our specific audience (our OIDC client) and that logic will have to be loosened to accept tokens from the same issuer but issued through other clients.

See [FORMS-960](https://bcdevex.atlassian.net/browse/FORMS-960?atlOrigin=eyJpIjoiMmJjZGM2NDhhY2M0NDY3Yzg1ZGYwYzM1ZDNkZmVmNWQiLCJwIjoiaiJ9) and [FORMS-969](https://bcdevex.atlassian.net/browse/FORMS-969?atlOrigin=eyJpIjoiY2ZkMWY3ODg3ZjBmNDU0NTkwMDczNGIwOWQ4ODA1NWEiLCJwIjoiaiJ9)

### Identity Provider Updates
This PR is to move the IDP configuration from hardcoded in the UX to configuration from the API server (stored in the db). Add columns to the `identity_provider` table and add a datastore with helper methods to the UX. Access within the UX is not by IDP but by *permissions* given to that IDP.

Columns added to `identity_provider`:
- `primary`: boolean, which IDP is the highest level access (currently IDIR)
- `login`: boolean, if this IDP should appear as a login option (Public does not)
- `permissions`: string array, what permissions within CHEFS (not forms) does this IDP have
- `roles`: string array, what Form Roles does this IDP have (designer, owner, submitter, etc)
- `tokenmap`: json blob. this contains the mapping of IDP token fields to userInfo fields.
- `extra`: json blob. this is where non-standard configuration goes. we don't want a column for everything.

`extra` for now is used for the BCeID logins. Currently, these two were grouped together for displaying IDIM messages, and also have specific errors for searching / adding members. This kind of logic doesn't apply to IDIR.

`tokenmap` replaces the logic that we had in our custom Keycloak realm. Moving to Standard realm means we cannot have our own mappers in Keycloak.  Each IDP fills a different set of field and we need those to populate our userInfo object. This json blob specifies { `userInfo field name`: `token field name` } pairs. Special logic is added to ensure our `userInfo.keycloakId`field is in the expected GUID format.

Very important: we have always had identityProvider fields for `code`, `display` and `idp`. In our custom realm, `code` and `idp` were identical. The logic in our code was using `code` and `idp` interchangeably and that is no longer possible. `code` is our referential integrity, `idp` is what is in the token. In the frontend `idp` is generally used as `idpHint`. Unfortunately, the `userInfo.idp` == `identityProvider.code`... I have added `userInfo.idpHint` and there are lookup/helper routines to find the `code` or `idp` depending on what you know.  Also, previously the frontend was using `idp` as a string, it is now an object:

```
{ 
  `code`: <identityProvider.code>, 
  `display`: <identityProvider.display>, 
  `hint`: <identityProvider.idp> 
}
```

*NOTE*:  when this is all stable, I think we should rename the db field and object from `idp` to `hint` and likewise update the userInfo object to use `code` and `hint` from `idp` and `idpHint`.

Most of the logic on the backend is encapsulated in `idpService`. In the frontend, there is a new store `idpStore`.

#### Service Clients
The biggest change is on the server side. Previously we had 2 service clients: `chefs` and `chefs-frontend`. Users would sign in through the browser and use `chefs-frontend`. That service client would grab assignments (user, roles, groups) from the `chefs` service client. So `chefs` would be the token audience we cared about. With this move, the frontend has the only client - and a big change there is it is a `public` client, not a `confidential` one. So no secrets required.

Now the backend/server only needs to validate/verify that the token passed along the request is what we expect. So configuration is more complicated but again contains no secrets. Basically the backend is configured to know what server/issuer to ask if this token is valid. We check the issuer, the audience and the time. *Note* we may have to relax the audience to open CHEFs up to true SSO... future investigation/work.

TLDR: there is only one service client and it is for the frontend.

#### Token Attribute Mapping
With the move to standard realm, we loose attribute/field mapping within keycloak. This is now handled as described above via the `identityProvider.tokenmap` data. 

#### Roles
We also lose `resource_access`. This is now replaced with `client_roles` and is not qualified by the service client. It is simply an array of strings. And now we only have `admin`. The admin role can be added through the SSO Console to users who have logged in to the app. The helper functions in the authStore are still called `resourceAccess` and `hasResourceRoles`.

```
  // new
  "client_roles": [
    "admin"
  ],
  // old
  "resource_access": {
    "chefs": {
      "roles": [
        "user"
      ]
    }
  },
```

*User* role now removed. Anyone who signed in was a user - it is equivalent to anyone authenticated. Removed `isUser` and removed logic in the `BaseSecure` component.

#### Frontend code
The frontend code remains basically the same. We (BCGOV) need a specific logout url so we can logout of Siteminder and Keycloak; so that configuration is new.

#### Backend code
We have dropped the `keycloak-connect` node library (repeatedly threatened to be deprecated) and added `jose`. `jose` is used to do the JWT/Token work. We've added a `jwtService` that hides the bearer token logic and token validation, more importantly it houses the middleware `protect` function that we lose by removing `keycloak-connect`. This is the middleware that verifies the token and can inspect for the `admin` role.

### Others
**also** added some changes to the `.devcontainer` to run `knex` and `jest` commands. This helps out when you want to use the terminal to run migrations or rollbacks and when you want to run specific backend unit tests.

## Checklist


matthewhall78 (commented):
Does this change account for the fact that we made it possible to add "team management" and "submission reviewer" roles to the business bceid IDP users that are added to the team management list?

usingtechnology:
> Does this change account for the fact that we made it possible to add "team management" and "submission reviewer" roles to the business bceid IDP users that are added to the team management list?

Hopefully! I do need some experts to make sure this works properly, but per IDP there are the form roles. See [here](https://github.com/bcgov/common-hosted-form-service/blob/77db78babee4a440f77848f9602c0ea245b42218/app/src/db/migrations/20240119172630_identity_provider_permissions.js#L89) for where we seeded BCeID Business. Then we disable roles NOT in the IDP list [here](https://github.com/bcgov/common-hosted-form-service/blob/77db78babee4a440f77848f9602c0ea245b42218/app/frontend/src/components/forms/manage/TeamManagement.vue#L204).

WalterMoar:
Looks great so far! Looking forward to the PR deployment and being able to try it out.

.devcontainer/devcontainer.json
```diff
@@ -10,10 +10,35 @@
...
+        "vscode": {
+            "extensions": [
+                "cweijan.vscode-postgresql-client2",
+                "Vue.volar",
+                "esbenp.prettier-vscode"
+            ],
+            "settings": {
+                "database-client.telemetry.usesOnlineServices": false,
+                "editor.defaultFormatter": null,
+                "editor.formatOnSave": false,
```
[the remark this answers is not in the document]
usingtechnology: yup, I'd defer to your recommendation as I've only worked on a limited set of files and am being very conservative. are their configurations for rules on yaml and json - so we all save them with consistent format/style?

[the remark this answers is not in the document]
WalterMoar: I have it set up for Prettier to format everything with its default settings. I don't think we're too particularly about how things are formatted, just as long as they're consistent and we don't have a lot of PRs where formatting flips back and forth between different formatters (builtin vs Prettier, etc).

openshift/app.dc.yaml
```diff
@@ -252,6 +252,8 @@ objects:
                     name: "${APP_NAME}-service-config"
                 - configMapRef:
                     name: "${APP_NAME}-files-config"
+                - configMapRef:
+                    name: "${APP_NAME}-oidc-config"
```
WalterMoar, line 256: With this new ConfigMap can the `SERVER_KC_CLIENTID` and `SERVER_KC_CLIENTSECRET` environment variables in this DeploymentConfig be removed?

usingtechnology: yes, all the XXX_KC_YYY env vars can go away. And there will be no more KC secret or public key

app/src/components/idpService.js
```diff
@@ -0,0 +1,169 @@
+const errorToProblem = require('./errorToProblem');
```
WalterMoar, line 1: Can we get unit tests added for idpService and jwtService? I think that Code Climate is going to complain about code complexity, and having the tests will make future refactoring easier.

usingtechnology: seems like a great idea. will do.

app/src/forms/common/constants.js
```diff
@@ -93,4 +82,21 @@ module.exports = Object.freeze({
...
+  APP_PERMISSIONS: {
+    VIEWS_FORM_STEPPER: 'views_form_stepper',
+    VIEWS_ADMIN: 'views_admin',
+    VIEWS_FILE_DOWNLOAD: 'views_file_download',
+    VIEWS_FORM_EMAILS: 'views_form_emails',
+    VIEWS_FORM_EXPORT: 'views_form_export',
+    VIEWS_FORM_MANAGE: 'views_form_manage',
+    VIEWS_FORM_PREVIEW: 'views_form_preview',
+    VIEWS_FORM_SUBMISSIONS: 'views_form_submissions',
+    VIEWS_FORM_TEAMS: 'views_form_teamS',
```
WalterMoar, line 98: Little typo here?
```suggestion
    VIEWS_FORM_TEAMS: 'views_form_teams',
```

app/frontend/src/utils/constants.js
```diff
@@ -63,20 +63,28 @@ export const FormManagePermissions = Object.freeze([
...
+export const AppPermissions = Object.freeze({
+  VIEWS_FORM_STEPPER: 'views_form_stepper',
+  VIEWS_ADMIN: 'views_admin',
+  VIEWS_FILE_DOWNLOAD: 'views_file_download',
+  VIEWS_FORM_EMAILS: 'views_form_emails',
+  VIEWS_FORM_EXPORT: 'views_form_export',
+  VIEWS_FORM_MANAGE: 'views_form_manage',
+  VIEWS_FORM_PREVIEW: 'views_form_preview',
+  VIEWS_FORM_SUBMISSIONS: 'views_form_submissions',
+  VIEWS_FORM_TEAMS: 'views_form_teamS',
```
WalterMoar, line 76: Little typo here?
```suggestion
  VIEWS_FORM_TEAMS: 'views_form_teams',
```

app/src/forms/common/constants.js
```diff
@@ -93,4 +82,21 @@ module.exports = Object.freeze({
     json: 'json',
     default: 'csv',
   },
+  // app permissions are not assigned on the form
+  // they are for flow within the UX, what views one can navigate
+  // what buttons one can have.
+  // these will be assigned via the user's IDP.
+  APP_PERMISSIONS: {
```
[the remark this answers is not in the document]
usingtechnology: Really only needed in the backed for the db migrations, so we've got two sets of constants - backend and frontend for the real use case. I think we've got that for all kinds of things which is kind of annoying but other than building a library, not sure how to share the constants in the both "apps".

app/src/forms/auth/middleware/userAccess.js
```diff
@@ -83,39 +56,31 @@ const _getForm = async (currentUser, formId, includeDeleted) => {
...
-
+    const token = await jwtService.getTokenPayload(req);
+    req.currentUser = await service.login(token);
     next();
   } catch (error) {
     next(error);
   }
 };
 
+const currentUser = async (req, res, next) => {
```
WalterMoar, line 69: It looks like some of the changes in PR1263 were lost on the merge. The only important thing is to wrap the code in `currentUser` (and all middleware) in `try`/`catch` blocks so that requests won't hang on error. Something like:
```javascript
const middleware = async (req, _res, next) => {
  try {
    // If possible the helper functions should throw a Problem rather than return a
    // response that needs handling logic.
    const formId = _do_stuff(req);
    const foo = service.getForm(formId);

    next();
  } catch (error) {
    next(error);
  }
};
```

app/src/forms/form/routes.js
```diff
@@ -15,7 +14,7 @@ routes.param('formId', validateParameter.validateFormId);
 routes.param('formVersionDraftId', validateParameter.validateFormVersionDraftId);
 routes.param('formVersionId', validateParameter.validateFormVersionId);
 
-routes.get('/', keycloak.protect(`${config.get('server.keycloak.clientId')}:admin`), async (req, res, next) => {
+routes.get('/', jwtService.protect('admin'), async (req, res, next) => {
```
WalterMoar, line 17: Previously the name of the administrator role was configurable. Any concern that another jurisdiction would want the administrator role to be named something other than `admin`?

usingtechnology: the previous syntax was for hierarchy to check through `resouce_access` which was key: [ roles ], with the key being a KC service client. in this PR, roles are just an array and we only have one service client. if the token isn't for our client it is rejected BEFORE we look for roles (now in `client_roles`). of course, if we change our token verification to not care about the client, we will need to adjust our `protect` middleware to not only look for the role but the client/audience too. TLDR - not a concern at the moment.

app/src/forms/common/constants.js
usingtechnology: this is fixed now. I've updated the data on the open shift PR too.

app/src/forms/auth/middleware/userAccess.js
usingtechnology: great catch and fixed this up (and related test)

app/src/components/idpService.js
WalterMoar: Looks great, thank you!

app/src/forms/common/constants.js
WalterMoar: Great!

app/src/forms/auth/middleware/userAccess.js
WalterMoar: Looks good, the request is no longer hanging when the token expires.

app/src/components/jwtService.js
```diff
@@ -0,0 +1,105 @@
...
+    return payload;
+  }
+
+  async validateAccessToken(token) {
+    try {
+      await this._verify(token);
+      // these claims passed, just return true.
+      return true;
+    } catch (e) {
+      if (e instanceof jose.errors.JWTClaimValidationFailed) {
```
WalterMoar, line 57: The call to `errorToProblem` below is returning a `502` response when:
- The JWT header is invalid (`jose.errors.JWSInvalid`)
- The JWT signature is invalid (`jose.errors.JWSSignatureVerificationFailed`)
- The JWT is expired (`jose.errors.JWTExpired`)

There's a bunch of other errors that might happen. What's the the best approach here? Do we maybe want to just treat all JOSE errors as unauthenticated? Something like:

```suggestion
      if (e instanceof jose.errors.JOSEError) {
```
???

Or should we throw the `JOSEError`s and in the caller use the message in the response details? So many options!

app/src/forms/auth/middleware/userAccess.js
```diff
@@ -97,17 +70,17 @@ const _getForm = async (currentUser, formId, includeDeleted) => {
 const currentUser = async (req, _res, next) => {
   try {
     // Validate bearer tokens before anything else - failure means no access.
-    const bearerToken = _getBearerToken(req);
+    const bearerToken = jwtService.getBearerToken(req);
     if (bearerToken) {
-      const ok = await keycloak.grantManager.validateAccessToken(bearerToken);
+      const ok = await jwtService.validateAccessToken(bearerToken);
       if (!ok) {
         throw new Problem(403, { detail: 'Authorization token is invalid.' });
```
WalterMoar, line 77: This is outside the scope of this change, but this should be a 401, not a 403, yes? I'm wondering if this PR is a good time to fix that, since we have to contact all our clients to tell them about the SSO change - might be a good way to get this "breaking" change done?

app/src/components/jwtService.js
usingtechnology: well, if JOSEError is a parent class, then perhaps we should catch that. Returning `false` is to simulate the behaviour of the `keycloak.grantManager` which just failed. No differentiation on why token was invalid or bad. So I am definitely good with catching `JOSEError` instead.

app/src/forms/auth/middleware/userAccess.js
usingtechnology: Sounds good to me. Will make that change. How does this flow into documentation of the API?

app/src/components/jwtService.js
usingtechnology: I believe jwtVerify will only throw `JWTClaimValidationFailed` but we aren't really doing anything with claim field in the error anyway, so catching everything JOSE related should be fine.

app/src/forms/auth/middleware/userAccess.js
WalterMoar: Oh boy. Good question. Maybe make the change here so that we're testing with the 401, and I'll update the docs in a separate PR since it's going to be a lot of (small) changes.

app/src/components/jwtService.js
WalterMoar: Looks great! Getting 401s for all of the above.

app/src/forms/auth/middleware/userAccess.js
WalterMoar: The docs are updated, merged, and deployed. Confirmed that invalid or expired tokens are now returning 401s - hooray!

app/src/forms/form/service.js
```diff
@@ -237,7 +237,7 @@ const service = {
       .withGraphFetched('idpHints')
       .throwIfNotFound()
       .then((form) => {
-        form.idpHints = form.idpHints.map((idp) => idp.code);
+        form.idpHints = form.idpHints.map((idp) => idp.idp);
```
WalterMoar, line 240: If I'm reading things correctly, this `idpHints` graph is from the `form_identity_provider` table, So this line 240 change is causing null to be returned for the IdPs. Is it more consistent in the long run to revert the line 240 change, or to change the name of the `form_identity_provider` column?

usingtechnology: I put this in the description and it is also in the documents, but not sure if I put a backlog ticket in? If there isn't one there should be.

> NOTE: when this is all stable, I think we should rename the db field and object from idp to hint and likewise update the userInfo object to use code and hint from idp and idpHint.

I did not want to do all that work in this PR. This is one area where it was very clear that `idp` and `code` were being used interchangeably and only worked because the values were the same.

usingtechnology: sigh, ok, I see what you are talking about (was right in the middle of something that used this). not sure why this worked and now doesn't. `code` certainly isn't correct (if the idp != code)...  not sure what exactly happened but yes, this needs fixing! I'll make one and push it up soon.

usingtechnology: No idea how this worked, best guess is that it actually didn't work and I just hit a case where it no longer failed like before but worked with a `null` value.  Anyway, fixing it!

usingtechnology:
```
CountryMac:~ jason$ curl https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-1258/api/v1/forms/2c7c02bd-9040-4315-bae7-04163424e2d3/options; echo
{"id":"2c7c02bd-9040-4315-bae7-04163424e2d3","name":"Auth - Public","description":"","idpHints":["public"],"snake":"auth_-_public"}
```

usingtechnology: `"idpHints":["public"]` is returned.

jasonchung1871 (approved):
At this point.. it looks good, I couldn't find any issues with it in the deployment for how I normally test and use CHEFS.

PR #1264, opened 2024-01-30
prevents export with no submissions

jasonchung1871:
# Description
If a reviewer is trying to export submissions with no submissions, it gives an infinite spinner. This will return an error to the user indicating that there are no submissions if there are no submissions. A valid submission is a form that is not in the draft state or deleted state.

resolves: https://bcdevex.atlassian.net/browse/FORMS-857

## Checklist


app/package-lock.json
```diff
@@ -1542,9 +1542,9 @@
       }
     },
     "node_modules/@types/express-serve-static-core": {
-      "version": "4.17.41",
-      "resolved": "https://registry.npmjs.org/@types/express-serve-static-core/-/express-serve-static-core-4.17.41.tgz",
-      "integrity": "sha512-OaJ7XLaelTgrvlZD8/aa0vvvxZdUmlCn6MtWeB7TkiKW70BQLc9XEPpDLPdbo52ZhXUCrznlWdCHWxJWtdyajA==",
+      "version": "4.17.42",
+      "resolved": "https://registry.npmjs.org/@types/express-serve-static-core/-/express-serve-static-core-4.17.42.tgz",
+      "integrity": "sha512-ckM3jm2bf/MfB3+spLPWYPUH573plBFwpOhqQ2WottxYV85j1HQFlxmnTq57X1yHY9awZPig06hL/cLMgNWHIQ==",
```
WalterMoar, line 1547: Is this file needed for the change? It's OK if it gets included in the PR, as we need to update the dependencies, but just curious.

jasonchung1871: It is not required!

WalterMoar:
Just waiting for this branch to be brought up to date and deployed.

app/tests/unit/forms/form/controller.spec.js
```diff
@@ -50,6 +52,21 @@ describe('form controller', () => {
     await controller.readFieldsForCSVExport(req, {}, jest.fn());
     expect(exportService.fieldsForCSVExport).toHaveBeenCalledTimes(1);
   });
+
+  it('should not continue with export if there are no submissions', async () => {
```
WalterMoar, line 56: It's great to see new tests for the backend! It looks like the frontend changes don't have any test coverage, can those be added too? I assume we're still aiming for unit test coverage and the Cypress tests are focused on e2e?

app/frontend/tests/unit/views/file/DownloadTwo.spec.js
jasonchung1871, line 1: Did something weird here. I'm unable to figure out exactly why all subsequent calls to the method `getFile` get skipped after the first test, so I needed to create a second file containing the next test. This should be added to the backlog as a bug that needs to be looked into as this could be affecting other component tests.

app/frontend/src/components/forms/ExportSubmissions.vue
```diff
@@ -82,7 +82,9 @@ export default {
     },
     FILTER_HEADERS() {
       return this.versionSelected !== ''
-        ? this.formFields.map((f) => ({ name: f, value: f }))
+        ? this.formFields.length > 0
```
jasonchung1871, line 85: This allows the formFields to be empty as sometimes when generating the headers for a submission, it can be empty because there are no submissions. In theory, it should contain the form fields from the form design but this is a separate issue.

app/frontend/src/views/file/Download.vue
```diff
@@ -45,11 +45,7 @@ export default {
     async getFile(fileId) {
       this.showDownloadLink = false;
       await this.downloadFile(fileId);
-      if (
-        this.downloadedFile &&
-        this.downloadedFile.data &&
-        this.downloadedFile.headers
-      ) {
+      if (this.downloadedFile && this.downloadedFile.headers) {
```
jasonchung1871, line 48: This allows the download to begin if there is no data inside of the file to be downloaded, no submissions permitted. This way the user is actually notified that something happened instead of it just spinning infinitely.

PR #1272, opened 2024-02-07
fix: FORMS-994 check version id in route params

WalterMoar:
# Description

James P from JEDI reported a problem when calling API routes that have both `formId` and `formVersionId` parameters. The data for the `formVersionId` will be returned even if the form version does not belong to the given `formId`. Whenever these two parameters exist on a route we need to check that form contains the form version, and return a 404 response if not. We also need to check the routes that have both `formId` and `formVersionDraftId`.

While this is a potential security problem, the exploit would involve having the `formVersionId` for someone else’s form. The large key space of UUIDs protects us here, but the problem still needs to be fixed.

Changes:
- Prevent `formId` and `formVersionDraftId` pairs where the `formVersionDraftId` doesn't belong to the form. Return HTTP 404 on mismatch
- Prevent `formId` and `formVersionId` pairs where the `formVersionId` doesn't belong to the form. Return HTTP 404 on mismatch
- Validate `formId`, `formVersionDraftId`, and `formVersionId` parameters and return HTTP 400 if they are not valid UUIDs (was previously returning HTTP 401)

## Types of changes

## Checklist

## Further comments

This is a breaking change if:
- Developers are relying on the incorrect behaviour that allows `formId` mismatches
- Developers are using the HTTP 401 status code to detect invalid parameters and are not handling the 400 status code

### DIFF
--- a/app/src/docs/v1.api-spec.yaml
+++ b/app/src/docs/v1.api-spec.yaml
@@ -526,10 +526,14 @@ paths:
               $ref: '#/components/headers/RateLimit'
             RateLimit-Policy:
               $ref: '#/components/headers/RateLimit-Policy'
+        '400':
+          $ref: '#/components/responses/BadRequest'
         '401':
           $ref: '#/components/responses/UnauthorizedError'
         '403':
           $ref: '#/components/responses/Forbidden'
+        '404':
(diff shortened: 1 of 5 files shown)


WalterMoar:
Replaced with PR1274.

PR #1275, opened 2024-02-07
fix/FORMS-994 validate forms routes params

WalterMoar:
# Description

James P from JEDI reported a problem when calling API routes that have both `formId` and `formVersionId` parameters. The data for the `formVersionId` will be returned even if the form version does not belong to the given `formId`. Whenever these two parameters exist on a route we need to check that form contains the form version, and return a 404 response if not. We also need to check the routes that have both `formId` and `formVersionDraftId`.

While this is a potential security problem, the exploit would involve having the `formVersionId` for someone else’s form. The large key space of UUIDs protects us here, but the problem still needs to be fixed.

### Breaking Changes:

The following routes were incorrectly returning _HTTP 422 Unprocessable Content_ when the `formVersionId` was not a valid UUID. These routes now return _HTTP 400 Bad Request_ for invalid UUIDs.
- `GET /api/v1/forms/:formId/versions/:formVersionId`
- `GET /api/v1/forms/:formId/versions/:formVersionId/fields`
- `POST /api/v1/forms/:formId/versions/:formVersionId/multiSubmission`
- `GET /api/v1/forms/:formId/versions/:formVersionId/publish`
- `GET /api/v1/forms/:formId/versions/:formVersionId/submissions`
- `POST /api/v1/forms/:formId/versions/:formVersionId/submissions`
- `GET /api/v1/forms/:formId/versions/:formVersionId/submissions/discover`

The following routes were incorrectly returning _HTTP 422 Unprocessable Content_ when the `formVersionDraftId` was not a valid UUID. These routes now return _HTTP 400 Bad Request_ for invalid UUIDs.
- `DELETE /api/v1/forms/:formId/drafts/:formVersionDraftId`
- `GET /api/v1/forms/:formId/drafts/:formVersionDraftId`
- `PUT /api/v1/forms/:formId/drafts/:formVersionDraftId`
- `POST /api/v1/forms/:formId/drafts/:formVersionDraftId/publish`

## Types of changes

## Checklist

### DIFF
--- a/app/src/forms/auth/middleware/params.js
+++ b/app/src/forms/auth/middleware/params.js
@@ -1,0 +1,92 @@
+const Problem = require('api-problem');
+const { validate } = require('uuid');
+
+const formService = require('../../form/service');
+
+/**
+ * Validates that the :formId route parameter exists and is a UUID.
+ *
+ * @param {*} _req the Express object representing the HTTP request - unused
+ * @param {*} _res the Express object representing the HTTP response - unused
+ * @param {*} next the Express chaining function
+ * @param {*} formId the :formId value from the route
+ */
+const validateFormId = async (_req, _res, next, formId) => {
+  try {
+    if (!validate(formId)) {
+      throw new Problem(400, {
+        detail: 'Bad formId',
+      });
+    }
+
+    next();
+  } catch (error) {
+    next(error);
+  }
+};
+
+/**
+ * Validates that the :formVersionDraftId route parameter exists and is a UUID.
+ * This validator requires that the :formId route parameter also exists.
+ *
+ * @param {*} req the Express object representing the HTTP request
+ * @param {*} _res the Express object representing the HTTP response - unused
+ * @param {*} next the Express chaining function
+ * @param {*} formVersionDraftId the :formVersionDraftId value from the route
+ */
+const validateFormVersionDraftId = async (req, _res, next, formVersionDraftId) => {
+  try {
+    if (!validate(formVersionDraftId)) {
+      throw new Problem(400, {
+        detail: 'Bad formVersionDraftId',
+      });
+    }
+
+    const formVersionDraft = await formService.readDraft(formVersionDraftId);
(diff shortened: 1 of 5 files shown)


jimmyPasta:
Looks good thanks for the quick turn around, if you deploy this to a pr env I will verify the fix.
I will check also that you can not POST to versions for a version change or grab Submissions

jimmyPasta (approved):
Let me know if you deploy to a pr env first and I can do some basic testing

WalterMoar:
> Let me know if you deploy to a pr env first and I can do some basic testing

Thank you! It's deployed to https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-1275

app/src/forms/auth/middleware/params.js
```diff
@@ -0,0 +1,92 @@
...
+ */
+const validateFormVersionDraftId = async (req, _res, next, formVersionDraftId) => {
+  try {
+    if (!uuid.validate(formVersionDraftId)) {
+      throw new Problem(400, {
+        detail: 'Bad formVersionDraftId',
+      });
+    }
+
+    const formVersionDraft = await formService.readDraft(formVersionDraftId);
```
WalterMoar, line 45: This service call does add another database query to these routes, and the same in `validateFormVersionId` below.

We could think about doing some some of caching of these values once they're pulled from the database, but that would make things more complicated. Plus we'd have to deal with cache invalidation and things like that. Maybe keep it simple for now and do optimization later if we need it.

Ideally the database calls should be small and quick. The database itself should be caching these things and do a better job of it than we would!

jasonchung1871 (approved):

app/src/forms/auth/middleware/params.js
usingtechnology: Am I missing something here? (I haven't run this). For draft and form version don't you need the form id too to validate if they belong to each other?

usingtechnology: offline discussion cleared this up. early morning reading comprehension must be low.

usingtechnology (approved):

PR #1288, opened 2024-02-16
chore: FORMS-998 updates to the devcontainers

WalterMoar:
# Description

When finally making the move to the devcontainers there were a few things to add. This is all based on Jason Sherman’s original work as well as some in progress updates he was doing.
- Add `vi` so that people can do interactive rebases
- Add extensions - the ones Jason S. was adding plus a few more
- Configure Prettier to format everything on save. We previously had some people doing it, and some people using the internal formatter for some files. This led to PRs where the formatting flipped back and forth between users.

## Types of changes

Chore (changes unrelated to the deployed application)

## Checklist


.devcontainer/README.md
```diff
@@ -19,42 +21,46 @@ Also included are convenient launch tasks to run and debug CHEFS.
...
-4.  Navigate to the CSS page, login with your IDIR, and download the ‘Development’ Installation JSON from your SSO Integration. 
-5.  Back in the `realm-export.json` file, search for all instances of `YYYYYYYYYYYY` and replace it with the `resource` you obtained from the downloaded JSON file. Search for all instances of `ZZZZZZZZZZZZ` and replace it with the `secret`. 
+
+1.  An IDIR account is required to access CHEFS.
+2.  Request an SSO Integration from the Common Hosted Single Sign-on (CSS) page in order to obtain a resource and secret that will be used for authentication when building CHEFS. View the [detailed documentation](https://bcdevex.atlassian.net/wiki/spaces/CCP/pages/961675282) about requesting the Pathfinder SSO integration.
+3.  Open realm-export.json located at build/docker/imports/keycloak and search for `XXXXXXXXXXXX`. This value must match the `clientSecret` value in `local.json` so that the CHEFS API can connect to your Keycloak instance. By default, these are set to be equal and don’t need to be altered.
```
WalterMoar, line 43: The only edit to this file is changing `chefs_build` to `build` for the keycloak imports. Everything else is Prettier doing its thing.

usingtechnology (approved):
Thanks for cleaning up the docs and enhancing the container.

I couldn't get autosave/format to work, but I find that flaky anyway.
Bigger question/issue is we have 2 prettier configs - app and frontend/app. Perhaps we should make a global setting for lint and prettier?

WalterMoar:
> Thanks for cleaning up the docs and enhancing the container.
> 
> I couldn't get autosave/format to work, but I find that flaky anyway. Bigger question/issue is we have 2 prettier configs - app and frontend/app. Perhaps we should make a global setting for lint and prettier?

Format on save: When you save a file, is it not formatting? I wonder if it's my custom settings that are doing it and the settings in the `devcontainer.json` aren't taking effect.

We actually have four prettier configs, as there's one in the components and one in the load tests. Agreed, though, it's weird to have different formatting in different parts of the code. Definitely something to do, and I'd love see the API formatted to 80 characters and not 180, but that's going to be a big change to most files. Maybe I'll (eventually) do that as a first step, and once they're all aligned we can remove the duplicates.

WalterMoar:
> Yeah, nothing is triggering the save/format. Not sure exactly where it fails, don't see anything in the output logs/terminal. There are also like 3 levels of configuration for extensions - user level, the workspace level and the devcontainer so it is not always going to be consistent for every developer (unfortunately). Anyway, I think devs can tweak and make PRs, let others see if it affects them adversely before merge. it'll be on going. I figure as long as there are tasks to do the work manually that's ok as a catch-all fallback. So I say we just go ahead and get the devcontainer going and tweak as needed.

Found it, the `settings` needed to be within the `vscode` clause.

PR #1293, opened 2024-02-22
fix: FORMS-1017 get file for drafts status 403

WalterMoar:
# Description

When the file `routes.get('/:id', ...` is called for a file upload that does not have a submission, we are returning a 500 response. It turns out that this situation happens for all file uploads for draft submissions. The API Key users should not be trying to fetch draft submissions, but the 500 response is wrong - it should be a 403.

Since this is happening in production it should be a priority. The user agent is `RestSharp/110.2.0.0`, so perhaps it is a Microsoft Dynamics issue.

## Types of changes

## Checklist

### DIFF
--- a/app/src/forms/auth/middleware/apiAccess.js
+++ b/app/src/forms/auth/middleware/apiAccess.js
@@ -1,6 +1,6 @@
 const Problem = require('api-problem');
 const basicAuth = require('express-basic-auth');
-const { validate: uuidValidate } = require('uuid');
+const uuid = require('uuid');
 
 const formService = require('../../form/service');
 const submissionService = require('../../submission/service');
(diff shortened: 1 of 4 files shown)


app/src/forms/auth/middleware/apiAccess.js
```diff
@@ -1,6 +1,6 @@
 const Problem = require('api-problem');
 const basicAuth = require('express-basic-auth');
-const { validate: uuidValidate } = require('uuid');
+const uuid = require('uuid');
```
WalterMoar, line 3: Standardizing on just including the module - "validate" is such a generic name for a method that it's a bit more readable to have the context of what we're validating.

redocly.yaml
```diff
@@ -0,0 +1,23 @@
+# See https://redocly.com/docs/cli/configuration/
```
WalterMoar, line 1: The Redocly extension was prompting if we wanted to create a config file. Added it so that everyone doesn't get annoyed by the prompts.

PR #1305, opened 2024-03-11
docs: FORMS-1047 http 401 and 403 for most endpoints

WalterMoar:
# Description

The SSO work is a big change that requires us to communicate the changes to all our clients. Since we are doing this communication it would be a good time to fix a longstanding bug: authentication failures are returning HTTP status 403 (Forbidden) when they should be returning 401 (Unauthorized).

PR1258 will make the 403->401 change, but the API documentation also needs to be updated. Putting the docs into a separate ticket/PR since they need to be done anyway and can be done out of sync with the PR1258 deployment (and done early if possible).

Major changes:
- Document that nearly every endpoint can return 401 and 403 responses
- Update the docs to provide the schema and examples for 401 and 403
- For consistency rename the `UnauthorizedError` response simply to `Unauthorized`

Minor housekeeping:
- Add the `Template` tag - it was being referenced but was never defined
- Reference responses (202 204, etc) whenever possible
- Add a 404 response for `/submissions/{formSubmissionId}/option` - this was to make the linter happy
- Put the schemas into alphabetical order (they were _mostly_ alphabetized, which made things confusing)
- Removed the `Conflict` schema and response since they weren’t used
- Moved the `BadRequest`, `NotFound`, and `ValidationError` schemas into their responses (they were not schemas)

## Types of changes

## Checklist

## Further comments

Docs here: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-1305/api/v1/docs

### DIFF
--- a/app/src/docs/v1.api-spec.yaml
+++ b/app/src/docs/v1.api-spec.yaml
@@ -84,6 +84,10 @@ tags:
       All calls are secured by a role that operational team members will be
       granted and allow fetching some details about forms without needing a
       user->form permission.
+  - name: Template
+    description: >
+      These API endpoints generate documents from document templates and form
+      submission data.
 paths:
   /forms:
     get:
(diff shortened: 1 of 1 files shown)


usingtechnology (approved):
Good stuff, thanks for getting that in there.

PR #1306, opened 2024-03-12
fix: FORMS-1048 remove submission id unique constraint

WalterMoar:
# Description

There’s a constraint in the database that a submission’s `confirmationId` has to be unique. We just had a collision in production, and someone’s submission failed because of the constraint. Need to double-check that this uniqueness is never relied on.

The `confirmationId` can’t be considered unique because the keyspace (16^8 = ~4 billion) is too small. The `id` is essentially unique, but the `confirmationId` should assume collisions, even within the same form.

A quick scan of the backend code didn't come up with any obvious places where unique `submissionId`s are needed. Thinking it's best to fix the problem so that submissions don't fail, and then if we do have problems with duplicates we at least have the data and can write tests and then fix the problems.

## Types of changes

## Checklist

### DIFF
--- a/app/src/db/migrations/20240312164557_remove-confirmationid-constraint.js
+++ b/app/src/db/migrations/20240312164557_remove-confirmationid-constraint.js
@@ -1,0 +1,23 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = function (knex) {
+  return Promise.resolve().then(() =>
+    knex.schema.alterTable('form_submission', (table) => {
+      table.dropUnique('confirmationId');
+    })
+  );
+};
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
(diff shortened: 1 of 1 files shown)


WalterMoar:
I don't know if it's much of a test, but I changed a submission's `confirmationId` to create a duplicate. I couldn't change the `id` due to a bunch of foreign key issues. No glaring problems with duplicates, though.
image

usingtechnology (approved):
Approving, but I let's be aware that there is a chance we cannot rollback/down this... if we have duplicate data in the column we won't be able to add back the constraint without modifying the data.  VERY slim chance and we can deal with it when/if it happens.

WalterMoar:
> Approving, but I let's be aware that there is a chance we cannot rollback/down this... if we have duplicate data in the column we won't be able to add back the constraint without modifying the data. VERY slim chance and we can deal with it when/if it happens.

Oh, good point! The `migrate:down` does work without duplicates, but I never thought of that :| Good thing we don't tend to rollback.

PR #1315, opened 2024-03-27
fix: FORMS-1017 move file uploads when updating submission

WalterMoar:
# Description

There is a bug where the experimental feature for “Draft” submissions does not set the `submissionId` or move the files to the `chefs/prod/submissions/` path. This works for the `POST` call, so the `PUT` call should be updated to match its behaviour.

## Types of changes

## Checklist

### DIFF
--- a/app/src/forms/submission/service.js
+++ b/app/src/forms/submission/service.js
@@ -1,8 +1,12 @@
 const { v4: uuidv4 } = require('uuid');
+
+const log = require('../../components/log')(module.filename);
+
 const { Statuses } = require('../common/constants');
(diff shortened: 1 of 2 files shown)


app/src/forms/submission/service.js
```diff
@@ -51,6 +55,38 @@ const service = {
     return [];
   },
 
+  /**
+   * Given the data for a submission, find the file uploads.
+   *
+   * @param {object} data the data for a submission.
+   * @returns an array of file UUIDs.
+   */
+  _findFileIds: (data) => {
```
WalterMoar, line 64: The is a direct duplicate of what's being done in the form service. If possible, we should not be doing submissions in the form service, but that's something to figure out later.

vijaivir (approved):

PR #1318, opened 2024-04-09
refactor: Forms 1066 composition api upgrade

jasonchung1871:
# Description
This is the initial step for upgrading from the options API to the composition API. This also includes updating vuetify to a newer version which required changes to the frontend as some vuetify components were experimental and are now stable but had breaking changes.

Updated most of the views to use the composition API.

Test coverage is at or near 100% now for
 - views/About.spec.js
 - views/Error.spec.js
 - views/Login.spec.js
 - App.spec.js
 - file/Download.spec.js
 - form/Create.spec.js (Close to 100%, unable to mock the onBeforeRouteLeave from the router)
 - form/Design.spec.js

## Types of changes

## Checklist


WalterMoar:
@usingtechnology How hard would it be to merge these changes into PR1258?

WalterMoar:
Looks like the icon for deleting all checkboxed submissions has been replaced by "event":
image

WalterMoar:
> Looks like the icon for deleting all checkboxed submissions has been replaced by "event": image

Also: the font for the submissions table is smaller, and there's a space between the table and the line underneath it.

jasonchung1871:
> Looks like the icon for deleting all checkboxed submissions has been replaced by "event": image

I need to redeploy but this is fixed now, good catch!

WalterMoar:
One little thing: on the Form Manage page, when I open an item the header gets bigger but the background stays the same colour. In dev the header stays the same size but the background gets darker:
image

jasonchung1871:
> > Looks like the icon for deleting all checkboxed submissions has been replaced by "event": image
> 
> Also: the font for the submissions table is smaller, and there's a space between the table and the line underneath it.

I'm not sure why this is. Vuetify may have changed their default font size and all of their spacing.

WalterMoar:
> I'm not sure why this is. Vuetify may have changed their default font size and all of their spacing. Edit: It seems to be a problem with Bootstrap and Vuetify. Bootstrap is making all of the font much finer and smaller.

Oh, OK! The spacing and the line seem fine. The new green spacing is nice separation and the red spacing and horizontal line are just new and different, not necessarily anything wrong with them. It would be awesome if bootstrap would behave itself, though, in case there are accessibility concerns about the fonts shrinking.
image

WalterMoar:
> One little thing: on the Form Manage page, when I open an item the header gets bigger but the background stays the same colour. In dev the header stays the same size but the background gets darker: image

It looks like this is a change in the default behaviour of the Vuetify Expansion Panel component. Probably best to stick with the default behaviour?

app/frontend/src/components/forms/manage/AddTeamMember.vue
```diff
@@ -220,8 +220,8 @@ export default {
                 <template v-else>
                   <v-list-item
                     v-bind="props"
-                    :title="`${item?.raw?.fullName} (${item?.raw?.email})`"
-                    :subtitle="`${item?.raw?.username} (${item?.raw?.idpCode})`"
+                    :title="`${item?.fullName} (${item?.email})`"
+                    :subtitle="`${item?.username} (${item?.idpCode})`"
```
WalterMoar, line 224: Hmm, something isn't working here. I'm seeing "undefined" for the fields:
image

app/frontend/src/components/forms/submission/StatusPanel.vue
```diff
@@ -403,14 +403,14 @@ export default {
...
+                  <v-chip v-bind="props" :text="item?.fullName" />
                 </template>
                 <!-- users found in dropdown -->
                 <template #item="{ props, item }">
                   <v-list-item
                     v-bind="props"
-                    :title="`${item?.raw?.fullName} (${item?.raw?.email})`"
-                    :subtitle="`${item?.raw?.username} (${item?.raw?.idpCode})`"
+                    :title="`${item?.fullName} (${item?.email})`"
+                    :subtitle="`${item?.username} (${item?.idpCode})`"
```
WalterMoar, line 413: Similar to the Team Management page, these are displaying as "undefined" (possibly the same underlying problem?)
image

app/frontend/src/components/forms/manage/AddTeamMember.vue
jasonchung1871: another very good catch, looking into this now

jasonchung1871: figured it out. turns out they removed "raw" from their tables when you're accessing the data but in their other components they kept it. weird decision to not keep things consistent.

WalterMoar: Looking good. They'll probably remove "raw" from the rest of them in the next update!

app/frontend/src/components/forms/manage/TeamManagement.vue
```diff
@@ -582,7 +582,7 @@ export default {
         <v-checkbox-btn
           v-if="!disableRole('submission_reviewer', item, form.userType)"
           key="submission_reviewer"
-          v-model="item.columns.submission_reviewer"
+          v-model="item.submission_reviewer"
```
WalterMoar, line 585: Might not be related to this line of code, but if I go to Team Management, click "Add a New Member" icon, it comes up as IDIR. I click "owner" and owner gets a checkbox as expected. Then switch to BCeID Basic, and when you click "form_submitter" it does not get a checkbox.
image

WalterMoar: Can we add tests for this kind of thing? Or is that something that's better to be done when these go to the Composition API?

jasonchung1871: thanks! it seems like it's not clearing the fields when changing the identity provider, allowing for the selected role "owner" to exist in basic bceid when it should not. i can try to add it in as a test!

app/frontend/src/components/forms/submission/StatusPanel.vue
```diff
@@ -407,10 +407,11 @@ export default {
                 </template>
                 <!-- users found in dropdown -->
                 <template #item="{ props, item }">
+                  {{ item.raw }}
```
WalterMoar, line 410: Was this for debugging? I'm seeing:
image

jasonchung1871: oh shoot yeah haha, i'll remove that while i'm fixing the previous issue

app/frontend/src/components/forms/submission/ManageSubmissionUsers.vue
```diff
@@ -131,7 +131,7 @@ export default {
     },
 
     filterObject(_itemTitle, queryText, item) {
-      return Object.values(item.raw)
+      return Object.values(item)
```
WalterMoar, line 134: Looks like something isn't working here too, but not necessarily with this line. If you create a form that allows drafts, start a submission and save as draft, then view the draft... then add team member, switch to basic bceid and type "DEV_WALTER" nothing happens:
image
But if you then click the "X" it appears:
image

Expected: in dev as soon as you type the last "R" in the BCeID username, it pops up the user info:
image

jasonchung1871: ah yeah, i believe this was intentional as some users using bceid basic were using their personal email addresses. to prevent it from listing all email addresses, we opted to have the user type in the exact username or address.

WalterMoar:
Somehow the BCeID Basic users can now create forms:
image
This is something that should be prevented in the backend, but I think that the app is unfortunately set up where the business rules are only in the frontend - we need both.

app/frontend/src/components/forms/submission/ManageSubmissionUsers.vue
WalterMoar: Just to clarify - in the top image I have typed in the entire username but it didn't show the user information. The bottom image is the dev behaviour using the same user.

app/frontend/src/components/forms/submission/StatusPanel.vue
jasonchung1871: This should be resolved now

app/frontend/src/components/forms/manage/TeamManagement.vue
jasonchung1871: This should be resolved now with added tests

app/frontend/src/components/forms/submission/StatusPanel.vue
jasonchung1871: This should be resolved now

jasonchung1871:
> Somehow the BCeID Basic users can now create forms: image This is something that should be prevented in the backend, but I think that the app is unfortunately set up where the business rules are only in the frontend - we need both.

This should be resolved now. Removal of the BaseStepper component left out the built in BaseSecure component.

jasonchung1871:
> Have all these changes been re-tested since the merge with the SSO code? It would be great if the reviewers could focus on code style, completeness, etc, and not be testing for bugs.

Everything has been retested, and should be in working order. At least the things you've mentioned, and the views that were changed. No new tests were really made. I was hoping to get this through so that other tests could be incrementally written. Just need the updated vuetify, vite, vitest.

WalterMoar (approved):
Approved, if Jason C is confident that everything has been tested.

usingtechnology:
i'll take a look this morning. @jasonchung1871 - "I am concerned about the number of problems that were found in my first tests" were these backend tests or frontend tests? Most of the backend tests required changes to the mocked middleware, particularly adding in the jwtService and userService.currentUser. Frontend tests, not so sure about except changes made where we don't protect with an IDP check but rather with a permissions check.

jasonchung1871:
> i'll take a look this morning. @jasonchung1871 - "I am concerned about the number of problems that were found in my first tests" were these backend tests or frontend tests? Most of the backend tests required changes to the mocked middleware, particularly adding in the jwtService and userService.currentUser. Frontend tests, not so sure about except changes made where we don't protect with an IDP check but rather with a permissions check.

They should just be front end tests as none of the back end was altered.

PR #1317, opened 2024-04-08
feat: FORMS-1042 new document templates backend

WalterMoar:
# Description

Backend work for storing one CDOGS template for a form.

DB:
- document_template: id, formId, active, filename, template binary?, create/update dates & users
- form: enableDocumentTemplates

API:
- GET /forms/:formId/documentTemplates (gets list of documentTemplateIds and filenames)
- GET /forms/:formId/documentTemplates/:documentTemplateId
- POST /forms/:formId/documentTemplates
- DELETE /forms/:formId/documentTemplates/:documentTemplateId (soft)

The double-GET is an annoyance when we only have a single template per form, but we will eventually have the case where forms have many templates and we don’t want to return all of the BLOBs in one call.

## Types of changes

New feature (non-breaking change which adds functionality)

## Checklist

- [x] I have read the [CONTRIBUTING](CONTRIBUTING.md) doc
- [x] I have checked that unit tests pass locally with my changes
- [x] I have run the npm script lint on the frontend and backend
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)
- [x] I have approval from the product owner for the contribution in this pull request

## Further comments

This is the first version of the backend, so that frontend work can begin. Still need to:
- Switch to using the object storage for the templates
- Write the openapi docs

PR #1319, opened 2024-04-11
feat: Upload Default CDOGS Template Frontend

vijaivir:
# Description
This PR introduces new functionality that allows users to provide a pre-existing CDOGS template that form submitters can use in the printing dialog. This complements #1317 and provides the front-end logic for the feature. 

A new Form Functionality setting has been added that opens up the 'CDOGS Template' component where users may manage a document template associated with that form. Users can upload, download, and delete the document template.  

This was a highly requested feature on [Fider](https://chefs-fider.apps.silver.devops.gov.bc.ca/posts/140/one-click-cdogs-print). 

[Jira Link ](https://bcdevex.atlassian.net/jira/software/c/projects/FORMS/boards/95?selectedIssue=FORMS-1110&sprints=749)

## Types of changes

## Checklist


app/frontend/src/components/designer/FormViewer.vue
```diff
@@ -179,6 +179,8 @@ export default {
     window.addEventListener('beforeunload', this.beforeWindowUnload);
 
     this.reRenderFormIo += 1;
+    const formStore = useFormStore();
+    formStore.setFormId(this.form.id);
```
vijaivir, line 183: I am using the form store here to pass the formId from the Form Viewer to the Print Options. I couldn't find an alternative to do this so please let me know if you have some suggestions.

jasonchung1871: The form id should already be set by the parent components, we fetch it and set it many times. It should work to just get the form from the form store and get the id from the form object inside of PrintOptions. Upon testing though, it seems like you would need to create a prop inside of `PrintOptions` maybe call it `f` for form id then create a computed value called `formId` that returns `this.f ? this.f : this.form.id`. Then in `FormViewerActions` you just need to pass the `formId` prop to the `PrintOptions` component so it would look like:

```
<PrintOptions
     :f="formId"
...
```
This would allow you to obtain the form id inside of `PrintOptions` without creating another variable for storing the form id. It'll get confusing in the future if we have multiple places where we store the form id.

vijaivir: Thank you so much!

PR #1320, opened 2024-04-16
docs/FORMS-1042 openapi docs for document templates

WalterMoar:
# Description

Adding the OpenAPI documentation for the new document templates routes. This documentation was skipped so that the routes could be merged and the frontend work started.

## Types of changes

## Checklist

### DIFF
--- a/app/src/docs/v1.api-spec.yaml
+++ b/app/src/docs/v1.api-spec.yaml
@@ -62,7 +62,7 @@ tags:
            a document template for a form.
         2. Subsequent calls to the `POST` route should be paired with calls to
            the `DELETE` route to delete the previously active template.
-        3. When it comes time to print, the
+        3. When it comes time to generate the document, the
            `GET /forms/{formId}/documentTemplates` route is used to get "all"
            the document template metadata for a form (although there is only one
            active document template). This route only returns metadata and not
--- a/app/src/docs/v1.api-spec.yaml
+++ b/app/src/docs/v1.api-spec.yaml
@@ -45,6 +45,34 @@ tags:
       This section of the API includes endpoints used to perform various
       operations related to form drafts, for example create or publish a draft
       from a specific version of a form.
+  - name: Document Template
+    description: >
+      Documents can be generated using the Common Document Generation Service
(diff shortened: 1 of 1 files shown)


app/src/docs/v1.api-spec.yaml
```diff
@@ -2028,7 +2228,7 @@ paths:
           schema:
             type: string
             format: uuid
-            example: c6455376-382c-439d-a811-0381a012d696
+            example: aeb3b705-1de5-4f4e-a4e6-0716b7671034
```
WalterMoar, line 2231: The docs were using random UUIDs and this often meant that the docs would say that a request using on ID would have a response with a different ID. Standardizing on using the same UUID for all sample form IDs, form submission IDs, etc.

app/src/docs/v1.api-spec.yaml
```diff
@@ -4155,20 +4417,51 @@ components:
               description: >-
                 The original file type of the `content` document template.
               example: txt
+    TimeStampCreate:
```
WalterMoar, line 4420: The example responses are all using TimeStampUserData, which include an updatedBy value and a different updatedAt timestamp. However, when creating a new object this is not correct - the updatedBy is not defined and the updatedAt is the same as the createdAt. Made this correct for the document template docs but will have to fix the others later.

app/src/docs/v1.api-spec.yaml
```diff
@@ -4155,20 +4417,51 @@ components:
...
+          description: >
+            The UTC date and time when the object was last modified.
+          example: '2020-06-04T18:49:20.672Z'
     TimeStampUserData:
       type: object
       properties:
         createdBy:
           type: string
+          description: >
+            The username of the person who created the object.
```
WalterMoar, line 4449: We should always be using descriptions. Once these are all set up we can configure the extension to enforce it.

WalterMoar:
New docs deployed here: https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-1320/api/v1/docs#tag/Document-Template

vijaivir (approved):
Looks good! I have also updated the API call so I'm not relying on the template from the /forms/{formId}/documentTemplates anymore.

PR #1326, opened 2024-04-24
Add encoded JWT token to CHEFS context

bcvesalink:
# Description

This PR is to add the encoded JWT token to CHEFS context to use it in `Calculated values` same way as `token` (decoded JWT token) and `user`. That would mean we can use it to call CHEFS API securely by using full `encoded JWT token` in a header instead unrestricted `API key`. Also, this would help identify the user who calling the API, which is required by RLS (Row level Security)

## Types of changes

## Checklist

### DIFF
--- a/app/frontend/src/components/designer/FormViewer.vue
+++ b/app/frontend/src/components/designer/FormViewer.vue
@@ -148,6 +148,7 @@ export default {
         evalContext: {
           token: this.tokenParsed,
           user: this.user,
+          encoded_token: this.keycloak.token,
         },
       };
     },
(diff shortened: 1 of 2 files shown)


usingtechnology (dismissed):
Looks good.

WalterMoar (changes requested):
Wouldn't this allow a form to impersonate the user for *any* forms that the user owns? What about problems with the token being valid for the standard realm?

WalterMoar:
I created a proof of concept form that, when displayed, will delete all the user's forms.
- Log into https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-1326
- Create a form
- Visit https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-1326/form/submit?f=c6a7fc79-d70b-4aa0-b589-22b6b21a037f

bcvesalink:
yes, there is a security issue making encoded token available for form designers, so closing this PR

PR #1329, opened 2024-04-24
Forms 1078 token refresh failure

jasonchung1871:
# Description
Created a function for updating the token. This function is executed when the keycloak detects the token as expired, as well as every normal 60 seconds, as well as whenever there is an update token failure. Upon failure of updating the token, it will create an interval to keep updating the token, upon success, it will clear that interval.

## Checklist

### DIFF
--- a/app/frontend/src/main.js
+++ b/app/frontend/src/main.js
@@ -196,20 +196,53 @@ function loadKeycloak(config) {
         authStore.ready = true;
         typeof options.onReady === 'function' && options.onReady();
       };
+
+      let updateTokenInterval;
+      // The expired token is also used for general update token failures
+      // i.e., network disconnection
+      let expiredTokenInterval;
+
+      function updateToken(seconds) {
+        try {
+          keycloak
+            .updateToken(seconds)
+            .success((refreshed) => {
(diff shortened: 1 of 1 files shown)


app/frontend/src/main.js
```diff
@@ -196,21 +196,60 @@ function loadKeycloak(config) {
...
+              // If we're failing to update the token, we will clear the token
+              keycloak.clearToken();
+            })
+            .catch(() => {
+              // If we're failing to update the token, we will clear the token
+              keycloak.clearToken();
+            });
+        } catch (error) {
+          // If we're failing to update the token, we will clear the token
+          keycloak.clearToken();
```
jasonchung1871, line 229: We don't do anything with the errors because so far, it seems like the errors that are caught or thrown are just undefined or a boolean value.

vijaivir (approved):
I reviewed this PR a couple of weeks ago, I have not been able to reproduce the error so I can't say that this fixes the issue. But based on your sprint demo, I am approving this. 

Nice work!

jasonchung1871:
Yeah, it seems to work on specific browsers. This one needs to be looked into further.

PR #1332, opened 2024-04-26
Bug Fix: 401 Error for the Form Subscriptions 

vijaivir:
# Description

The call to the subscriptions API was not protected by any permissions. It was being called even if someone without the `FORM_UPDATE` permission navigated to manage the form. The 'Event Subscription' panel was also being loaded. This resulted in a 401 error on the manage form page as they didn't have permissions but the API call was still being made. 

The fix adds a permission check for the API call and the component's visibility. 

Jira Link: https://bcdevex.atlassian.net/browse/FORMS-958

## Types of changes

## Checklist


WalterMoar (approved):
Functionally it looks great - no more API calls that error out, and the layout fix is awesome.

Don't know if/how you could write tests for these but otherwise looks good to me.

vijaivir:
> Functionally it looks great - no more API calls that error out, and the layout fix is awesome.
> 
> Don't know if/how you could write tests for these but otherwise looks good to me.

I have added some tests that check for appropriate permissions 👍

PR #1333, opened 2024-04-29
fix/FORMS-1194 redo username logging

WalterMoar:
# Description

We used to have something like `"username": "WAMOAR"` in the logs, but this is no longer appearing after the move to the SSO standard realm. Update the logging code to provide the IDIR or BCeID username, if they exist.

## Types of changes

## Checklist

## Further comments

We unfortunately have no tests around this, but need to get it into prod to help diagnose other problems.

### DIFF
--- a/app/src/components/log.js
+++ b/app/src/components/log.js
@@ -77,9 +77,11 @@ const httpLogger = logger({
     const token = jwt.decode((req.get('authorization') || '').slice(7));
     return {
       azp: (token && token.azp) || undefined,
+      bceid_username: (token && token.bceid_username) || undefined,
       contentLength: res.get('content-length'),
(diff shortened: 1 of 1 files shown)


WalterMoar:
IDIR:
image
BCeID:
image

usingtechnology (commented):
just thinking... maybe we should enhance the `idpService` so we can get values from various tokens, then we can log using the `user` object field names...`username`, `email`.
i think we can make `idpService.parseToken` non-async (or add something similar) and we can just use that here. that method is only async in case the provider list hasn't been fetched.

WalterMoar:
Closing this for now. Will live with the problem until there is time to figure out the SSO changes needed.

PR #1338, opened 2024-05-06
fix/FORMS-1202 update follow-redirects dependency

WalterMoar:
# Description

The GitHub Dependabot process has created an alert for the **follow-redirects** dependency. To satisfy the requirements outlined in the Security Threat and Risk Assessment's (STRA) Statement of Acceptable Risks (SoAR), this vulnerability must be handled by updating the package version (or mitigated in some other way, if updating the package is not possible).
- https://github.com/bcgov/common-hosted-form-service/security/dependabot/161
- https://github.com/bcgov/common-hosted-form-service/security/dependabot/166 

## Types of changes

## Checklist

## Further comments

We don't directly use this package, it is a dependency of Axios. It's difficult to test as it would require setting up CHES, CDOGS, the BC Geo API, or a subscription service that redirects to another location. This is a low-risk change that we're doing for the sake of completeness, even though we probably don't use the functionality of this package.


WalterMoar:
Since this is a dependency of Axios I did a manual regression test of using CHES, CDOGS, the address components, and the event subscription.

PR #1339, opened 2024-05-06
fix/FORMS-1215 csv export of datagrid data

WalterMoar:
# Description

The frontend sends in a list of fields that are to be included in the export. For components in a Data Grid they will have names like `dataGrid.simpletextinput` but in the submission data they will be `dataGrid.0.simpletextinput`, `dataGrid.1.simpletextinput`, etc. A comparison is made between the list of fields and the submission name, and it is failing because of the array index `.0`, etc. The fix is to remove the array index before doing the comparison with the list of fields.

## Types of changes

## Checklist

### DIFF
--- a/app/tests/fixtures/submission/kitchen_sink_submission_data_export_datagrid_fields_selection.json
+++ b/app/tests/fixtures/submission/kitchen_sink_submission_data_export_datagrid_fields_selection.json
@@ -10,13 +10,10 @@
   "email",
   "forWhichBcLakeRegionAreYouCompletingTheseQuestions",
   "didYouFishAnyBcLakesThisYear",
-  "oneRowPerLake.0.lakeName",
-  "oneRowPerLake.0.closestTown",
-  "oneRowPerLake.0.numberOfDays",
-  "oneRowPerLake.0.dataGrid.0.fishType",
-  "oneRowPerLake.0.dataGrid.1.fishType",
-  "oneRowPerLake.0.dataGrid.0.numberCaught",
-  "oneRowPerLake.0.dataGrid.1.numberCaught",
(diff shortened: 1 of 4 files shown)


app/tests/fixtures/submission/kitchen_sink_submission_data_export_datagrid_fields_selection.json
```diff
@@ -10,13 +10,10 @@
...
   "didYouFishAnyBcLakesThisYear",
-  "oneRowPerLake.0.lakeName",
-  "oneRowPerLake.0.closestTown",
-  "oneRowPerLake.0.numberOfDays",
-  "oneRowPerLake.0.dataGrid.0.fishType",
-  "oneRowPerLake.0.dataGrid.1.fishType",
-  "oneRowPerLake.0.dataGrid.0.numberCaught",
-  "oneRowPerLake.0.dataGrid.1.numberCaught",
-  "oneRowPerLake.0.dataGrid.0.numberKept",
-  "oneRowPerLake.0.dataGrid.1.numberKept"
```
WalterMoar, line 21: This fixture was incorrect - this is not how the fields selection works.

app/tests/unit/forms/form/exportService.spec.js
```diff
@@ -269,7 +454,7 @@ describe('_buildCsvHeaders', () => {
     // get result columns if we need to filter out the columns
     const result = await exportService._buildCsvHeaders(form, submissionsExport, 1, fields, true);
 
-    expect(result).toHaveLength(20);
+    expect(result).toHaveLength(29);
```
WalterMoar, line 457: This test was incorrect, so the length is different now that it's returning the proper column names.

PR #1340, opened 2024-05-07
fix/FORMS-1208 remove crypto dependency

WalterMoar:
# Description

When running `npm install` there is a deprecation warning for the `crypto` package. It looks like this package is only used to create a hash in the CDOGS service.

> crypto@1.0.1: This package is no longer supported. It's now a built-in Node module. If you've depended on crypto, you should switch to the one that's built-in.

When investigating the use of the package, it was noticed that the following code is never used:
- `CdogsService.health()`
- `CdogsService.templateRender()`
- `CdogsService.getTemplate()`
- `CdogsService.uploadTemplate()`
- `CdogsService.getHash()`

We are storing the templates on our side rather than uploading them to CDOGS, because there is no guarantee how long the template will remain on CDOGS. If we ever do mass document generation with the same template then we can bring the required method(s) back. In the meantime there is no point in maintaining and writing tests for code we don't use.

This also resulted in the removal of the `form-data` dependency, although it's a sub-dependency of other dependencies so it's still installed.

Added `axios-mock-adapter` as a dev dependency for the unit tests. There previously were no unit tests for the CdogsService, and no mocking of Axios. The tests are minimal but are a starting point if we ever run into bugs and need regression tests for the fix.

## Types of changes

## Checklist


WalterMoar:
Tested with the three ways to generate a document:
1. Draft submission with an uploaded template
2. Saved submission with an uploaded template
3. Saved submission with a stored template

PR #1359, opened 2024-05-17
refactor: FORMS-1265 create new problems in a consistent way

WalterMoar:
# Description

In `userAccess.hasRolePermissions` Problems were being created in four different ways:
1. Create Problem and call `next` and end the request (code smell - early returns)
1. Create Problem and call `next` without ending the request (code smell - early returns)
1. Create Problem and return it  (code smell - early returns) (!!! Returning without calling `next` is a bug and the request will hang indefinitely)
1. Create Problem and throw it, which is caught and calls `next` with it

For consistency standardize on the last way of doing things. It's confusing when reading the code and there are four different ways - in a single function - of doing the same thing.

Also:
- make the Problem creation code consistent with detail always on its own line, even if it's short enough for the whole create to be on one line
- update the Problem creation in `filterMultipleSubmissions` to also throw Problems instead of using early returns
- rename `hasFormRole` to `_hasFormRole` and do not export it - it's only used by `hasRolePermissions`
  
## Types of changes

Refactor (no functional change but an improvement to code quality)

## Checklist

## Further comments

No tests for this - it passes the existing tests (questionable) but my next PR will be to fully test `hasRolePermissions`.

### DIFF
--- a/app/src/forms/auth/middleware/userAccess.js
+++ b/app/src/forms/auth/middleware/userAccess.js
@@ -64,7 +64,9 @@ const _hasAnyPermission = (userPermissions, requiredPermissions) => {
  */
 const _getForm = async (currentUser, formId, includeDeleted) => {
   if (!uuid.validate(formId)) {
-    throw new Problem(400, { detail: 'Bad formId' });
+    throw new Problem(400, {
+      detail: 'Bad formId',
+    });
   }
 
   const forms = await service.getUserForms(currentUser, {
@@ -102,7 +104,9 @@ const currentUser = async (req, _res, next) => {
     if (bearerToken) {
       const ok = await jwtService.validateAccessToken(bearerToken);
       if (!ok) {
-        throw new Problem(401, { detail: 'Authorization token is invalid.' });
+        throw new Problem(401, {
+          detail: 'Authorization token is invalid.',
+        });
       }
     }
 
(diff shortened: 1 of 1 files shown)


app/src/forms/auth/middleware/userAccess.js
```diff
@@ -277,26 +287,23 @@ const filterMultipleSubmissions = () => {
...
+        detail: 'Current user does not have required permission(s) for to delete submissions',
+      });
     } catch (error) {
       next(error);
     }
   };
 };
 
-const hasFormRole = async (formId, user, role) => {
+const _hasFormRole = async (formId, user, role) => {
```
WalterMoar, line 306: This is only used in this file, so mark it as internal and don't export it.

app/src/forms/auth/middleware/userAccess.js
```diff
@@ -379,68 +384,64 @@ const hasRolePermissions = (removingUsers = false) => {
...
+              throw new Problem(401, {
+                detail: "You can't remove a form designer role.",
+              });
             }
           }
         }
       } else {
         const userId = req.params.userId || req.query.userId;
         if (!userId || (userId && userId.length === 0)) {
-          return new Problem(401, { detail: 'User Id not found on request.' });
```
WalterMoar, line 402: Yikes! When `next()` isn't called here the request will hang.

PR #1370, opened 2024-05-28
External API

usingtechnology:
# Description
Allow form components (ex. Drop Down) to call external APIs in a secured manner. Form designers can specify what headers to include when calling their API. CHEFs can provide the current user's information (name, email, etc) and the current form id.

CHEFS Admins will need to approve the external api before it can actually fetch any data. Form Owner/Designers will have to ask for approval and explain how they plan to use this feature.

## Types of changes

feat (a new feature)

## Checklist


app/src/forms/proxy/routes.js
```diff
@@ -0,0 +1,21 @@
...
+// need to allow cors for OPTIONS call
+// formio component will call OPTIONS pre-flight
+routes.options('/external', cors());
+
+// called with encrypted headers, no current user!!!
+routes.get('/external', cors(), async (_req, res, next) => {
+  await controller.callExternalApi(_req, res, next);
+});
+
+routes.post('/headers', currentUser, async (_req, res, next) => {
```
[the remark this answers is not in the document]
WalterMoar: Just going to re-open this - we've been ignoring them but not dismissing them, so we know the routes that need to have rate limiting. Hopefully I'll be able to get the rate limiter (FORMS-1303) in place in the next sprint or two.

robinwn:
A couple of notes:
- I don't see the OIDC 'sub' attribute available; I do see other non standard attributes that might be the same.
- should be using public/private key pairs; encrypt with the public key, receiving endpoint will use the private key to decrypt.  Loss or exposure of the key used here means a security incident; loss of the public key not so much.

app/src/forms/proxy/routes.js
```diff
@@ -0,0 +1,21 @@
...
+
+const routes = require('express').Router();
+
+// need to allow cors for OPTIONS call
+// formio component will call OPTIONS pre-flight
+routes.options('/external', cors());
+
+// called with encrypted headers, no current user!!!
+routes.get('/external', cors(), async (_req, res, next) => {
+  await controller.callExternalApi(_req, res, next);
```
WalterMoar, line 14: This is the part I'm not getting - why use the bearer token to create this encrypted object, and then use the encrypted object in this call, and extract what was in the bearer token? Why not use the bearer token here?

usingtechnology: This call comes from the form component (like drop down control) which doesn't have access to the token unless we make it available to the form designer. If we do that, then we have no control over what a form designer uses that token for...

So the initial call on load we do have the token because CHEFS frontend is making the call (just like fetching the form schema). Only the CHEFS backend can decrypt this and read the true token. Not sure if we will even leave the token in that encrypted bundle or not - but it is up to the designer whether they want it added to the call to their API.

usingtechnology:
> A couple of notes:
> 
> * I don't see the OIDC 'sub' attribute available; I do see other non standard attributes that might be the same.
> * should be using public/private key pairs; encrypt with the public key, receiving endpoint will use the private key to decrypt.  Loss or exposure of the key used here means a security incident; loss of the public key not so much.

Thanks for taking a look @robinwn! 

Couple of things going on here (nothing set in stone either)...

The non-standard fields are the "standard" CHEFS user object attributes. That is the same object/attributes that the form designer can programmatically use when designing the form. The values are pulled from the token, but each Identity Provider has a different set of attributes so we have a mapping process to standardize them. So those attributes are not OIDC but CHEFS. On the receiving end (the external API) you could chose to include the current user's token and then decrypt and parse and pick over whatever attributes you want. Keep in mind that token was valid at form load and could be expired. So verifying the token would have to ignore the `iat`/ `issued at` claim. But I do like your thinking of using OIDC attributes... maybe we can make that an option - CHEFS user attributes or OIDC user attributes.

There will be more encryption/decryption algorithms at some point. It is possible that we support pub/private keys but the main issue there is data payload size. Thinking forward to message/events queue, pub/private just can't handle the submissions data. Unless we had an absurdly large and slow private key.

app/src/forms/proxy/routes.js
WalterMoar: Thanks for the info! I completely forgot that the form runs in a sort-of sandbox. This does bring up another problem I need to sort out, and maybe that solution will be useful - will see! (and in the meantime I'll try to understand the rest)

app/src/forms/form/controller.js
```diff
@@ -387,4 +387,52 @@ module.exports = {
       next(error);
     }
   },
+  listExternalAPIs: async (req, res, next) => {
```
WalterMoar, line 390: One of my responsibilities this quarter is to make sure that maintainability of the backend improves (we currently rate a `D`). This file (and many others) are flagged by [Code Climate](https://codeclimate.com/github/bcgov/common-hosted-form-service/app/src/forms/form/controller.js) as having too many functions (or too many lines, etc). We're never going to fix complaints about similar code where a `delete` function calls `service.delete`, `update` calls `service.update`, etc, but the size of some of the files is getting awkward. The way Code Climate is set up we don't really have a way of seeing problems until merges to main happen - so it's entirely expected that new code will follow the old pattern that Code Climate complains about! My way of working has been to check Code Climate for any files I change, and see if I can avoid adding to the list of issues.

Unfortunately I don't have a new pattern figured out yet that can be worked from, but it would be great if new code didn't add to the maintenance that will need to be done. Two very vague ideas I had were:
1. Have subdirectories for each "feature", like `forms/form/externalApi`, and that directory would have its own `route.js` (or not?) and `controller.js` and `service.js`. Preferred?
2. Have multiple files in the `forms/form` directory, like `routeExternalApi.js` (?) , `controllerExternalApi.js`, `serviceExternalApi.js`, etc.

If we could figure a pattern out, then eventually I'll start to break down the big controller and service files to match the pattern.

app/src/forms/form/service.js
```diff
@@ -1040,6 +1043,114 @@ const service = {
...
+      trx = await ExternalAPI.startTransaction();
+      data.id = uuidv4();
+      // set status to SUBMITTED
+      data.code = ExternalAPIStatuses.SUBMITTED;
+      await ExternalAPI.query(trx).insert({
+        ...data,
+        createdBy: currentUser.usernameIdp,
+      });
+
+      await trx.commit();
```
WalterMoar, line 1109: I see this pattern a lot in the services (and I've followed it too!) but I'm starting to wonder about it. Does this transaction `trx` actually do anything? We're only doing a single `insert`, so if it fails then there is nothing to roll back, is there?

app/src/forms/form/controller.js
usingtechnology: let's come up with something and try it out. the folder approach is probably the one I'd try first.

app/src/forms/form/controller.js
```diff
@@ -387,4 +387,52 @@ module.exports = {
...
+      const response = await service.listExternalAPIStatusCodes();
+      res.status(200).json(response);
+    } catch (error) {
+      next(error);
+    }
+  },
+  createExternalAPI: async (req, res, next) => {
+    try {
+      const response = await service.createExternalAPI(req.params.formId, req.body, req.currentUser);
+      res.status(200).json(response);
```
WalterMoar, line 417: I think this should be a `201` status for an `insert`?

app/src/forms/form/controller.js
```diff
@@ -387,4 +387,52 @@ module.exports = {
...
+      const response = await service.updateExternalAPI(req.params.formId, req.params.externalAPIId, req.body, req.currentUser);
+      res.status(200).json(response);
+    } catch (error) {
+      next(error);
+    }
+  },
+  deleteExternalAPI: async (req, res, next) => {
+    try {
+      const response = await service.deleteExternalAPI(req.params.formId, req.params.externalAPIId);
+      res.status(200).json(response);
```
WalterMoar, line 433: For the delete I don't think `service.deleteExternalAPI` returns anything, so there is no `response` to send? I think for deletes we usually return a 204 response. (And again I think this is something that probably needs to be fixed in other places in the controllers!)
```suggestion
      await service.deleteExternalAPI(req.params.formId, req.params.externalAPIId);
      res.status(204).send();
```

app/src/forms/proxy/controller.js
```diff
@@ -0,0 +1,43 @@
...
+      // add path to endpoint url if included in headers...
+      const extUrl = service.createExternalAPIUrl(req.headers, extAPI.endpointUrl);
+      // build list of request headers based on configuration...
+      const extHeaders = service.createExternalAPIHeaders(extAPI, proxyHeaderInfo);
+      let axiosInstance = axios.create({
+        headers: extHeaders,
+      });
+      //for (const [key, value] of Object.entries(extHeaders)) {
+      //  axiosInstance.defaults.headers[key] = value;
+      //}
```
WalterMoar, line 34: I think Code Climate does complain about commented-out code.

app/frontend/src/internationalization/trans/chefs/en/en.json
```diff
@@ -65,11 +65,12 @@
     "update": "Update",
     "cancel": "Cancel",
     "eventSubscription": "Event Subscription",
-    "cdogsTemplate": "CDOGS Template"
+    "cdogsTemplate": "CDOGS Template",
+    "externalAPIs": "External APIs"
```
WalterMoar, line 69: Probably goes without saying, but we'll also need the translations for the other languages.

app/src/forms/form/controller.js
WalterMoar: Oh, and once the return values are figured out we do have OpenAPI docs in `app/src/docs/v1.api-spec.yaml`. It's available as http://localhost:8080/app/api/v1/docs

app/frontend/src/components/forms/manage/ExternalAPIs.vue
```diff
@@ -0,0 +1,603 @@
...
+        { title: 'Name', key: 'name' },
+        { title: 'URL', key: 'endpointUrl' },
+        { title: 'Status', key: 'code' },
+        { title: 'Actions', key: 'actions', align: 'end' },
+      ],
+      externalAPIAlgorithmList: [],
+      externalAPIStatusCodes: [],
+      items: [],
+      techdocsLink:
+        'https://developer.gov.bc.ca/docs/default/component/chefs-techdocs/Capabilities/Functionalities',
```
WalterMoar, line 28: Just so we don't forget, this link will need to be updated once the docs are in the https://github.com/bcgov/common-hosted-form-service-techdocs repo / Backstage.

app/src/forms/form/controller.js
usingtechnology: moved it into sub-folder as per suggestion 1.

app/src/forms/form/service.js
usingtechnology: so removing transaction works in runtime... however, the unit tests will be pretty useless? at least with the transaction we could test for commit being called and rollback.  let me know if you want it removed and the unit tests updated.

app/frontend/src/internationalization/trans/chefs/en/en.json
usingtechnology: I've added a new ticket in the backlog to update i18n.

app/frontend/src/components/forms/manage/ExternalAPIs.vue
usingtechnology: I've added a new ticket in the backlog to get tech docs published and this link updated.

app/src/forms/form/externalApi/routes.js
```diff
@@ -0,0 +1,36 @@
...
+  await controller.listExternalAPIStatusCodes(req, res, next);
+});
+
+routes.put('/:formId/externalAPIs/:externalAPIId', hasFormPermissions([P.FORM_READ, P.FORM_UPDATE]), async (req, res, next) => {
+  await controller.updateExternalAPI(req, res, next);
+});
+
+routes.delete('/:formId/externalAPIs/:externalAPIId', hasFormPermissions([P.FORM_READ, P.FORM_UPDATE]), async (req, res, next) => {
+  await controller.deleteExternalAPI(req, res, next);
+});
```
WalterMoar, line 34: We've previously had problems when there are two IDs in the parameters. I don't think anything is checking that the `externalAPIId` belongs to the `formId`. The exploit would be to set up a form and have FORM_READ and FORM_UPDATE on it, but then update/delete an `externalAPIId` for another form.

As an example, in the `validateParameter` middleware there are some validators like `validateFormVersionId` that both validate that the parameter is a UUID but also that it matches the `formId`.

usingtechnology: yes, relying on the validateFormId to do it's job, then the service will do the check for formId + externalApiID to make sure they are paired.

app/src/forms/form/controller.js
WalterMoar: Looks great!

app/src/forms/form/externalApi/routes.js
WalterMoar: I think the idea of the route / controller / service design is to do have request validation in the route (middleware) layer. Can you create another ticket for this?

app/src/forms/form/service.js
WalterMoar: I think it would be better to remove the transactions (unless needed for multiple database changes) and have simple tests, unless we gain something in the code by having the transaction.

usingtechnology: sounds good. will remove where it is a single table. we may lose a couple of unit tests (I'm not too fond of tests that are 100% mocked). off for a couple of days but will pick this up ASAP.

app/src/forms/proxy/service.js
```diff
@@ -0,0 +1,118 @@
...
+  } else {
+    throw new Error('Headers or Header Name not provided.');
+  }
+};
+
+const trimLeadingSlashes = (str) => str.replace(/^\/+|\$/g, '');
+const trimTrailingSlashes = (str) => str.replace(/\/+$/g, '');
+
+const service = {
+  generateProxyHeaders: async (payload, currentUser, token) => {
```
WalterMoar, line 20: I think that there is a replay problem here, as the token has no expiry. If someone was to capture a token it could be reused forever and have the victim's authorizations.

app/src/forms/proxy/service.js
```diff
@@ -0,0 +1,118 @@
...
+      versionId: payload['versionId'],
+      submissionId: payload['submissionId'],
+      userId: currentUser.idpUserId,
+      username: currentUser.username,
+      firstName: currentUser.firstName,
+      lastName: currentUser.lastName,
+      fullName: currentUser.fullName,
+      email: currentUser.email,
+      idp: currentUser.idp,
+      token: token,
```
WalterMoar, line 36: The `token` is always being included, even when the external API is not allowed to receive the token. I realize that it would be more effort to make the check, but I think it would be better to only include the token when it will be used.

app/src/forms/form/externalApi/routes.js
usingtechnology: Sure. Will add a new middleware for external id

app/src/forms/proxy/service.js
usingtechnology: We are only passing the token to external apis that we trust and authorized. Tokens don’t have expiry dates, you check the issued date and decide what time frame you accept.

usingtechnology: I guess if you’re talking about the proxy headers we generate only the chefs backend understands how to decrypt that data. And the front end stores it for that page only. So it’s ephemeral. We could add an issued time and reject if it’s older than whatever our timeframe is. Of course we then limit the use cases for external api to pretty much form load time only.

app/src/forms/proxy/service.js
usingtechnology: Sure. Makes sense

app/src/components/encryptionService.js
```diff
@@ -0,0 +1,166 @@
+const config = require('config');
+const crypto = require('crypto');
+
+const SERVICE = 'EncryptionService';
+
```
WalterMoar, line 5: Some of the code in this file appears to be taken from a [gist](https://gist.github.com/AndiDittrich/4629e7db04819244e843), and one concern is that if a vulnerability in the gist code is found, then we would be unaware of problem unless we periodically check the gist for updates. Plus how do we know that the code is correct? I couldn't find a "standard" package offered by some big reliable company that we could use, but what about [cryptr](https://www.npmjs.com/package/cryptr)? It seems to be maintained and widely used (for better or worse!).

app/src/forms/common/models/tables/externalAPI.js
```diff
@@ -0,0 +1,107 @@
...
+      },
+      findByFormIdAndName(query, formId, name) {
+        if (name !== undefined && formId !== undefined) {
+          query.where('name', name).where('formId', formId);
+        }
+      },
+    };
+  }
+
+  async $beforeInsert(context) {
```
WalterMoar, line 48: I'm concerned about encrypting data and then storing it in the database. While this does add some security, such as if someone was able to steal one of the backups, it only adds security if the person is unable to steal the secret key from OpenShift - but the person would still have all the other data in the database. I think that if we have concerns about this sort of attack then we should be looking at encrypting the entire database, and at a much wider scale.

However, if we decide to do this database encryption, then we need to have an easy way of rotating the key, for example: on a regular basis to mitigate problems with the use of AES256GCM using random IVs, after a contractor leaves, or if it accidentally gets exposed.

app/src/forms/form/service.js
WalterMoar: Agreed, the unit tests are usually useless for one-line service functions.

app/src/forms/proxy/service.js
WalterMoar: Yes, sorry, too many "tokens" - I was meaning the proxy headers. Even though only the backend can decrypt it, the backend has no idea if the proxy headers object is recent or years old.

I wonder if we can just add a timestamp to the proxy headers object and let the third-party API decide what "too old" means to them.

app/frontend/src/internationalization/trans/chefs/en/en.json
```diff
@@ -79,7 +80,46 @@
...
+    "formApiKeyHeader": "API Key Header Name",
+    "formApiKey": "API Key Value",
+    "formSendUserToken": "Send User Token",
+    "formUserTokenHeader": "User Token Header Name",
+    "formUserTokenBearer": "User Token as Bearer Token",
+    "formSendUserInfo": "Send User Information",
+    "formUserInfoEncrypted": "Encrypt User Information",
+    "formUserInfoHeader": "Encrypt User Information Header Name",
+    "formUserInfoEncryptionKey": "User Information Encryption Key",
+    "formUserInfoEncryptionAlgo": "User Information Encryption Alogrithm",
```
WalterMoar, line 116: Small typo here:
```suggestion
    "formUserInfoEncryptionAlgo": "User Information Encryption Algorithm",
```

docs/chefs-external-api-configuration.md
```diff
@@ -0,0 +1,122 @@
...
+| email        | X-CHEFS-USER-EMAIL        | user's email (ex. alex.smitty@gov.bc.ca)                           |
+| idp          | X-CHEFS-USER-IDP          | the Identity Provider code (ex. 'idir')                            |
+
+**Note** - Although the current user (the form submitter) token is available, there is no guarantee that it has not timed out. It will be stored upon entry/load of the form and is not refreshed.
+
+#### User Info Encryption
+
+The Form Designer will provide the encryption key for the selected encryption algorithm. This algorithm and key will encrypt the User Info into an encrypted value and passed in a single header (with header name = `userInfoHeader` value).
+
+The receiving api will have the same encryption key and algorithm and will decrypt on their end. They can then parse the decrypted user info object.
```
WalterMoar, line 90: Assuming that nobody is using (non-TLS) HTTP, is there a use case for this extra layer of encryption? Can't the third-party API encrypt how they want when they receive it?

usingtechnology: A bcgov api that would care about form id and user info and not have TLS is not likely to exist, true. But it’s entirely optional and costs us nothing to allow it. When messaging/events come online they will be using the same encryption methods at which point the consumer may find it easier to handle all data from CHEFS as encrypted. 

If you want to remove this and deal with it later (message queue) that’s fine and makes sense. Simple to remove this option.

app/frontend/src/components/designer/FormDesigner.vue
jasonchung1871, line 1: Do the proxy headers need to be set in the form designer? Are the external API called from within this view? It's no problem if it is but I think some of the formio javascript functionality doesn't work in the form designer.

docs/chefs-external-api-configuration.md
usingtechnology: Yes it gets called on the designer. It’s actually kind of annoying. It’ll call once you put a url in. We will have to recommend setting all the other values first and set the url last. Not a huge deal but the proxy errors out without expected headers.

app/src/forms/proxy/service.js
usingtechnology: interesting... i like that - add a timestamp to know when this user info was gathered. great idea! will put that in.

app/src/forms/proxy/service.js
usingtechnology: after thinking about it, we need to include all the information in this call. this is called on form load and we don't know what the API config is... (and the form could use multiple configs with different options).

can't implement this change.

app/src/forms/admin/controller.js
```diff
@@ -87,6 +87,34 @@ module.exports = {
     }
   },
 
+  //
+  // APIs
+  //
```
WalterMoar, line 92: The backend test coverage is looking great, but it looks like the additions to this file got missed.

app/src/components/encryptionService.js
usingtechnology: the `crypto` lib itself is standard and that's how you use it. if you look at `cryptr` it's basically the same code. So up to you if you want the lib included (it won't introduce a bunch of other dependencies so that's good).

app/src/forms/proxy/controller.js
```diff
@@ -0,0 +1,40 @@
...
+      const extUrl = service.createExternalAPIUrl(req.headers, extAPI.endpointUrl);
+      // build list of request headers based on configuration...
+      const extHeaders = service.createExternalAPIHeaders(extAPI, proxyHeaderInfo);
+      let axiosInstance = axios.create({
+        headers: extHeaders,
+      });
+      // call the external api
+      const { data } = await axiosInstance.get(extUrl);
+      // if all good return data
+      res.status(200).json(data);
```
WalterMoar, line 35: Does this handle errors from the external API?

usingtechnology: yes, the axiosInstance will throw the error into the handler and onto next.

app/src/forms/proxy/service.js
usingtechnology: Ok, covered off a few things related to this.

So adding headers with values straight from the token:

- `X-CHEFS-TOKEN-IAT` : token `iat` - token issued 
-` X-CHEFS-TOKEN-EXP` : `exp` - token expiry, between expiry and issued at the external API can determine their own valid timeframe
- `X-CHEFS-TOKEN-SUB` : `sub` - @robinwn wanted this so including it, subject from IDP provider

I like your idea of including the timestamp and leaving it to external API to do as they see fit.

.devcontainer/chefs_local/test.json
```diff
@@ -56,6 +56,10 @@
         "windowMs": "900000",
         "max": "100"
       }
+    },
+    "encryption": {
+      "proxy": "5fb2054478353fd8d514056d1745b3a9eef066deadda4b90967af7ca65ce6505",
+      "db": "055cc521987c070c72880d988fd40233fe86efe64e99960fd93bb87145beb6c9"
```
WalterMoar, line 62: I think this database encryption setting should be removed, yes?

app/src/components/encryptionService.js
```diff
@@ -0,0 +1,119 @@
+const config = require('config');
+const Cryptr = require('cryptr');
+
+const SERVICE = 'EncryptionService';
+
+const ENCRYPTION_KEYS = {
+  PROXY: 'proxy',
+  DATABASE: 'db',
```
WalterMoar, line 8: I think this database encryption setting should be removed, yes?

openshift/app.dc.yaml
```diff
@@ -234,6 +234,16 @@ objects:
                     secretKeyRef:
                       key: password
                       name: "${APP_NAME}-objectstorage-secret"
+                - name: SERVER_ENCRYPTION_DB
+                  valueFrom:
+                    secretKeyRef:
+                      key: db
+                      name: "${APP_NAME}-encryption-keys"
```
WalterMoar, line 241: Here's another place that the db encryption stuff wasn't removed. It's also still in the secrets in the dev environment, so it should be removed from there to not cause confusion later on.

docs/chefs-external-api-configuration.md
```diff
@@ -0,0 +1,105 @@
...
+
+A proof of concept was implemented allowing the Form Designer to place the current user's (the Form Submitter) token in the request headers. This also allowed a Form Designer to execute Javascript to call any API programmatically. This opens a security backdoor for the Form Designer to surreptitiously access data and execute functions with the current user's token without their (or CHEFS) knowledge or consent.
+
+The solution documented here allows a Form Designer to specify an External API outside of a form component and have that API called with the current user's (and form) context. There will be an approval process so the CHEFS team (and submitters) have an assurance that the External API performs no malicious activities.
+
+To achieve this, Form Designers will configure their External API, get approval from CHEFS and configure their form components to call a CHEFS proxy that forwards the call to the External API.
+
+Any private data required in the configuration (i.e. API Key) will be stored as encrypted data in the CHEFS database. It will not be a part of the form schema.
```
WalterMoar, line 11: This is no longer true. It would be great if you could you do a self-review of this PR, and then we'll look at it again.

usingtechnology: done.

openshift/app.dc.yaml
usingtechnology: done. instructions for secret do not reference `db` so new ones will not have `db` field.

WalterMoar (approved):
I've run out of time to review this PR. I gave the backend a quick look after recent changes, but it's not a thorough review. Jason C is happy with the frontend changes/design/etc, so I guess it's approved.

PR #1372, opened 2024-05-29
Updated usage of vue-i18n

jasonchung1871:
# Description
Any translations are now used by importing the useI18n function and then mapping out the t function and in some cases the locale variable. Removed all instances of `lang` which was stored in the form store.

Something to look into is the test warnings for $vuetify.open not having the open translation.. as well as a better way to handle isRTL however there may not be a better way.

## Types of changes

## Checklist


app/frontend/src/components/base/BaseInternationalization.vue
jasonchung1871, line 1: This is probably the biggest change here. It was not easy to simply change the global value of the locale from within the script but it could be done within the template. However, this required us to use the v-model attribute in the element but we couldn't set that in a menu so I opted to use the v-select. This changed how it looks however, if you look at it, you'll see. It has some transparency and is slightly bigger.

app/frontend/tests/unit/views/Error.spec.js
jasonchung1871, line 1: We don't need to test the translation library.

PR #1378, opened 2024-05-31
feat: FORMS-1279 cdogs file output type

WalterMoar:
# Description

We want to allow the user to select the output file type when they generate a document using CDOGS. The frontend is what is restricting the output file type - it always chooses PDF. The backend, as the frontend is right now, needs no changes. However, the new endpoint that takes a `submissionId` and a `documentTemplateId` is not being used by the frontend, and probably a good thing.
- fixed a bug where the file would never generate correctly
- added the `convertTo` query parameter.

## Types of changes

## Checklist

### DIFF
--- a/app/src/forms/submission/controller.js
+++ b/app/src/forms/submission/controller.js
@@ -137,6 +137,7 @@ module.exports = {
       const template = await formService.documentTemplateRead(req.params.documentTemplateId);
       const fileName = template.filename.substring(0, template.filename.lastIndexOf('.'));
       const fileExtension = template.filename.substring(template.filename.lastIndexOf('.') + 1);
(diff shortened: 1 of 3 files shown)


app/src/forms/submission/controller.js
```diff
@@ -148,12 +149,12 @@ module.exports = {
...
         },
         options: {
-          convertTo: 'pdf',
+          convertTo: convertTo,
           overwrite: true,
           reportName: fileName,
         },
         template: {
-          content: btoa(template.template),
+          content: template.template.toString(),
```
WalterMoar, line 157: One oops of a couple - we're storing the file in the DB base64 encoded. We shouldn't be storing it in the DB anyway, but it's one more thing to keep in mind when we move to the object storage.

PR #1379, opened 2024-06-04
refactor: FORMS-1139 remove unused packages

WalterMoar:
# Description

When working on the Express problem it was noticed that we:
- have a dependency on `body-parser`, which is never used
- have a dependency on `fast-json-patch`, which is never used
- configure `express.urlencode` with an unnecessary default value

## Types of changes

Refactor (non-breaking change to clean up code / dependencies)

## Checklist


app/app.js
```diff
@@ -24,7 +24,7 @@ let probeId;
 const app = express();
 app.use(compression());
 app.use(express.json({ limit: config.get('server.bodyLimit') }));
-app.use(express.urlencoded({ extended: true }));
+app.use(express.urlencoded());
```
WalterMoar, line 27: The default value for `extended` is `true`, so there is no need to set it.

PR #1382, opened 2024-06-05
fix: FORMS-1139 update api-problem

WalterMoar:
# Description

Update the API Problem package. This package is used to produce informative HTTP responses that include a `title`, `detail`, etc.

The major change from version 7 of `api-problem.js` to version 9 is that the `type` field will now reference a page on MDN rather than `httpstatuses.com`, which is now being used to serve ads.

Old v7 responses:
```json
{
  "type": "https://httpstatuses.com/401",
  "title": "Unauthorized",
  "status": 401,
  "detail": "You do not have access to this form."
}
```

New v9 responses:
```json
{
  "type": "https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401",
  "title": "Unauthorized",
  "status": 401,
  "detail": "You do not have access to this form."
}
```

## Checklist

### DIFF
--- a/app/package-lock.json
+++ b/app/package-lock.json
@@ -11,7 +11,7 @@
       "dependencies": {
         "@json2csv/node": "^6.1.3",
         "@json2csv/transforms": "^6.1.3",
-        "api-problem": "^7.0.4",
+        "api-problem": "^9.0.2",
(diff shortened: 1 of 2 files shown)


WalterMoar:
image

PR #1392, opened 2024-06-17
fix: FORMS-883 help Excel with BCSans CSVs

WalterMoar:
# Description

It was reported that exporting submissions containing Indigenous language characters is not working correctly for CSV exports. The export is fine for JSON, and the CSV when opened in Notepad contains the correct data. The problem is when loading the CSV export into Excel, then the data becomes garbled.

As an example:
- The text `Nəxʷsƛ̕ay̓əmúcən` appears in Excel as `NÉ™xÊ·sÆ›Ì•ayÌ"É™mÃºcÉ™n`
- The text `nk̓ iʔ snm̓am̓áyaʔtn iʔ k̓l sqilxʷtət` appears in Excel as `nkÌ" iÊ" snmÌ"amÌ"aÌyaÊ"tn iÊ" kÌ"l sqilxÊ·tÉ™t`

There is an excellent write-up on the problem, including proposed solutions, in: https://github.com/bcgov/inclusive-names-service/blob/main/docs/file_formats/csv_excel/csv.md

Solution: add a Byte Order Mark (BOM) to the exports.

Risk: this is a change to the CSV output, so there is a possibility that it could affect some users. Without a use-case of how this could be a problem, it’s hard to determine if this risk is legitimate and/or what the impact would be. It would be good to test it with non-Windows applications.

## Types of changes

fix (a bug fix)

## Checklist

## Further comments

This chunk of code isn't covered by the existing tests, so I just left things as-is.

### DIFF
--- a/app/frontend/src/components/forms/ExportSubmissions.vue
+++ b/app/frontend/src/components/forms/ExportSubmissions.vue
@@ -197,7 +197,10 @@ export default {
         );
 
         if (response && response.data && !emailExport) {
-          const blob = new Blob([response.data], {
+          // Create a UTF-8 Byte Order Mark to handle BC Sans in Excel.
+          const BOM = new Uint8Array([0xef, 0xbb, 0xbf]);
+
+          const blob = new Blob([BOM, response.data], {
             type: response.headers['content-type'],
           });
           const url = window.URL.createObjectURL(blob);


WalterMoar:
Excel file in Windows 10:
image

WalterMoar:
LibreOffice Calc on Ubuntu 24.04:
image

PR #1404, opened 2024-07-01
Enhanced error handling and warning logging for Proxy/External API.

usingtechnology:
# Description
Calls to the CHEFS Proxy for External APIs are made during design of the form and configuration of the form components. Without full configuration (expected headers, a configure External API) the `/api/v1/proxy/external` will throw (and log errors). The log gets polluted with "false" negatives and are not truly indicative of the proxy or the external API failing.

This PR is intended to reduce the number of errors written to the log and enhance the return status codes so the form designer and the CHEFS support team can determine appropriate actions.

https://bcdevex.atlassian.net/browse/FORMS-1353

## Types of changes

Changes are primarily to the proxy controller and service. The service will throw `ProxyServiceError` on known logic errors. This allows the controller to log those errors as warnings (ex. missing an expected header) and return a specific status code (299) that indicates to the designer that there is missing information but will not log an error. 

An error handling enhancement around the actual call to the external API will allow the CHEFS support team and the form designer know when the proxy succeeds but the external API fails. This will return a 502 status with the message from the external API error response.

Documentation update with a status code table and their meanings.

 feat (a new feature)

## Checklist

### DIFF
--- a/app/src/forms/proxy/controller.js
+++ b/app/src/forms/proxy/controller.js
@@ -59,13 +59,13 @@ module.exports = {
         // formio components will call as soon as the URL is entered while designing.
         // calls will fire before the designer has added the headers.
         log.warn(error.message);
-        // send an empty result back with status 299 Miscellaneous Persistent Warning
-        res.status(299).json([]);
+        // send back status 400 Bad Request
+        res.sendStatus(400);
       } else if (error instanceof NotFoundError) {
         // may have created formio component before adding the External API config.
         log.warn('External API configuration does not exist.');
-        // send an empty result back with status 299 Miscellaneous Persistent Warning
-        res.status(299).json([]);
+        // send back status 400 Bad Request
+        res.sendStatus(400);
       } else {
         next(error);
       }
--- a/app/src/forms/proxy/controller.js
+++ b/app/src/forms/proxy/controller.js
@@ -3,6 +3,9 @@ const jwtService = require('../../components/jwtService');
 const axios = require('axios');
 const { ExternalAPIStatuses } = require('../common/constants');
 const Problem = require('api-problem');
+const ProxyServiceError = require('./error');
+const { NotFoundError } = require('objection');
+const log = require('../../components/log')(module.filename);
 
 module.exports = {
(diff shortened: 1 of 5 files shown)


app/src/forms/proxy/controller.js
```diff
@@ -19,22 +22,53 @@ module.exports = {
...
     } catch (error) {
-      next(error);
+      if (error instanceof ProxyServiceError) {
+        // making an assumption that the form component making this call
+        // has not been setup correctly yet.
+        // formio components will call as soon as the URL is entered while designing.
+        // calls will fire before the designer has added the headers.
+        log.warn(error.message);
+        // send an empty result back with status 299 Miscellaneous Persistent Warning
+        res.status(299).json([]);
```
WalterMoar, line 63: Changes look great! Just a couple of questions here:
1. Instead of the `299` status, what happens if something like `400` is returned? Does this cause warnings in the UI (we don't want that) but just wondering why we'd return a `2XX` success response when it is a bad request.
2. Would `res.sendStatus(299);` have the effect you want? The code has a few different ways to return just status codes, and I'm hoping to standardize on one of them (I don't care which one, but if they all do the same thing it's a bit less confusing than asking "what's the difference" all the time). (The `sendStatus` makes for slightly shorter unit tests.)

usingtechnology: 1. that was more for checking logs, filtering on 4xx would include these when we probably don't want them. so up to you - make the call for the status you want returned.
2. yup, `res.sendStatus(299);` will work. i was concerned that not sending a body back would mess up the formio components - seems ok with select/dropdown.

WalterMoar: 1. Let's go with 400, since it is a "Bad Request"? We don't worry toooo much about 400s unless we see a lot unexplained ones coming from our own frontend.
2. `sendStatus` would be great!

usingtechnology: done - 400 and sendStatus.

WalterMoar (approved):
Looks great!

PR #1410, opened 2024-07-02
fix: FORMS-1008 add address lookup url

WalterMoar:
# Description

Fix a problem where the address lookup components are not calling the backend. The URL for the backend route was missing from the environment variables, so added it into the devcontainers launcher.

> Note: This is not a complete fix to get the address components working in localhost, it's just enough to call the backend so that work on other bugs is unblocked. There is a bigger problem with CORS errors to be figured out in FORMS-1364.

## Types of changes

fix (a bug fix)

## Checklist

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -21,11 +21,12 @@
     {
       "cwd": "${workspaceFolder}/app/frontend",
       "env": {
-        "VITE_TITLE": "Common Hosted Forms - Local",
+        "VITE_CHEFS_ADVANCE_GEO_ADDRESS_APIURL": "http://localhost:8080/app/api/v1/bcgeoaddress/address",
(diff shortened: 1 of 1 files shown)


.vscode/launch.json
```diff
@@ -21,11 +21,12 @@
     {
       "cwd": "${workspaceFolder}/app/frontend",
       "env": {
-        "VITE_TITLE": "Common Hosted Forms - Local",
+        "VITE_CHEFS_ADVANCE_GEO_ADDRESS_APIURL": "http://localhost:8080/app/api/v1/bcgeoaddress/address",
```
WalterMoar, line 24: The `VITE_CHEFS_ADVANCE_GEO_ADDRESS_APIURL` is the only addition, the rest is just alphabetizing the variables.

PR #1425, opened 2024-07-04
fix: FORMS-1140 run zap on pr deploys

WalterMoar:
# Description

The OWASP ZAP Scan stopped running when we changed the CI/CD pipeline away from the "feature branch" way of doing things. As part of our STRA SoAR we need to run the scan and handle any problems that are identified.

The scan is set up as a reusable workflow. If everything works OK with the PR scans then it's fairly easy to also scan dev / test after deployments, but perhaps that's not necessary.

> Note: GitHub Actions are tricky to test, but this _should_ work. Perhaps best to not merge it until I'm available to deal with problems?

## Types of changes

fix (a bug fix)

## Checklist


.github/workflows/.deploy.yaml
```diff
@@ -1,4 +1,5 @@
-name: Deploy
+name: Deploy PR
+run-name: Deploy PR-${{ github.event.inputs.pr-number }}
```
WalterMoar, line 2: So instead of the current names:
image

We will have:
image

nimya-aot (approved):
Looks good

jasonchung1871 (approved):

PR #1427, opened 2024-07-04
fix: FORMS-1140 run zap on dev

WalterMoar:
# Description

The OWASP ZAP Scan stopped running when we changed the CI/CD pipeline away from the "feature branch" way of doing things. As part of our STRA SoAR we need to run the scan and handle any problems that are identified.

The scan was previously set up as a reusable workflow and is working for PR scans. This PR is to also scan dev after deployment, which is what we had for the old scans.

> Note: GitHub Actions are tricky to test, but this should work. Perhaps best to not merge it until I'm available to deal with problems?

## Types of changes

fix (a bug fix)

## Checklist

### DIFF
--- a/.github/workflows/.deploy.yaml
+++ b/.github/workflows/.deploy.yaml
@@ -85,7 +85,7 @@ jobs:
           pr_number: ${{ github.event.inputs.pr-number }}
 
   deploy:
-    name: Deploys to selected environment
+    name: Deploy
     environment:
       name: pr
       url: ${{ needs.set-vars.outputs.URL }}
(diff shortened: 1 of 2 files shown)


.github/workflows/.deploy.yaml
```diff
@@ -85,7 +85,7 @@ jobs:
           pr_number: ${{ github.event.inputs.pr-number }}
 
   deploy:
-    name: Deploys to selected environment
+    name: Deploy
```
WalterMoar, line 88: We only deploy to one environment, so simplify this name.

.github/workflows/on_push.yaml
```diff
@@ -21,7 +21,7 @@ jobs:
     timeout-minutes: 10
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
```
WalterMoar, line 24: I'm updating these, although I'm not sure we actually need to check out the repo in this file at all.

PR #1442, opened 2024-07-18
Fix: Update error message for unauthorized form access

vijaivir:
# Description

If an unauthorized user tries to access a form, the error message says that the form submission schedule has expired, which is incorrect. This PR updates that error message and adds a new tooltip to the 'Specific Team Members' radio in the Form Access settings. 

### Tooltip
First-time CHEFS users might need help understanding the 'Specific Team Members' form access setting. Since this is the default setting, I have added a tooltip explaining how it works. 

Screenshot 2024-07-18 at 2 46 16 PM

### Error Message
The `getFormSchema()` function in the FormViewer.vue displayed the submission scheduled expired message for any error. It has now been updated. 

Screenshot 2024-07-18 at 2 46 57 PM

## Types of changes

fix (a bug fix)

## Checklist


app/frontend/src/components/designer/FormViewer.vue
```diff
@@ -394,9 +397,10 @@ export default {
         }
       } catch (error) {
         if (this.authenticated) {
-          this.isFormScheduleExpired = true;
-          this.isLateSubmissionAllowed = false;
-          this.formScheduleExpireMessage = error.message;
```
vijaivir, line 399: I was confused about this error handling. It seems like any error in this `getFormSchema()` function would set `this.isFormScheduleExpired` to true, and I am not sure if there is any case where that variable should be set here. I set the error based on the status code, currently only specifying a message for a 401: Unauthorized error. There is also a generic error message for any other cases. I am still testing this to ensure that the `isFormScheduleExpired` variable does not need to be set here.

jasonchung1871 (approved):

PR #1445, opened 2024-07-29
fix: FORMS-1387 update node version

WalterMoar:
# Description

The Red Hat Advanced Cluster Security (ACS) application has identified the image `node` as having vulnerabilities that are fixable. To satisfy the requirements outlined in the Security Threat and Risk Assessment's (STRA) Statement of Acceptable Risks (SoAR), this image must be updated to resolve fixable vulnerabilities (or mitigated in some other way, if updating the image is not possible).

Current: `18.20.3-alpine3.20`
Latest: `18.20.4-alpine3.20`

Also update the image used for the devcontainers.

## Types of changes

fix (a bug fix)

## Checklist


.tool-versions
```diff
@@ -1 +0,0 @@
-nodejs 18.18.2
```
WalterMoar, line 1: Don't think we need this file - seems to be used by "asdf", which we aren't using. I also didn't update the version number when we moved to node 18.20 - oops - but maybe it's better to remove it.

vijaivir (approved):

PR #1448, opened 2024-07-29
ci: FORMS-1405 update un/deploy actions

WalterMoar:
# Description

Rename the Actions:
- `Close pull request deployment` -> `PR Undeploy` (since the Action doesn't close the PR, it undeploys it)
- `Deploy PR` -> `PR Deploy` (so that the two PR Actions are located next to each other in the list)

Do a little cleanup too:
- No need to specify that the undeploy is in dev
- Add the PR number to the undeploy run name
- Try to get multiple deploys to write to the PR multiple times
- Remove a bunch of logic that we no longer need

## Types of changes

ci (change in continuous integration / deployment)

## Checklist

### DIFF
--- a/.github/workflows/pr_deploy.yaml
+++ b/.github/workflows/pr_deploy.yaml
@@ -9,7 +9,7 @@ on:
     inputs:
       pr-number:
         description: Pull request number
-        required: false
+        required: true
         type: string
 
 concurrency:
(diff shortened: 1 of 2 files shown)


.github/workflows/pr_deploy.yaml
```diff
@@ -118,8 +104,9 @@ jobs:
           ref: ${{ needs.set-vars.outputs.ref }}
       - name: Release Comment on PR
         uses: marocchino/sticky-pull-request-comment@v2
-        if: ${{ github.event.inputs.pr-number }} != '' && success()
+        if: success()
         with:
+          append: true
```
WalterMoar, line 109: I'm hoping this is the magic needed so that when we do a rebase & push and then manually re-deploy, we'll have a way of knowing that the re-deploy wasn't forgotten and we're using the latest main code.

PR #1454, opened 2024-07-31
refactor: FORMS-1407 move orm code to service

WalterMoar:
# Description

A new validator called validateExternalAPIId was added to validateParameter. Instead of putting ORM calls in a service and then calling the service, the validator middleware directly calls the ORM. Ideally only the service layer should be making ORM calls - this sort of abstraction makes it easier to re-use code and reduces interconnectedness of the layers.

Also (and what brought this issue up) add testing of this new middleware. The tests should mimic the other middleware tests in mocking the service layer, and shouldn’t have to deal with mocking the ORM layer. Also re-arrange the file to restore the alphabetization of the functions.

## Types of changes

## Checklist

### DIFF
--- a/app/src/forms/common/middleware/validateParameter.js
+++ b/app/src/forms/common/middleware/validateParameter.js
@@ -1,9 +1,9 @@
 const Problem = require('api-problem');
 const uuid = require('uuid');
 
+const externalApiService = require('../../form/externalApi/service');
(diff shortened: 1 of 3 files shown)


usingtechnology (approved):
Love it, it's the right thing to do!
Thanks for fixing it up :)

PR #1457, opened 2024-08-01
fix: FORMS-1237 validate formSubmissionId parameters

WalterMoar:
# Description

The `GET /app/api/v1/submissions/XXX` routes are good candidates for fixing invalid UUID problems. These errors are probably due to custom API calls, rather than bugs in the CHEFS frontend.
1. Add a new `formSubmissionId` validator to the routes that use it.
1. We currently have “route” tests that are actually only testing the controller - add proper route tests and later the other tests will be used as a basis for the controller tests.

## Types of changes

fix (a bug fix)

## Checklist

### DIFF
--- a/app/tests/unit/forms/auth/middleware/apiAccess.spec.js
+++ b/app/tests/unit/forms/auth/middleware/apiAccess.spec.js
@@ -1,5 +1,5 @@
 const { getMockReq, getMockRes } = require('@jest-mock/express');
-const { v4: uuidv4 } = require('uuid');
+const uuid = require('uuid');
 
 const apiAccess = require('../../../../../src/forms/auth/middleware/apiAccess');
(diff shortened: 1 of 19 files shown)


app/tests/unit/forms/auth/middleware/apiAccess.spec.js
```diff
@@ -1,5 +1,5 @@
 const { getMockReq, getMockRes } = require('@jest-mock/express');
-const { v4: uuidv4 } = require('uuid');
+const uuid = require('uuid');
```
WalterMoar, line 2: For ease of understanding, will standardize on importing libraries and then explicitly using the library name and its method in the call. Will move away from this sort of thing where we make up a new name that is specific to this codebase.

PR #1466, opened 2024-08-12
fix: FORMS-1209 add user id validation

WalterMoar:
# Description

FORMS-1204 was tracking the most common database errors that we have due to bad requests. The `GET /app/api/v1/users/preferences/forms/XXX` endpoint is one place to fix the problem. This error is the most common and is probably due to CHEFS frontend bug(s) rather than third-party calls to the API. See FORMS-996.

- Add `formId` validation to the three routes that use it
- There is one route that takes a `userId` parameter - for the sake of completeness of the tests, while we’re in this file add this as a new validator
- The `/user` routes are very simple in terms of middleware use, so it won’t take too much effort to write unit tests for all the routes. Current coverage is high, but we have “route” tests that are actually only testing the controller.
  - 92.59% Statements 25/27
  - 100% Branches 0/0
  - 80% Functions 8/10
  - 92.59% Lines 25/27

## Types of changes

fix (a bug fix)

> This is a _potentially_ breaking change because the API will now return the correct 400 response for bad UUIDs, rather than 422. Any calling code should hopefully be robust in its error handling and this change will not cause problems.

## Checklist

### DIFF
--- a/app/src/forms/common/middleware/validateParameter.js
+++ b/app/src/forms/common/middleware/validateParameter.js
@@ -192,6 +192,24 @@ const validateFormVersionId = async (req, _res, next, formVersionId) => {
   }
 };
 
+/**
+ * Validates that the :userId route parameter exists and is a UUID.
+ *
+ * @param {*} _req the Express object representing the HTTP request - unused.
+ * @param {*} _res the Express object representing the HTTP response - unused.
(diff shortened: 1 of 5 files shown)


WalterMoar:
Tests for the route are now at 100% coverage.
image

jasonchung1871 (approved):

PR #1475, opened 2024-08-15
Fix: FORMS-1012 make add team member in team management case insensitive

bhuvan-aot:
# Description

Fix for following bug tickets

FORMS-1012 - 
In Team Management page, when add new team member is clicked, a popup appears with the the text field that expects username or email address. 
When selected Business or Basic BCeID, the search for users is case sensitive. These changes will make it case insensitive. 

FORMS-1277 - Copy submission option not visible for submissions other than SUBMITTED state, Made this to be visible for COMPLETED status.

## Types of changes

fix (a bug fix)

## Checklist


docs/chefs-identity-provider-changes.md
```diff
@@ -4,26 +4,27 @@ Within the CHEFs application a user's identity provider determines a lot of thei
 
 A User's Identity Provider (IDP) is who vouches for them. In a simplified manner: they provide a username and password (generally) and an Identity Provder verifies them and they end up with a token. Currently for CHEFs we have 3 Identity Providers: `IDIR`, `BCeID Basic` and `BCeID Business`. `IDIR` is for employees/contractors on the BC Government. In CHEFs, the `IDIR` Identity Provider allows for greater power within CHEFs; as far as the CHEFs application is concerned IDIR is the `primary` Identity Provider.
 
-Previously, all IDP logic was hardcoded within the frontend code and was difficult to change and maintain. 
+Previously, all IDP logic was hardcoded within the frontend code and was difficult to change and maintain.
```
WalterMoar, line 7: FYI: added a ticket to Jira so that we're more consistent with when Prettier is run.
https://bcdevex.atlassian.net/browse/FORMS-1475

app/src/components/idpService.js
```diff
@@ -159,9 +159,10 @@ class IdpService {
             const filterName = f.name;
             const paramName = f.param;
             const value = params[paramName];
-            if ('exact' in f) {
+            if ('exact' in f || 'caseSensitive' in f) {
               const exact = f.exact;
-              q.modify(filterName, value, exact);
+              const caseSensitive = f.caseSensitive;
+              q.modify(filterName, value, exact, caseSensitive);
```
WalterMoar, line 165: I have to admin I'm struggling to understand this whole method, but it seems that you're checking for 'exact' **or** 'caseSensitive', and then setting values for both of them. What if only one is defined? Should there be two if clauses here, one for each value?

WalterMoar: And having said that, Code Climate is giving complaints about the complexity and length of the idpService methods - this change will increase that. I'm hoping we an get better reporting from Code Climate so that this is flagged in the PRs, but just an FYI.

usingtechnology: I had the same thought when I looked at it. We should have defaults `exact = false, caseSensitive = true` and only override them if they've been set. I'm not sure if you'll get the expected query result passing in `undefined` for `caseSensitive`.

jasonchung1871 (approved):

jasonchung1871:
Oops, I just meant to approve the front end side of things!

WalterMoar:
It looks like the unit tests are failing.

bhuvan-aot:
> It looks like the unit tests are failing.

These are fixed now and deployed latest code.

app/src/forms/common/models/tables/user.js
```diff
@@ -40,11 +40,13 @@ class User extends Timestamps(Model) {
...
-      filterUsername(query, value, exact = false) {
+      filterUsername(query, value, exact = false, caseSensitive = true) {
         if (value) {
-          if (exact) query.where('username', value);
+          if (exact && caseSensitive) query.where('username', value);
           // ilike is postgres case insensitive like
-          else query.where('username', 'ilike', `%${value}%`);
+          else if (exact && !caseSensitive) query.where('username', 'ilike', String(value).toLowerCase());
+          else if (!exact && caseSensitive) query.where('username', 'ilike', `%${value}%`);
+          else query.where('username', 'ilike', `%${String(value).toLowerCase()}%`);
```
WalterMoar, line 49: If the comment on line 46 is correct, then the `.toLowerCase()` isn't needed, is it? And if it isn't needed, then don't lines 48 and 49 become the same thing? If so, then is one of them incorrect? (apologies for the quick and vague comment - haven't had time to play around with this).

bhuvan-aot: I think the line 46 talks about `lower` keyword where we put in column name for the database query, value still needs to be applied with `.toLowerCase()`.

WalterMoar, line 49: It looks like the line 46 comment is correct: PostgreSQL has a non-standard extension `ilike`, which is `like` but case-insensitive: https://www.postgresql.org/docs/13/functions-matching.html#FUNCTIONS-LIKE. I confirmed this with SQL queries against the DB, but haven't tried it in knex - hopefully it's the same!

So that would make lines 48 and 49 the same. So I think the problem there is that line 48 should be `like` instead of `ilike`, since it's case sensitive.

```suggestion
          else if (exact && !caseSensitive) query.where('username', 'ilike', value);
          else if (!exact && caseSensitive) query.where('username', 'like', `%${value}%`);
          else query.where('username', 'ilike', `%${value}%`);
```

bhuvan-aot: right, line 48 should be like, fixed it. And `.toLowerCase()` is not needed as well.

WalterMoar (approved):
Looks great!

PR #1493, opened 2024-09-04
Fix: Extend Form Submitter permissions for GET /documentTemplates

vijaivir:
# Description

This fixes an incomplete functionality that prevented form submitters from accessing an uploaded document template for CDOGS print.

In the print options dialog, the GET /documentTemplate route is called which requires the `document_template_read` permission, previously only granted to form owners. The intended functionality is that form submitters can use the template while printing a form or a submission, so I have assigned that permission to the `form_submitter` role. 

Bug reported on Confluence:
https://bcdevex.atlassian.net/wiki/spaces/CCP/pages/1550909456/Permissions+on+CDOGS+Print+Template+feature

Jira ticket:
https://bcdevex.atlassian.net/browse/FORMS-1471

## Types of changes

fix (a bug fix)

## Checklist

### DIFF
--- a/app/src/db/migrations/20240904140843_047-extend-submitter-role.js
+++ b/app/src/db/migrations/20240904140843_047-extend-submitter-role.js
@@ -1,0 +1,34 @@
+const uuid = require('uuid');
+
+const { Permissions, Roles } = require('../../forms/common/constants');
+
+const CREATED_BY = 'migration-047';
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = function(knex) {
+    return Promise.resolve().then(() => {
+        const rolePermission = {
+          id: uuid.v4(),
+          createdBy: CREATED_BY,
+          role: Roles.FORM_SUBMITTER,
+          permission: Permissions.DOCUMENT_TEMPLATE_READ,
+        };
+        return knex('role_permission').insert(rolePermission);
+      });
+};
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.down = function(knex) {
+    return Promise.resolve().then(() =>
+        knex('role_permission')
+            .where({
+                createdBy: CREATED_BY,
+            })
+            .del()
+    );  
+};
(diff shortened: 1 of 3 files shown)


app/src/forms/common/constants.js
```diff
@@ -29,7 +29,7 @@ module.exports = Object.freeze({
     FORM_API_UPDATE: 'form_api_update',
     FORM_DELETE: 'form_delete',
     FORM_READ: 'form_read',
-    FORM_SUBMITTER: ['form_read', 'submission_create'],
+    FORM_SUBMITTER: ['form_read', 'submission_create', 'document_template_read'],
```
vijaivir, line 32: This permission is used for the idp users.

WalterMoar: Just an FYI: This new `FORM_SUBMITTER` will be removed in the future: it is not a permission, it's a role. It's going to be a bit of work to figure out how to best provide these permissions for that role.
https://bcdevex.atlassian.net/browse/FORMS-1452

vijaivir: Yes, I noticed that, and it makes it confusing. It's used to apply a few permissions together in one spot. It should be a speedy fix I think.

app/src/db/migrations/20240904140843_047-extend-submitter-role.js
```diff
@@ -0,0 +1,34 @@
...
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = function(knex) {
+    return Promise.resolve().then(() => {
+        const rolePermission = {
+          id: uuid.v4(),
+          createdBy: CREATED_BY,
+          role: Roles.FORM_SUBMITTER,
+          permission: Permissions.DOCUMENT_TEMPLATE_READ,
```
WalterMoar, line 16: @jasonchung1871 - This is going to make the `/rbac/current` call have a larger body (call is currently 840ms in prod). Any concerns about hitting a limit to how big it can be? Maybe we need to look at FORMS-1000 soon, to drop the load on the database?

jasonchung1871: If we can figure out exactly what roles are needed and when, maybe we can retrieve the roles as we need them.

jasonchung1871 (approved):

PR #1494, opened 2024-09-05
update formiojs

vijaivir:
# Description

This PR updates the `formiojs` library from version `4.14.13` to `4.19.5`. The primary motivator for this upgrade is that  version`4.19` fixes an [issue](https://github.com/formio/formio.js/issues/5229) where a select list component within a data grid component isn't rendered correctly. 

### Before Upgrade

select_before

### After Upgrade

select_after

# Issues

This upgrade also introduces an issue where the HTML component using a `<p>` tag renders twice. For example, using the following HTML code will render the `<p>` content twice. 

```
<p style=" word-break: keep-all; min-width: 80px; text-align: center;">
  <b>Sample<br>Heading</b>
</p>
```

Screenshot 2024-09-05 at 2 37 35 AM
Screenshot 2024-09-05 at 2 38 14 AM

This is likely due to a bug in Formio's HTML rendering logic. A user has raised this with their team, and it's currently being [worked on](https://github.com/formio/formio.js/issues/5775). 

## Solution 1: Use Attributes

The Formio HTML element component is meant to be used as a low-code component that provides configuration for CSS classes and attributes in its settings. Formio recommends using the 'Content' component for more complex HTML in their [documentation](https://help.form.io/userguide/form-building/layout-components#html-element). The first solution is to remove the `<p>` tag from the content field and use the attributes to define the style. 

Screenshot 2024-09-05 at 3 05 18 AM
Screenshot 2024-09-05 at 3 07 33 AM

## Solution 2: Use `<div>`

The second solution would be to use `div` as the HTML tag value instead of the default `p`. This makes it so the content is placed within a `<div>` tag and the content is not duplicated. 

Screenshot 2024-09-05 at 3 15 39 AM

Screenshot 2024-09-05 at 3 15 16 AM

This is a faster solution, but it will remove the default margins applied to `<p>` tags. 

## Further Comments

The PR deployment can be used to test out more complex forms and ensure that the HTML content is rendering correctly. There might be similar solutions, such as using the `<span>` tag, depending on the use case. If the issue with the select list not rendering isn't critical, we can wait for the next version in which Formio fixes the issue, but more testing would need to be done as that would be an upgrade to version `^4.21.2`.

## Types of changes

## Checklist

### DIFF
--- a/app/frontend/package-lock.json
+++ b/app/frontend/package-lock.json
@@ -19,7 +19,7 @@
         "crypto-js": "^4.1.1",
         "fast-json-patch": "^3.1.1",
         "font-awesome": "^4.7.0",
-        "formiojs": "^4.14.13",
+        "formiojs": "^4.19.5",
         "keycloak-js": "^21.1.1",
         "lodash": "^4.17.21",
         "mitt": "^3.0.0",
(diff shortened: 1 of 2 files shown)


WalterMoar:
I think we need to keep in mind that the "double render" bug will affect all users of the HTMLElement. Maybe it would be best to wait for the bug fix from form.io?

PR #1496, opened 2024-09-05
build: FORMS-1301 update aws sdk to v3

WalterMoar:
# Description

On every deployment, and in the unit tests, we get the message:

> NOTE: We are formalizing our plans to enter AWS SDK for JavaScript (v2) into maintenance mode in 2023.
> Please migrate your code to use AWS SDK for JavaScript (v3).
> For more information, check the migration guide at https://a.co/7PzMCcy

Upgrade from the v2 to v3 version of the AWS SDK since the v2 is no longer supported.

### Acceptance Criteria
- We are running a supported version of the AWS SDK
- We no longer have warnings when the pods start up

## Type of Change

## Checklist

## Further comments

We didn't have any tests for the S3 storage, and not really sure how we would test this. Leaving it as-is for now, but in the upcoming test writing will be sure to cover this.

### DIFF
--- a/app/src/forms/file/storage/objectStorageService.js
+++ b/app/src/forms/file/storage/objectStorageService.js
@@ -1,4 +1,4 @@
-const { CopyObjectCommand, DeleteObjectCommand, PutObjectCommand, S3Client, GetObjectCommand } = require('@aws-sdk/client-s3');
+const { CopyObjectCommand, DeleteObjectCommand, GetObjectCommand, PutObjectCommand, S3Client } = require('@aws-sdk/client-s3');
 const config = require('config');
 const fs = require('fs-extra');
 const mime = require('mime-types');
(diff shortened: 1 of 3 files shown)


app/src/forms/file/storage/objectStorageService.js
```diff
@@ -103,19 +108,12 @@ class ObjectStorageService {
...
-              path: data.Key,
-              storage: StorageTypes.OBJECT_STORAGE,
-            });
-          }
-        });
-      });
+      await this._s3.send(new PutObjectCommand(params));
+
+      return {
+        path: params.Key,
```
WalterMoar, line 114: The `PutObjectCommand` doesn't return the `Key` the way `upload` did, but the storage shouldn't be changing the key on us!

jasonchung1871:
I'm not sure what I'm doing wrong, but when I use the file upload component and then make a submission, then try to view the uploaded file by clicking the filename, I get an error 403 Unauthorized.

jasonchung1871 (approved):

matthewhall78:
Does the OCIO Object Storage service we use leverage AWS S3?   I thought it was a service provided on prem by Dell.

WalterMoar:
> Does the OCIO Object Storage service we use leverage AWS S3? I thought it was a service provided on prem by Dell.

Yes indeed - the Object Storage we use is a Dell appliance, but the interface to it is AWS S3 compatible. So any S3 package will work for it. I chose to stick with the AWS package, although we could think about switching to something like the MinIO package (MinIO also being a S3-compatible storage system).

PR #1507, opened 2024-09-25
feat: Added column persistence to MySubmissionsTable

jasonchung1871:
# Description
MySubmissionsTable now persists the column selection for the currently active form in local storage. This adds a new package called @vueuse/core. This allows us to store data in the local session storage. We've replaced filterData with mySubmissionPreferences which is in the form store. mySubmissionPreferences is stored to local storage using vueuse/core.

This was a requested feature. It allows for a similar user experience on the previous submissions page to the form submissions page.

## Type of Change

feat (a new feature)

This is a breaking change because ... We've introduced a new package called @vueuse/core. This may break things on the front end, but should not affect data on the backend.

## Checklist

### DIFF
--- a/app/frontend/package-lock.json
+++ b/app/frontend/package-lock.json
@@ -14,6 +14,7 @@
         "@fortawesome/fontawesome-svg-core": "^6.4.2",
         "@fortawesome/free-solid-svg-icons": "^6.4.2",
         "@fortawesome/vue-fontawesome": "^3.0.3",
+        "@vueuse/core": "^11.1.0",
         "axios": "^1.4.0",
         "bootstrap-scss": "^5.3.1",
         "crypto-js": "^4.1.1",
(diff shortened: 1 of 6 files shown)


WalterMoar:
Works great for one form, but in Chrome (and others?) if I save preferences for a second form then it looks like it overwrites the first preferences. Also, I think that all deployments in dev (standard dev and the PRs) will share the same preferences, yes? Probably not important as long as someone's settings for dev aren't overwritten by PR settings.

jasonchung1871:
> Works great for one form, but in Chrome (and others?) if I save preferences for a second form then it looks like it overwrites the first preferences. Also, I think that all deployments in dev (standard dev and the PRs) will share the same preferences, yes? Probably not important as long as someone's settings for dev aren't overwritten by PR settings (collisions with the form ID are unlikely).

Yeah, I didn't know how we wanted to handle this. I think the maximum storage we can store is 5 MB, so I just set it to replace the existing submission columns. I'm uncertain how this will affect forms with thousands of fields, but we will have to see.

PR #1508, opened 2024-09-26
fix: FORMS-1303 rate limit admin, permission, and role

WalterMoar:
# Description

The CodeQL security scan recommends having rate limiting on all routes that touch the database. This is to prevent denial of service attacks by calling these “resource heavy” routes repeatedly.

This work will be split into a few Pull Requests so that we can monitor behaviour. This PR will apply API rate limiting to the currently non-limited routes in /admin, /permission, and /role.

## Type of Change

fix (a bug fix)

## Checklist

## Further comments

> Note: no tests for this. The end goal is to have the rate limiter applied in the root app.js, and removed from each of these route.js files. No point in writing tests for the route.js files only to turf them later.

### DIFF
--- a/app/src/forms/admin/routes.js
+++ b/app/src/forms/admin/routes.js
@@ -2,9 +2,12 @@ const routes = require('express').Router();
 
 const jwtService = require('../../components/jwtService');
 const currentUser = require('../auth/middleware/userAccess').currentUser;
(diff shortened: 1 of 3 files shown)


WalterMoar:
> Note: to be merged and deployed on 2024-10-07.

PR #1509, opened 2024-10-01
FORMS-1412 Allow designers to add custom metadata to the form.

usingtechnology:
# Description
To support Row level security/filtering and decoupling CHEFS and external systems, allow form designers to add custom metadata for the form. For example, add an external id.

For flexibility, the form metadata is a JSON blob. 

Calls to External APIs will add this JSON as a `utf8`/`base64` encoded string with the header: `X-FORM-METADATA`. Future revisions will support them specifyin their own header name, but keeping simple for now.

## Type of Change

feat (a new feature)

## Checklist

## Further comments

Documentation is in the [techdocs repository](https://github.com/bcgov/common-hosted-form-service-techdocs/blob/8eba4981eb925d8d32e2242957957ccabd267217/docs/Capabilities/Integrations/Form-Metadata.md)


app/src/forms/form/formMetadata/index.js
jasonchung1871, line 1: As far as I know this is the only issue, codewise this all looks great!

usingtechnology: Yeah, it's empty; just to keep a consistent pattern.

abhilash-aot (approved):
looks good for the custom Metadata Update to the form

app/frontend/src/components/designer/settings/FormMetadataSettings.vue
```diff
@@ -0,0 +1,110 @@
+<script setup>
+import { storeToRefs } from 'pinia';
+import { useI18n } from 'vue-i18n';
+import { onMounted, ref } from 'vue';
+import { useFormStore } from '~/store/form';
+
+const { t, locale } = useI18n({ useScope: 'global' });
+const { form, isRTL } = storeToRefs(useFormStore());
+const techdocsLink =
+  'https://developer.gov.bc.ca/docs/default/component/chefs-techdocs/Capabilities/Integrations/Form-Metadata/';
```
WalterMoar, line 10: Looks like a deploy is needed to pick up the change with the URL here.

usingtechnology: deployed.

app/frontend/src/internationalization/trans/chefs/en/en.json
```diff
@@ -199,7 +199,10 @@
     "eventSubscription": "Event Subscription",
     "validEndpointRequired": "Please enter a valid endpoint starting with https://",
     "validBearerTokenRequired": "Enter a valid bearer token example: 89abddfb-2cff-4fda-83e6-13221f0c3d4f",
-    "wideFormLayout": "Wide Form Layout"
+    "wideFormLayout": "Wide Form Layout",
+    "formMetadataTitle": "Form Metadata",
+    "formMetadataMessage": "Define identifiers and labels for this form. Calls to external systems will include this metadata in their payloads.",
```
WalterMoar, line 204: Might this wording cause confusion with the "Labels" in the "Form Profile" section below? And what exactly is an "identifier"? Maybe it could be something like "Define JSON data to be included in calls to external systems"? JSON is maybe a little too technical, but the user doesn't know it needs to be JSON until after they enter non-JSON data and get an error. Maybe these are questions for the UX/UI folks, but I think it would be ideal if the user could figure out whether this is important to them or not (probably not) without having to read the techdocs.

app/frontend/src/components/designer/settings/FormMetadataSettings.vue
```diff
@@ -0,0 +1,110 @@
...
+      /* empty */
+    }
+  }
+}
+
+/* c8 ignore start */
+const metadataRules = ref([
+  (v) => {
+    try {
+      JSON.parse(v);
```
WalterMoar, line 41: Don't know why, but this is passing validation:
image

WalterMoar: (that is, until the form is saved, and then a generic error is shown)

app/frontend/src/internationalization/trans/chefs/en/en.json
```diff
@@ -199,7 +199,10 @@
     "eventSubscription": "Event Subscription",
     "validEndpointRequired": "Please enter a valid endpoint starting with https://",
     "validBearerTokenRequired": "Enter a valid bearer token example: 89abddfb-2cff-4fda-83e6-13221f0c3d4f",
-    "wideFormLayout": "Wide Form Layout"
+    "wideFormLayout": "Wide Form Layout",
+    "formMetadataTitle": "Form Metadata",
```
WalterMoar, line 203: For accessibility, we do need translations for all languages for any added text.

app/frontend/src/components/designer/settings/FormMetadataSettings.vue
WalterMoar: Can this backslash be removed?
image
It's probably a Backstage-ism, but we might have people trying:
image

usingtechnology: Yup, can remove it.

app/src/db/migrations/20240925201114_049-form-metadata.js
```diff
@@ -0,0 +1,27 @@
...
+ * @returns { Promise<void> }
+ */
+
+exports.up = function (knex) {
+  return Promise.resolve().then(() =>
+    knex.schema.createTable('form_metadata', (table) => {
+      table.uuid('id').primary();
+      table.uuid('formId').references('id').inTable('form').notNullable().index();
+      table.string('headerName', 255).notNullable();
+      table.string('attributeName', 255).notNullable();
```
WalterMoar, line 14: It seems like these aren't used by the frontend. How soon will the frontend be changed to allow the user to set them? We don't usually add things like this until they're needed, to keep the size of PRs down and so that they don't have to be removed down the line if setting them is never implemented.

app/src/forms/form/formMetadata/routes.js
```diff
@@ -0,0 +1,30 @@
+const routes = require('express').Router();
```
WalterMoar, line 1: We do need OpenAPI docs for all new routes - they're in `app/src/docs/v1.api-spec.yaml`. I think maybe the form object in the docs will need to be updated too?

app/src/db/migrations/20240925201114_049-form-metadata.js
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+
+exports.up = function (knex) {
+  return Promise.resolve().then(() =>
+    knex.schema.createTable('form_metadata', (table) => {
+      table.uuid('id').primary();
+      table.uuid('formId').references('id').inTable('form').notNullable().index();
```
WalterMoar, line 12: Will we ever allow multiple metadata rows for a single formId? The concern here would be that a bug in the frontend creates multiples but it's non-deterministic which one is retrieved.

usingtechnology: no, one per form id.

app/src/forms/form/formMetadata/routes.js
```diff
@@ -0,0 +1,30 @@
...
+const P = require('../../common/constants').Permissions;
+
+const controller = require('./controller');
+
+routes.use(currentUser);
+routes.use(rateLimiter);
+
+routes.param('formId', validateParameter.validateFormId);
+
+routes.post('/:formId/formMetadata', hasFormPermissions([P.FORM_READ, P.FORM_UPDATE]), async (req, res, next) => {
```
WalterMoar, line 14: The techdocs say that "Custom Form Metadata allows the owner/designer..." but I think `FORM_UPDATE` is only for the OWNER role. I'm thinking the permissions here are correct, but it's the techdocs that need to change to remove "designer"? Just want to make sure that the right people have access.

app/src/forms/form/formMetadata/service.js
```diff
@@ -0,0 +1,104 @@
...
+  update: async (formId, data, currentUser) => {
+    return service.upsert(formId, data, currentUser);
+  },
+
+  delete: async (formId) => {
+    const existing = await FormMetadata.query().modify('filterFormId', formId).first().throwIfNotFound();
+    await FormMetadata.query().deleteById(existing.id);
+  },
+
+  upsert: async (formId, data, currentUser, transaction) => {
```
WalterMoar, line 74: We don't have Sonar set up yet, but I think this function might fail the cognitive complexity checks in Code Climate.

I think that our usual pattern is that `service.create` does an SQL `insert` and `service.update` does a SQL `modify`. With this function, though, `create` does an `insert` OR a `modify`, and update also does an `insert` OR a `modify`. So POST and PUT are the same thing, yes? While it's not technically wrong for POST to update or PUT to create, it's an inconsistency in the code that would need to be changed later.

app/frontend/src/internationalization/trans/chefs/en/en.json
usingtechnology: added a ticket, will address this.

app/src/forms/form/formMetadata/service.js
usingtechnology: I will be removing all unused code, which will only leave the upsert function. i can refactor it to reduce complexity.

app/src/forms/form/formMetadata/routes.js
usingtechnology: changed to owner

app/src/forms/form/formMetadata/service.js
usingtechnology: refactored as much as possible. there are no create or update endpoints. should minimize confusion. in order to ensure a single record per form we encapsulate the logic in the upsert method.

app/src/db/migrations/20240925201114_049-form-metadata.js
usingtechnology: removed

app/frontend/src/internationalization/trans/chefs/en/en.json
usingtechnology: changed the wording and added i18n. yes, i think the UX team should be reviewing the text but it shouldn't be a roadblock. that stuff can get changed at any time.

app/frontend/src/internationalization/trans/chefs/en/en.json
usingtechnology: done for all languages.

WalterMoar (approved):

PR #1515, opened 2024-10-10
feat/fix: FORMS-1483 queued file handling

jasonchung1871:
# Description
Adds the functionality so that files are not uploaded/deleted until the submitter clicks submit or save draft. Users that would delete or upload a file, then navigate away without saving the draft or submitting would have their files either exist in our storage or deleted with no recovery. This mitigates that issue.

Resolves https://bcdevex.atlassian.net/browse/FORMS-1483

When adding a file to the file upload, it will give you a notification that it won't be uploaded until you hit save or save as draft.
image

## Type of Change

feat (a new feature)
fix (a bug fix)

## Checklist


WalterMoar (commented):
I think we may not be able to queue the uploaded files. How does this behave when uploading 100 files that are multi-megabytes each?

What happens if one file upload fails in the middle? Is the current behaviour that the user can delete it and re-upload? What will the behaviour be with this?

I think the file deletes can be queued, as there is not a lot of data being transmitted. But the uploads can be very long on slow connnections.

jasonchung1871:
> I think we may not be able to queue the uploaded files. How does this behave when uploading 100 files that are multi-megabytes each?
> 
> What happens if one file upload fails in the middle? Is the current behaviour that the user can delete it and re-upload? What will the behaviour be with this?
> 
> I think the file deletes can be queued, as there is not a lot of data being transmitted. But the uploads can be very long on slow connnections.

Yes, that would be the behaviour. The formio file component has some a property called dataValue which determines whether a file was successfully uploaded or not. When a file is dragged or selected into the file component, that information is forwarded to the front end where we queue it for upload. When the user submits the form or saves the draft, we synchronously begin uploading the files. If it fails to upload a single file, it will show an error to the user and continue uploading the rest of the files synchronously. If a file succeeds, it gets removed from the queue. If all files upload successfully, then it will continue to submit or save the draft.

Some potential changes:

- Display a message to the user that files are being uploaded, the file upload component might be hidden in a container like tabs.
- Decide if we want to asynchronously upload files, it may be a concern for users with slower internet connections that are uploading multiple large files. It may end up timing out for all the files if they take too long.
- Decide if we want to attempt a re-upload. It would depend on what the issue is though. It could be client sided errors where the user is uploading a document, selects it in the file component, moves the file on their local machine, then submits the form and then it fails to upload because they moved it or deleted it, etc. Attempts to re-upload would just keep failing because it's no longer in the location that the file upload thinks it is.

WalterMoar:
There's a variety of behaviour with files around 25MB. Dev/test/prod have strange behaviour, but sticking to this PR:

If I upload a 25MB + 1B (26,214,401B) file, I get the expected error but I can still submit (it's not required) and the file isn't attached to the submission - maybe this isn't the desired behaviour?
image

---

If I upload a 25MB (26,214,400B) file, I do not get an error:
image
but when I click "Submit" the button continues to spin for 30s after the upload completes. Then it produces some error alerts:
image

---

If I upload a 25MB - 1B (26,214,399B) file, everything works as expected.

---

So perhaps the backend is < 25MB, and the frontend is <= 25MB? This is a super-unlikely case for real files, but maybe we should fix it now to make sure the error handling in the frontend is correct (or that the problem just goes away)?

app/frontend/src/components/designer/FormViewer.vue
```diff
@@ -831,15 +833,25 @@ async function saveDraft() {
...
+async function sendSubmission(isDraft) {
+  let err = await uploadQueuedFiles();
+  if (err) {
+    return;
+  }
+
+  err = await deleteQueuedFiles();
+  if (err) {
+    return;
+  }
```
WalterMoar, line 845: If there are 100 files to upload, and 100 files to delete, and one of the uploads fails, then the delete don't happen, yes? Should we be trying to delete, even if the upload fails? What I mean is something like:
```suggestion
  let uploadError = await uploadQueuedFiles();
  let deleteError = await deleteQueuedFiles();
  if (uploadError || deleteError) {
    return;
  }
```

jasonchung1871: Hmmmm, I suppose file deletions are much more reliable than file uploads, since it'll need to be an error on our end to prevent it from happening. I think your way does look much cleaner though. Either works, my biggest concern would still just be when file uploads fail. I'll update it with your changes!

app/frontend/src/components/designer/FormViewer.vue
```diff
@@ -1148,8 +1162,74 @@ async function getFile(fileId, options = {}) {
...
+      headers: cfg.headers,
+    },
+    onUploaded: cfg.onUploaded,
+    onError: cfg.onError,
+  });
+}
+
+async function uploadQueuedFiles() {
+  let err = false;
+  for (let i = queuedUploadFiles.value.length - 1; i >= 0; i--) {
```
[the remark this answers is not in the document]
jasonchung1871: The reasoning for this was because it was uploading files from the array, and when successful, would remove those files from the array. That way when running the upload file function again, it would just run through the array until it is empty.

app/frontend/src/components/designer/FormViewer.vue
WalterMoar: Looks great!

jasonchung1871:
> > There's a variety of behaviour with files around 25MB. Dev/test/prod have strange behaviour, but sticking to this PR:
> > If I upload a 25MB + 1B (26,214,401B) file, I get the expected error but I can still submit (it's not required) and the file isn't attached to the submission - maybe this isn't the desired behaviour? image
> > If I upload a 25MB (26,214,400B) file, I do not get an error: image but when I click "Submit" the button continues to spin for 30s after the upload completes. Then it produces some error alerts: image
> > If I upload a 25MB - 1B (26,214,399B) file, everything works as expected.
> > So perhaps the backend is < 25MB, and the frontend is <= 25MB? This is a super-unlikely case for real files, but maybe we should fix it now to make sure the error handling in the frontend is correct (or that the problem just goes away)?
> 
> I suspect that this may end up being desired behaviour, unless we intend to enforce that all selected files are valid even for forms that don't have the required set for the file components.
> 
> This may end up being a back end change for the validation. The validation for the file upload component is a built-in method for the component. Their method is much cleaner than handling it ourselves, because it translates the string 1GB into a scalar that can be used for validating the file size.


So, I looked into this again. Both the front end and the back end use the config to set their max file size. The back end uses multer for handling files, and the front end uses the built in formio methods. I don't think we can as easily customize the functionality for multer, so we might need to copy the formio functionality and rewrite it just so we can convert the configs file size into a scalar and then subtract it by 1.

jasonchung1871:
The SonarCloud issues seem to think blocks are duplicated but they're not entirely. Most of the questioned duplication comes from mounting the component, which some are duplicated but some aren't even though it's saying it is. I can see this being resolved by creating a setup function that takes in some props and creates the mount then returns it. However, I think SC will think it's duplication again because I'm just wrapping what I'm doing inside of another function.

app/frontend/src/components/designer/FormViewer.vue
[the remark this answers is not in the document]
jasonchung1871: Replaced the reverse iteration with a map, then I filter and remove the successfully uploaded files.

PR #1516, opened 2024-10-10
fix: FORMS-1524 Updated the minLength for idp

jasonchung1871:
# Description
bceid-business and bceid-basic now have a search minLength of 4 instead of 6.

Resolves: https://bcdevex.atlassian.net/browse/FORMS-1524

This change is required because the minimum length for a username when registering a bceid-basic or bceid-business is 4 characters. This resolves the issue of users not being searchable because they have a username shorter than 6.

## Type of Change

fix (a bug fix)

This is a breaking change because we're updating the database table for the `identity_provider` and changing the minLength entry for bceid-basic and bceid-business rows.

## Checklist


WalterMoar (approved):
Works great!

Although for the PR description, I don't think this is a breaking change. We're not breaking or removing anything that users are relying on. A breaking change would be if we went from 4 to 6 characters and some BCeIDs couldn't be added!

PR #1521, opened 2024-10-21
fix: FORMS-1303 move rate limiting to app.js

WalterMoar:
# Description

The CodeQL security scan recommends having rate limiting on all routes that touch the database. This is to prevent denial of service attacks by calling these “resource heavy” routes repeatedly.

We already had `common/middleware/rateLimiter.js` for the API Key callers, but we need one for all callers. The existing limiter was extended to provide a different (higher) limit for the web application users. The final rate limiter solution is applied to all routes automatically and doesn’t need to be added on a route-by-route basis.

### Acceptance Criteria
- All API routes must use IP-based rate limiting
- The rate limiting must never impact normal web app use
- The CodeQL rate limiting alerts should be automatically resolved as ‘fixed’

## Type of Change

fix (a bug fix)

## Checklist

## Further comments

- It's unknown if this will fix the CodeQL warnings or not. If not, one option would be to back this out and put the rate limiter on all the routes.js files, but that's far from ideal.
- Plan to deploy first thing on 2024-10-23.

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -8,6 +8,7 @@ const querystring = require('querystring');
 const log = require('./src/components/log')(module.filename);
 const httpLogger = require('./src/components/log').httpLogger;
 const middleware = require('./src/forms/common/middleware');
+const rateLimiter = require('./src/forms/common/middleware').apiKeyRateLimiter;
 const v1Router = require('./src/routes/v1');
 
 const DataConnection = require('./src/db/dataConnection');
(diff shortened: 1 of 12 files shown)


usingtechnology (commented):
very slick, nice and easy!
is there a test that proves the rate limiter is called for all endpoints?

WalterMoar:
> very slick, nice and easy! is there a test that proves the rate limiter is called for all endpoints?

One good thing is that now we aren't able to forget the rate limiting for a `routes.js` file (although CodeQL would remind us). The downside to this is that since we only have unit tests, we can't test the rate limiting since it is outside the `routes.js` units. When Bruno/Newman tests are eventually created then we can test the rate limiting for each route.

usingtechnology (approved):
nice work!

PR #1529, opened 2024-11-05
fix: FORMS-1586 scheduler errors

jasonchung1871:
# Description
Forms are closing earlier than they should be when using the scheduling feature. This may be due to the difference between local time and UTC time on the server. The default message for when the submission is not available as the scheduled submission period has expired rather than the custom message is being shown.

This resolves: https://bcdevex.atlassian.net/browse/FORMS-1586

## Type of Change

The biggest change is that we no longer format the date so we can keep it a moment object instead of a string. We also set the closing date to be the end of the date, since we're using a date and not a date time, it was assuming start of the day.

fix (a bug fix)

## Checklist

### DIFF
--- a/app/src/forms/common/utils.js
+++ b/app/src/forms/common/utils.js
@@ -3,6 +3,7 @@ const falsey = require('falsey');
 const moment = require('moment');
 const clone = require('lodash/clone');
 const _ = require('lodash');
+const { ScheduleType } = require('./constants');
 
 const setupMount = (type, app, routes) => {
   const p = `/${type}`;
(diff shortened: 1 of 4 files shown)


app/src/forms/common/utils.js
```diff
@@ -209,109 +210,98 @@ const checkIsFormExpired = (formSchedule = {}) => {
...
-          if (formSchedule.closingMessageEnabled) {
-            if (formSchedule.closingMessage) {
-              result = { ...result, message: formSchedule.closingMessage };
-            } else {
-              result = { ...result, message: 'Something went wrong.' };
-            }
-          } else {
-            result = { ...result, message: 'The form submission period has expired.' };
-          }
+  if (formSchedule && formSchedule.enabled && formSchedule.openSubmissionDateTime) {
```
jasonchung1871, line 213: Combined the conditionals because it didn't need to be separated.

app/src/forms/common/utils.js
```diff
@@ -209,109 +210,98 @@ const checkIsFormExpired = (formSchedule = {}) => {
...
+    // The start date is the date that the form should be scheduled to be open to allow submissions
+    let startDate = moment(formSchedule.openSubmissionDateTime).startOf('day');
```
jasonchung1871, line 215: Instead of formatting the open date to just the date the user picked which didn't even include the time, we set it to the date the user picked at the start of the day.

app/src/forms/common/utils.js
```diff
@@ -209,109 +210,98 @@ const checkIsFormExpired = (formSchedule = {}) => {
...
+    // The closing date is the date that the form should be scheduled to be closed
+    let closingDate = null;
+    if (formSchedule.scheduleType === ScheduleType.CLOSINGDATE && formSchedule.closeSubmissionDateTime) {
+      closingDate = moment(formSchedule.closeSubmissionDateTime).endOf('day');
```
jasonchung1871, line 219: Instead of formatting the closing date to just the date the user picked which didn't even include the time, we set it to the date the user picked at the end of the day.

app/src/forms/common/utils.js
```diff
@@ -361,12 +351,12 @@ const getSubmissionPeriodDates = (
   let graceDate = null;
 
   calculatedCloseDate.add(keepOpenForTerm, keepOpenForInterval);
-  if (allowLateTerm && allowLateInterval) graceDate = calculatedCloseDate.clone().add(allowLateTerm, allowLateInterval).format('YYYY-MM-DD HH:MM:SS');
```
jasonchung1871, line 364: From here on out, removed a lot of the `.format` function because it would convert the moment object into a string, preventing us from using moment functions until we converted it back into a moment object.

WalterMoar (approved):
Looks great and love the tests. For the backend the scheduler was one of the big gaps in the tests, so this is really going to improve things.

PR #1536, opened 2024-12-04
fix: FORMS-1587 set up crunchydb

WalterMoar:
> Note: this PR contains the README.md at the top, and everything else is copied from https://github.com/bcgov/crunchy-postgres

As part of the database upgrades we will be switching from standalong Patroni to CrunchyDB (aka managed Patroni).

This PR is to keep a copy of the Helm charts in our repo. This is _good_ because it means we will have consistent deployments of CrunchyDB with no surprises if the Helm charts in the external repo change. This is _bad_ because we will periodically need to update these files from the upstream repo, but it should be as simple as dropping the new files overtop and committing the changes.

Other PRs will follow to supply the parameters used for each of our database, changes to the CI/CD pipeline, etc.

# Description

## Type of Change

fix (a bug fix)

## Checklist


jasonchung1871 (approved):
This looks good to me, we may want to configure the schedules at some point to allow for a limit on incremental backups and a time retention for full backups.

WalterMoar:
> This looks good to me, we may want to configure the schedules at some point to allow for a limit on incremental backups and a time retention for full backups.

Yes indeed! This PR is a copy of the repo for creating a CrunchyDB cluster. The PRs that follow will be the settings specific to our clusters.

PR #1542, opened 2024-12-11
fix: FORMS-1661 - Update API Spec for Preferences Parameter Format

bhuvan-aot:
# Description

The preference parameter in the API specification is currently formatted incorrectly as follows 

image
 

This format should be updated to a valid JSON format.

 

{"minDate":"2020-12-17T08:00:00Z","maxDate":"2020-12-18T08:00:00Z","updatedMinDate":"2020-12-17T08:00:00Z","updatedMaxDate":"2020-12-18T08:00:00Z"}

## Type of Change

docs (change to documentation) 

## Checklist

### DIFF
--- a/app/src/docs/v1.api-spec.yaml
+++ b/app/src/docs/v1.api-spec.yaml
@@ -619,7 +619,7 @@ paths:
           name: preference
           schema:
             type: object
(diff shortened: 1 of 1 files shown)


WalterMoar (approved):
Great fix! It doesn't look like this has been deployed, so it can't be checked. But if it looks good in your local environment then go for it!

bhuvan-aot:
> Great fix! It doesn't look like this has been deployed, so it can't be checked. But if it looks good in your local environment then go for it!

not sure if any available pods for deployment, this is how it looks in my local.
image

PR #1543, opened 2024-12-13
FORMS-1563: auto-approve new ext api when same ministry/url already a…

usingtechnology:
…pproved.

# Description
Simplify Approving External APIs.
If a Ministry and an Endpoint URL (everything except query string) have already been approved, then new External API configurations for the same Ministry and Endpoint URL can be auto-approved.

On the Admin side, if you approve a Ministry/Endpoint URL for an External API configuration, then update all similar configurations (same Ministry, same endpoint).

This just makes life easier when the same owner/designer creates numerous forms or numerous External API configurations (or both!) that point at the same base API. This will also speed things up for the form owner/designer as they don't have to seek approval and wait.

## Type of Change

feat (a new feature)

## Checklist


WalterMoar:
It would be great to deploy this, so we can try it out. Now that we're not.... _limited_... by k8s limits, we can have a lot more PRs deployed. This is good but also probably a bad thing, because four of the nine deployed PRs are no longer open - but it would be great to automate the undeployment.

usingtechnology:
> It would be great to deploy this, so we can try it out. Now that we're not.... _limited_... by k8s limits, we can have a lot more PRs deployed. This is good but also probably a bad thing, because four of the nine deployed PRs are no longer open - but it would be great to automate the undeployment.

Removed 4 "dead" deployments with your help, and have deployed this one. Best way to test is 2 forms with different ministries then add some multiple External APIs with the same base URL to each form and do approvals in the Admin/APIs section. All matching External APIs should be approved after first Admin approval. Then back in the form, add another API with the same base url and it should be auto-approved.

app/src/forms/admin/service.js
```diff
@@ -148,14 +149,33 @@ const service = {
...
+  _approveMany: async (id, data, trx) => {
+    // if we are setting to approved, approve all similar endpoints.
+    // same ministry, same base url...
+    if (data.code === ExternalAPIStatuses.APPROVED) {
+      const adminExternalAPI = await AdminExternalAPI.query(trx).findById(id);
+      const delimiter = '?';
+      const baseUrl = data.endpointUrl.split(delimiter)[0];
+      await ExternalAPI.query(trx)
+        .patch({ code: ExternalAPIStatuses.APPROVED })
+        .whereRaw(`"endpointUrl" like '${baseUrl}%' and code = 'SUBMITTED' and "formId" in (select id from form where ministry = '${adminExternalAPI.ministry}')`);
```
WalterMoar, line 176: Is it possible to do this without using raw SQL? Using the objection/knex stuff does allow us to not worry about SQL injection.

app/src/forms/form/externalApi/service.js
```diff
@@ -82,17 +107,30 @@ const service = {
...
-      });
-
-    return ExternalAPI.query().findById(externalAPIId);
+    let trx;
+    try {
+      trx = await ExternalAPI.startTransaction();
+      await ExternalAPI.query(trx)
+        .modify('findByIdAndFormId', externalAPIId, formId)
+        .update({
+          ...data,
```
WalterMoar, line 123: I think that `...data` is setting too much. We might want to be more explicit about which fields are being updated.

app/src/forms/form/externalApi/service.js
```diff
@@ -82,17 +115,38 @@ const service = {
...
     service.checkAllowSendUserToken(data, existing.allowSendUserToken);
-    await ExternalAPI.query()
-      .modify('findByIdAndFormId', externalAPIId, formId)
-      .update({
-        ...data,
+    let trx;
+    try {
+      trx = await ExternalAPI.startTransaction();
+      await ExternalAPI.query(trx).modify('findByIdAndFormId', externalAPIId, formId).update({
+        formId: formId,
```
WalterMoar, line 129: I think that we don't want the user sending in a request with a different `formId`. This would allow setting an external API for someone else's forms.

app/src/forms/form/externalApi/service.js
```diff
@@ -82,17 +115,38 @@ const service = {
...
+        name: data.name,
+        code: data.code,
```
WalterMoar, line 131: I think by allowing the user to set the `code` they could create a request that had a code of `APPROVED`, yes?

usingtechnology: NoOn update, `code` is set to the existing code value and if the URL has changed is set to `SUBMITTED`. It could get changed to `APPROVED` if the ministry/url is pre-approved. The value passed in is discarded.

app/src/forms/form/externalApi/service.js
usingtechnology: externalAPIId and formId are paired. You wouldn't succeed on the 'findByIdAndFormId' if you passed in a different form.

app/src/forms/form/externalApi/service.js
WalterMoar: Oh, right - it's changing the incoming data up above!

app/src/forms/form/externalApi/service.js
usingtechnology: ha, no worries, glad your reviewing it!

WalterMoar:
I don't know if this auto approval is possible. If I was to find out an URL for another ministry's external API, then I could create a form misrepresented as being from that ministry, and with a matching external API. Then I just wait for their external API to get approved and I could then start sending data to their URL from CHEFS.

app/src/forms/admin/service.js
```diff
@@ -148,14 +149,35 @@ const service = {
...
+  _approveMany: async (id, data, trx) => {
+    // if we are setting to approved, approve all similar endpoints.
+    // same ministry, same base url...
+    if (data.code === ExternalAPIStatuses.APPROVED) {
+      const adminExternalAPI = await AdminExternalAPI.query(trx).findById(id);
+      const delimiter = '?';
+      const baseUrl = data.endpointUrl.split(delimiter)[0];
+      await ExternalAPI.query(trx)
+        .patch({ code: ExternalAPIStatuses.APPROVED })
+        .whereRaw(`"formId" in (select id from form where ministry = '${adminExternalAPI.ministry}')`)
```
WalterMoar, line 176: There is an SQL injection possible with this raw query. It's convoluted, but unknown if/how a bigger problem could be made:
1. Create a form "A"
2. Copy your bearer token and curl in a put request for the form data and set the ministry to `' OR 'x'='x`, so that the where clause evaluates to true no matter what the ministry is
3. Create an external API for "A"
4. Create a form "B" in a different ministry and give it the same URL for an external API
5. Approve form A's external API, and form B's external API will be automatically approved.

Hopefully the admin will notice this:
image
but perhaps there are ways to obscure it.

usingtechnology: That would be a bug in form API. And would be a db issue that we have no integrity on that column. So I can "block"  that here but that's not where the bugs are. Even worse, Ministry values are frontend only.  See: #1188 

Adding a bandaid to prevent sql injection in this query, but bugs should be filed and addressed.

usingtechnology:
> I don't know if this auto approval is possible. If I was to find out an URL for another ministry's external API, then I could create a form misrepresented as being from that ministry, and with a matching external API. Then I just wait for their external API to get approved and I could then start sending data to their URL from CHEFS.

External APIs are read-only and should be protected, either with an API Key or by examining the incoming headers (who the user is and from which form) or both.

usingtechnology:
Added "sendApiKey" to the Admin approval screen. Admins can be assured that any other External API configs will not work without providing a key from the API owner. Or they can get assurance that it is a public API with no restrictions.

abhilash-aot (approved):
The changes for auto-approval based on checking the base URL for already approved ministry URLs look good.

PR #1554, opened 2024-12-23
build: FORMS-1679 remove unused components dependencies

WalterMoar:
It’s a best practice to have as few dependencies as possible. The “components” project that houses the form.io components has many packages in its dependencies that don’t appear to be used. Remove whatever is not needed.

> Note: while we can search the code for package use, it’s entirely possible that some packages are needed but not in an obvious way. After this change the app should be monitored for errors.

### Acceptance Criteria
- Unused dependencies are removed
- The app continues to function as before

# Description

## Type of Change

## Checklist

## Further comments

Including the dev dependencies this reduces the package count from 656 to 461. A side effect is that it will probably get rid of some security vulnerabilities picked up by dependabot.

### DIFF
--- a/components/package.json
+++ b/components/package.json
@@ -22,8 +22,6 @@
   "main": "lib/index.js",
   "module": "node",
   "scripts": {
-    "test:coverage": "nyc --reporter=text mocha --reporter spec './{,!(node_modules)/**/}*.spec.js'",
-    "test": "mocha --require ts-node/register --reporter spec './{,!(node_modules)/**/}*.spec.ts'",
     "watch": "tsc -w",
     "webpack:dev": "webpack",
     "webpack:prod": "webpack --config webpack.prod.js",
@@ -35,8 +33,7 @@
     "lint": "tslint -p ."
   },
   "pre-commit": [
-    "lint",
-    "test"
+    "lint"
   ],
   "files": [
     "dist",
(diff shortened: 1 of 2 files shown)


components/package.json
```diff
@@ -22,8 +22,6 @@
   "main": "lib/index.js",
   "module": "node",
   "scripts": {
-    "test:coverage": "nyc --reporter=text mocha --reporter spec './{,!(node_modules)/**/}*.spec.js'",
-    "test": "mocha --require ts-node/register --reporter spec './{,!(node_modules)/**/}*.spec.ts'",
```
WalterMoar, line 26: Since we don't have tests for the components, removing these scripts allows us to further reduce the number of dependencies.

components/package.json
```diff
@@ -49,43 +46,26 @@
...
-    "autocompleter": "^7.0.1",
     "css-loader": "^7.1.2",
     "formiojs": "^4.14.6",
     "leaflet": "^1.9.4",
     "leaflet-draw": "^1.0.4",
     "leaflet-geosearch": "^4.0.0",
     "lodash": "^4.17.21",
-    "native-promise-only": "^0.8.1",
-    "path-browserify": "^1.0.1",
     "webpack": "^5.75.0"
```
WalterMoar, line 55: Is `webpack` a runtime dependency, or should it be in the devDependencies? Likewise with `css-loader`?

jasonchung1871: As far as I know, `webpack` is a bundler so it should be in devDependencies. I believe `css-loader` should also be a devDependency. It's building fine with this package.json but we don't have tests in place to know if anything will truly be broken. The cypress kitchen sink is passing but I'm not sure if that includes all of the new formio components.

WalterMoar: I'll move them into the `devDependencies`.

jasonchung1871 (approved):
We don't have enough tests to verify that nothing will be broken, but it seems to be running fine on a manual build.

components/package.json
```diff
@@ -49,43 +46,26 @@
   "dependencies": {
     "@types/leaflet": "^1.9.12",
     "@types/leaflet-draw": "^1.0.11",
-    "autocompleter": "^7.0.1",
-    "css-loader": "^7.1.2",
```
WalterMoar, line 53: Note: `css-loader` is needed, but moving it to the `devDependencies`.

components/package.json
```diff
@@ -49,43 +46,26 @@
...
     "formiojs": "^4.14.6",
     "leaflet": "^1.9.4",
     "leaflet-draw": "^1.0.4",
     "leaflet-geosearch": "^4.0.0",
-    "lodash": "^4.17.21",
-    "native-promise-only": "^0.8.1",
-    "path-browserify": "^1.0.1",
-    "webpack": "^5.75.0"
```
WalterMoar, line 61: Note: `webpack` is needed, but moving it to the `devDependencies`.

nimya-aot:
I can run cypress tests on this  env https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-1554 if all these changes  deployed there

WalterMoar:
> I can run cypress tests on this env https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-1554 if all these changes deployed there

That would be great! The PR environment has just been updated with moving a couple of packages to the dev dependencies.

nimya-aot:
@WalterMoar form.io components seems to be working . But a few of the tests are failing since my email id is not exist on Manage team section. Can someone add me to the CHEFS keycloak  group on which PR env is running?
image

WalterMoar:
> @WalterMoar form.io components seems to be working . But a few of the tests are failing since my email id is not exist on Manage team section. Can someone add me to the CHEFS keycloak group on which PR env is running? image

I think that error message is saying that you need to log into the PR environment. That will create your user in CHEFS and then it will appear in the search.

nimya-aot:
@WalterMoar Thanks for the info. Now its working good.

PR #1568, opened 2025-01-07
Fix/forms-1654 Print and wide layout icons not visible for public form submissions

bhuvan-aot:
# Description
Print and wide layout option is not available for public forms

For the forms with logged in users this is how the submission view looks.
image

but the public forms do not have these icons.

image

These changes make Print and Wide layout icon available for public form access forms.

## Type of Change
fix (a bug fix)
## Checklist


app/frontend/src/components/designer/FormViewerActions.vue
```diff
@@ -94,7 +98,7 @@ watch(
     class="mt-6 d-flex flex-md-row justify-space-between flex-sm-column-reverse flex-xs-column-reverse gapRow"
     :class="{ 'dir-rtl': isRTL }"
   >
-    <div v-if="formId">
+    <div v-if="formId && !publicForm">
```
WalterMoar, line 101: In the code coverage report it looks like this new condition is not being tested:
image

WalterMoar (changes requested):
Looks great and seems to function as expected, with both owner- and user-uploaded templates. The code coverage report is showing that FormActionViewer needs test(s) but once those are done it should be good to go.

app/frontend/src/components/designer/FormViewerActions.vue
bhuvan-aot: thank you for pointing this, I have added more tests to cover this path.

WalterMoar (approved):

PR #1576, opened 2025-01-11
fix: FORMS-1703 Added Tags to port numbers, removed unused port number

revanth-banala:
# Description
In the devcontainers we now have seven ports, but they are not labeled and it’s not obvious what they’re used for.

It would be better if we labeled these ports so that it was obvious what they were used for. It also appears that the 8081 port isn’t being used.

An example of these labels can be found in portsAttributes settings of this repo:
 [queue-management/.devcontainer/devcontainer.json at main · bcgov/queue-management](https://github.com/bcgov/queue-management/blob/main/.devcontainer/devcontainer.json) 

Acceptance Criteria
All of the ports have labels that make it obvious what they are used for.

## Type of Change

fix (a bug fix)

## Checklist

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -67,6 +67,9 @@
     "4224": {
       "label": "NATS Server 3"
     },
+    "5173": {
+      "label": "CHEFS Frontend"
+    },
     "5432": {
       "label": "PostgreSQL"
     },
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -21,13 +21,12 @@
 
   // Use 'forwardPorts' to make a list of ports inside the container available locally.
   "forwardPorts": [
+    4222, // NATS Server 1
+    4223, // NATS Server 2
+    4224, // NATS Server 3
     5173, // CHEFS Frontend
     5432, // PostgreSQL
-    8080, // CHEFS Backend
-    8081,
-    4222, //nats
-    4223, //nats
-    4224 // nats
+    8080 // CHEFS Backend
   ],
 
   // Use 'postCreateCommand' to run commands after the container is created.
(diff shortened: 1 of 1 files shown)


.devcontainer/devcontainer.json
```diff
@@ -21,13 +21,12 @@
 
   // Use 'forwardPorts' to make a list of ports inside the container available locally.
   "forwardPorts": [
+    4222, // NATS Server 1
+    4223, // NATS Server 2
+    4224, // NATS Server 3
     5173, // CHEFS Frontend
     5432, // PostgreSQL
-    8080, // CHEFS Backend
-    8081,
```
jasonchung1871, line 27: If someone is using the container for keycloak, for whatever reason, that would be port 8081.

usingtechnology: Local Keycloak has been removed so that port can go.

.devcontainer/devcontainer.json
```diff
@@ -57,5 +56,22 @@
...
+    "4222": {
+      "label": "NATS Server 1"
+    },
+    "4223": {
+      "label": "NATS Server 2"
+    },
+    "4224": {
+      "label": "NATS Server 3"
+    },
+    "5432": {
```
usingtechnology, line 70: 5173 - CHEFS Frontend should be included here.

usingtechnology (changes requested):

.devcontainer/devcontainer.json
revanth-banala: Added CHEFS Frontend to the file

WalterMoar (approved):
Looks great, and this will be really helpful to clarify what the port numbers are used for. Thanks to @usingtechnology for catching the missing frontend config.

usingtechnology (approved):

PR #1579, opened 2025-01-16
FORMS-1707: decouple forms from current user, optimize performance

usingtechnology:
Get Current User always returns form data even when not used. Make the fetch of user form data explicit by adding new API to fetch the current user forms.

# Description
See JIRA tickets: [FORMS-1000](https://bcdevex.atlassian.net/browse/FORMS-1000) and [FORMS-1677](https://bcdevex.atlassian.net/browse/FORMS-1677).

We want to eliminate fetching and returning the current user's forms on each call to `getCurrentUser`. The forms are not always needed and the list can be pretty large (when having access to IDIR forms, for instance).

Where we need the forms, we make an explicit call to a new endpoint. Note that the current user form list is different than other form list APIs. This new endpoint is to return the exact same data in the exact same structure as it appeared previously.

## Type of Change

feat (a new feature)

perf (change to improve performance)

This is (potentially) a breaking change because the API has changed. If someone was leveraging `getCurrentUser` and using the forms returned, then they will have to add in the new call to `getCurrentUserForms`.

## Checklist

### DIFF
--- a/app/frontend/src/components/admin/Developer.vue
+++ b/app/frontend/src/components/admin/Developer.vue
@@ -25,7 +25,8 @@ onBeforeMount(async () => {
 async function getUser() {
   try {
     const user = await rbacService.getCurrentUser();
-    apiRes.value = user.data;
+    const forms = await rbacService.getCurrentUserForms();
+    apiRes.value = { ...user.data, forms: forms.data };
   } catch (error) {
     notificationStore.addNotification({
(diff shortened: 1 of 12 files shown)


jasonchung1871 (approved):
Looks good, as far as I know, every use case for getCurrentUser/GetFormsForCurrentUser/getFormPermissionsForUser is working as intended.

PR #1581, opened 2025-01-17
fix: FORMS-1560 updated validation message for revising status recipient emails section

revanth-banala:
# Description
When a user try to search another recipient email rather than default one and if it does not exist, wrong validation message shows up. Needs to modify the message  to “No results found, Add team members on draft/submission manage page” instead of current message “No Form Reviewers found with the search, Add form reviewers on the manage page"
Acceptance Criteria

User can search recipient emails on recipient email section and if its not found, they should able to know where to add team members so that it should show up on search

## Type of Change

fix (a bug fix)

## Checklist


jasonchung1871 (commented):
This is technically complete, but do we require translations for all languages? This might be a question to raise in SCRUM.

abhilash-aot:
> This is technically complete, but do we require translations for all languages? This might be a question to raise in SCRUM.

would be good to add it all the transalation json files now itself so it doesnt need to be revisited again.

WalterMoar (approved):

PR #1587, opened 2025-01-24
fix: FORMS-1776 remove role and permission routes

WalterMoar:
# Description

We have API endpoints that allow the creation and updating of roles and permissions. We do not use these endpoints to change the roles and permissions, though, as those actions are done through the Knex migrations (to maintain consistency between environments).

We should remove these API endpoints so that we no longer need to maintain the code.

Acceptance Criteria
- The `POST` and `PUT` endpoints for permissions and roles are removed
- The corresponding controller, service, and ORM functions are also removed
- The OpenAPI documentation is removed

## Type of Change

fix (a bug fix)

## Checklist

## Further comments

The acceptance criteria listed removing the ORM part of these routes, but it turns out that the database logic is actually in the service.

### DIFF
--- a/app/src/docs/v1.api-spec.yaml
+++ b/app/src/docs/v1.api-spec.yaml
@@ -2112,40 +2112,6 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
-    post:
-      summary: Create a new permission
-      operationId: createPermission
-      tags:
-        - Permission
-      security:
-        - BearerAuth: []
-          OpenID:
-            - admin
-      requestBody:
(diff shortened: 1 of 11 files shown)


usingtechnology (approved):
Nice, that's a good clean up! Hopefully we can catch more of these "dead" APIs.

Are we ignoring duplicated line warnings? It passes the gates so I think it's ok? We can always have a (1/2) sprint later that is dedicated to clearing Sonar warnings... I'll leave it to you for your judgement on this one.

abhilash-aot (approved):

PR #1591, opened 2025-01-28
fix: FORMS-1109 setting length limits on fields for email templates form

revanth-banala:
## Description

Had the following error with the email templates. It should be updated to set length limits on all input fields.

STATEMENT: update "form_email_template" set "id" = $1, "formId" = $2, "type" = $3, "subject" = $4, "title" = $5, "body" = $6, "createdBy" = $7, "createdAt" = $8, "updatedBy" = $9, "updatedAt" = $10 where "id" = $11

ERROR: value too long for type character varying(4096)

Acceptance Criteria
The subject, title, and body fields have length limits 

## Type of Change

fix (a bug fix)

## Checklist


usingtechnology (approved):
Good stuff Revanth!
Just for future fixes, we would also have these checks in the backend - formService.createOrUpdateEmailTemplate and the model FormEmailTemplate.jsonSchema.
The API is only used through the front end, so this is okay, but in other cases, the API can be called without our front end, so frontend protection isn't always enough.

PR #1599, opened 2025-02-04
build: FORMS-1799 node update for security fixes

WalterMoar:
# Description

The Red Hat Advanced Cluster Security (ACS) application has identified the image `node` as having vulnerabilities that are fixable. To satisfy the requirements outlined in the Security Threat and Risk Assessment's (STRA) Statement of Acceptable Risks (SoAR), this image must be updated to resolve fixable vulnerabilities (or mitigated in some other way, if updating the image is not possible).

### Acceptance Criteria
- We are using the most recent version of Node
- Node is updated both for the devcontainer and the deployed runtime

## Type of Change

## Checklist

## Further comments

- Removed setting the VARIANT value so that we have fewer places to update
- Addressed a Sonar complaint about packages not being in alphabetical order
- Addressed a Sonar complaint about `apt-get clean` not being run
- Addressed a Sonar complaint about RUN commands being able to be combined

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,11 +1,21 @@
 ARG VARIANT="20.18.2-bookworm"
 FROM node:${VARIANT}
 
+# Install some extras such as vim for interactive rebases. Also some
+# Cypress prerequisites for running in Debian containers:
+# https://docs.cypress.io/app/get-started/install-cypress#UbuntuDebian
+
 RUN apt-get update && \
     DEBIAN_FRONTEND=noninteractive apt-get install -y \
-        # Cypress prerequisites for running in Debian containers:
-        # https://docs.cypress.io/app/get-started/install-cypress#UbuntuDebian
-        libgtk2.0-0 libgtk-3-0 libgbm-dev libnotify-dev libnss3 libxss1 \
-        libasound2 libxtst6 xauth xvfb \
-        # For interactive git rebases
-        vim
+        libasound2 \
+        libgbm-dev \
+        libgtk-3-0 \
+        libgtk2.0-0 \
+        libnotify-dev \
+        libnss3 \
+        libxss1 \
+        libxtst6 \
+        vim \
+        xauth \
+        xvfb \
+    && apt-get clean
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-ARG VARIANT="20.18.1-bookworm"
+ARG VARIANT="20.18.2-bookworm"
 FROM node:${VARIANT}
 
 RUN apt-get update && \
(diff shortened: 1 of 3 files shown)


.devcontainer/devcontainer.json
```diff
@@ -4,10 +4,7 @@
 
   "build": {
     "dockerfile": "Dockerfile",
-    "context": "..",
-    "args": {
-      "VARIANT": "20.18.1-bookworm"
-    }
+    "context": ".."
```
WalterMoar, line 7: While we could just update the value of this VARIANT argument and leave the Dockerfile as-is, I think it will be less confusing for future updates if we just remove this and update the Dockerfile value.

.devcontainer/Dockerfile
```diff
@@ -1,11 +1,21 @@
...
+        libgbm-dev \
+        libgtk-3-0 \
+        libgtk2.0-0 \
+        libnotify-dev \
+        libnss3 \
+        libxss1 \
+        libxtst6 \
+        vim \
+        xauth \
+        xvfb \
```
WalterMoar, line 20: Sonar wants the new packages to be in alphabetical order. While I'm usually 100% onboard with alphabetizing things, this does make it a bit more effort to check that we have all the Cypress prerequisites (in case those change).

.devcontainer/Dockerfile
```diff
@@ -1,11 +1,21 @@
...
+    && apt-get clean
```
WalterMoar, line 21: Sonar recommended the clean step to help reduce image size.

Dockerfile
```diff
@@ -1,17 +1,22 @@
...
+  && mkdir /.npm \
+  && chgrp -R 0 /.npm \
+  && chmod -R g=u /.npm
+  
 EXPOSE 8000
-# Create the /.npm directory and grant access to group 0 to allow npm v9 to work
-# See: https://docs.openshift.com/container-platform/4.11/openshift_images/create-images.html#use-uid_create-images
-RUN mkdir /.npm
-RUN chgrp -R 0 /.npm && \
-    chmod -R g=u /.npm
```
WalterMoar, line 16: Sonar recommended combining RUN steps so that there are fewer layers in the image.

.devcontainer/devcontainer.json
usingtechnology: Sounds good. Fine with me.

.devcontainer/Dockerfile
usingtechnology: Since this is not a production image we could just ignore it in Sonar. Up to you... but I would be ok with ignoring this file.

jasonchung1871 (approved):
Looks clean!

.devcontainer/Dockerfile
WalterMoar: Maybe it's good to follow the Sonar recommendation - I don't think we'll be checking the Cypress prerequisites that often.

usingtechnology (approved):
LGTM, certainly doesn't hurt to follow some standards!
Caveat, I didn't try it out as I have too many local changes in flight.

PR #1605, opened 2025-02-12
FORMS-1706 Updated vitest dependencies to v2.1.9

revanth-banala:
# Description
Our frontend test suite is out of date and should be updated. The vitest version is 0.32.2, which was released 2023-06-13. An attempt was made to update to the new v3 of vitest but there was a large number of test failures. Instead update to v2, which had a fewer number of failures.

Acceptance Criteria

We are updated to vitest v2.1.9

The tests all pass

## Type of Change

## Checklist


revanth-banala:
I have updated the vitest to the v2.1.9 but 3 tests are failed, so not sure how to fix this I might need someone's help who is reviewing it.

WalterMoar:
For the failure with FormViewer.spec.js, there is an `it` nested within another `it`. This isn't allowed, but I guess `vitest` didn't complain about it in earlier versions. The usual thing is to group tests using `describe` and then the tests themselves use `it` (or `test`).

TL;DR: on line 296 use `describe` instead of `it`

app/frontend/tests/unit/components/designer/FormViewer.spec.js
```diff
@@ -294,7 +294,7 @@ describe('FormViewer.vue', () => {
   });
 
   it('onMounted', () => {
```
WalterMoar, line 296: I think the way this is usually done is that the `describe` elements are on the outside, with a collection of `it` elements on the inside. That way you can have multiple nested levels of `describes`.

WalterMoar (approved):
Looks great!

PR #1616, opened 2025-02-25
build: FORMS-1927 use consistent version of formio.js

WalterMoar:
# Description

We have dependencies on `formio.js` in both the frontend and the components projects. Ideally the version being used for both should be the same. However, the frontend is using 4.17.4 but the components is using 4.14.7. Although these two packages should be compatible, it would be best if they used the same package version.

> Note 1: the frontend has ^4.13.4 in its `package.json`, and it’s using 4.14.7. Update the `package.json` to use 4.14.7 and be strict about it (remove the semver minor version operator ^)

> Note 2: As a side effect of the components update, it will fix the security vulnerability https://github.com/bcgov/common-hosted-form-service/security/dependabot/226

### Acceptance Criteria
- Update the frontend’s package.json to show the actual version of the package being used
- Update the component’s package.json to match the frontend
- Update the component’s package-lock.json to update the package and its dependencies
- Remove the semantic versioning operator ^ for formio.js since we have to be so careful about changes in the library

## Type of Change

## Checklist

### DIFF
--- a/app/frontend/package-lock.json
+++ b/app/frontend/package-lock.json
@@ -20,7 +20,7 @@
         "crypto-js": "^4.1.1",
         "fast-json-patch": "^3.1.1",
         "font-awesome": "^4.7.0",
-        "formiojs": "^4.14.13",
+        "formiojs": "4.17.4",
         "keycloak-js": "^21.1.1",
         "leaflet": "^1.9.4",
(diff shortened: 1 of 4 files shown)


usingtechnology (approved):
I like forcing the exact version 👍 

Created a form with a custom component so seems all good.

PR #1618, opened 2025-02-25
fix: FORMS-1639 fix formscheduler component to open and close forms in user localtime

revanth-banala:
# Description
 Issue with the form scheduler where forms were closing at midnight UTC instead of midnight in the local time zone. As a result, users in different time zones were seeing the "closing message" at the wrong time.
What Happened:
The Unity team tested the form scheduler and found that even though a form’s closing time was set to midnight in the local time zone (e.g., 6 PM PST), the form was being calculated based on UTC. This caused the closing message to appear prematurely. For example, after 6 PM PST, the form shouldn’t show the closing message yet, but it was appearing at 4:25 PM due to the UTC calculation.

What We’ve Done:
We’ve adjusted the form scheduler logic to respect local time when determining the closing time and showing the related messages.
Now, the form’s "closing message" won’t appear until after midnight in the user’s local time, ensuring consistency no matter where the user is located.
What’s Changed:
The closing time is now calculated based on the user’s local time zone instead of UTC.
The "closing message" will only appear after the scheduled midnight local time, rather than based on UTC.
Acceptance Criteria:
The form closes at midnight local time, not UTC.
The "closing message" appears at the correct local time, after midnight

## Type of Change

fix (a bug fix)

## Checklist


usingtechnology:
I think the issue is more significant and more confusing than what is going on here. Sorry @revanth-banala - I think the work you did here is good, but the ticket/issue was not vetted and analyzed correctly. I'd ask that this ticket go into analysis by @WalterMoar first before proceeding.

I'm not sure that the (frontend) data being in UTC was the issue. It's the backend that determines whether to show the closed message and it calculates based on server timezone (pacific). So I think it's the server logic that needs to initialize itself to work with UTC date/times.

I think there are other issues in the backend too... looks like multiple methods calculating whether a form is closed? That is asking for trouble, so analysis on what is using the different methods and if we can consolidate to a single calculation function.

Changing the frontend to move from UTC to local won't fix any of the existing schedules... they will have UTC dates and the server will calculate if it is closed using Pacific time.

Also, if we are going through the effort of changing the Frontend, I don't think it should be doing any calculations - let the server do that. Date/times should be UTC everywhere but can be displayed as local. Perhaps using different controls where they specifically choose a date, a time, and a time zone (default to local). That way it is obvious. They can choose end of day PST which is stored as 7:59 am the following day in UTC. Server does all calculations against UTC. User's browser can render UTC as local timezone.

PR #1619, opened 2025-02-26
Forms 1676 unintended deletion bug

RyanBirtch-aot:
# Description

After deletion of a feature (A point, circle, polygon, or line), subsequent clicks on the features cause them to be deleted also.

## Type of Change

fix (a bug fix)

## Checklist

## Further comments

This bug was caused by a state issue internal to the Leaflet.Draw library while the Map Component is validating that the form data matches what is rendered in the Map Component. This required refactoring how the component is rendering pre-existing data in the case of reviewing a submission or rendering default features.


components/src/components/Map/services/MapService.ts
```diff
@@ -67,7 +78,19 @@ class MapService {
...
+
+      map.on(L.Draw.Event.DELETED, (e: any) => {
+        e.layers.eachLayer((layer) => {
+          const match = this.isDefaultFeature(layer as L.Layer);
+          if (match) {
+            // re-add the feature/layer to the map
+            drawnItems.addLayer(layer);
+            L.popup()
+              .setLatLng(map.getCenter())
+              .setContent('<p>Please do not delete pre-existing features</p>')
```
abhilash-aot, line 90: why is this popup required here , seems like readding the deleted features?

RyanBirtch-aot: Just to remind users to not remove the default values.

abhilash-aot (approved):

PR #1625, opened 2025-02-26
build: FORMS-1870 Updated package vite in Components and frontend projects

revanth-banala:
# Description
Update package vite so that security vulnerabilities are fixed

The GitHub Dependabot process has created an alert for the esbuild dependency. Since esbuild is a sub-dependency of our dependency vite, we should update vite to the latest version. This needs to be done in both the components and frontend projects so that we are using a consistent version of vite.

To satisfy the requirements outlined in the Security Threat and Risk Assessment's (STRA) Statement of Acceptable Risks (SoAR), this vulnerability must be handled by updating the package version (or mitigated in some other way, if updating the package is not possible).
https://github.com/bcgov/common-hosted-form-service/security/dependabot/224
https://github.com/bcgov/common-hosted-form-service/security/dependabot/225 

Acceptance Criteria
We are using the latest version of vite in both the components and the frontend projects

The vite update also results in esbuild being updated to >=0.25.0

## Type of Change

## Checklist


WalterMoar:
Can you update this with the changes in `main`, and then do another PR Deploy? I think it's still using the broken SSO code that was fixed late last week.

WalterMoar (approved):

PR #1626, opened 2025-02-26
Revert "Test: FORMS- 1787 Added cypress test scripts for Geosearch location feature "

nimya-aot:
Reverts bcgov/common-hosted-form-service#1610


nimya-aot:
Not reverting since only one test script is to be reverted

PR #1629, opened 2025-03-07
 test: FORMS-1997 Modify cypress workflow to run scripts from PR branch

nimya-aot:
# Description

Modified cypress workflow to run the cypress test scripts from PR branch
## Type of Change

Modified **cypress.yaml**  and **pr_deploy.yaml** files

## Checklist


WalterMoar (commented):
I think that eventually we'll want developers to be running the Cypress tests locally and updating them to match their frontend changes. So this will cause problems when:
- a developer makes frontend changes in a branch
- the developer adds Cypress changes to the branch to match the frontend changes, so that the Cypress tests will pass locally
- the developer creates a PR and deploys the PR

In this situation the branch code is deployed but the `main` tests are used, and they will fail.

In the case where a PR is created and the tests are changed in `main`, then the PRs should be updated to include those changes (deploying PRs with code that doesn't include the latest changes to `main` can result in bugs sneaking through)

nimya-aot:
> I think that eventually we'll want developers to be running the Cypress tests locally and updating them to match their frontend changes. So this will cause problems when:
> 
> * a developer makes frontend changes in a branch
> * the developer adds Cypress changes to the branch to match the frontend changes, so that the Cypress tests will pass locally
> * the developer creates a PR and deploys the PR
> 
> In this situation the branch code is deployed but the `main` tests are used, and they will fail.
> 
> In the case where a PR is created and the tests are changed in `main`, then the PRs should be updated to include those changes (deploying PRs with code that doesn't include the latest changes to `main` can result in bugs sneaking through)

ok, but usually automation tests runs against main branch/preferred branch. If any change happens on the front end it fails, then we needs to update the cypress test scripts accordingly. For devs to run cypress tests locally I can create a docker file as @usingtechnology suggested earlier and everyone can run cypress tests with their changes.

abhilash-aot (approved):

nimya-aot:
@WalterMoar Modified workflow to grab cypress tests from PR branch

PR #1632, opened 2025-03-12
Forms 1676 unintended deletion bug

RyanBirtch-aot:
# Description

After deletion of a feature (A point, circle, polygon, or line), subsequent clicks on the features cause them to be deleted also.

## Type of Change

fix (a bug fix)

## Checklist

## Further comments

This bug was caused by a state issue internal to the Leaflet.Draw library while the Map Component is validating that the form data matches what is rendered in the Map Component. This required refactoring how the component is rendering pre-existing data in the case of reviewing a submission or rendering default features.


usingtechnology:
Not sure if this is expected behaviour...
If add some features, then decide to delete one, don't hit save but hit another add another feature, the original feature re-appears (effectively cancelling the delete). is that expected? is that what we want? It seems weird that delete has a save and nothing else does... ¯\_(ツ)_/¯

RyanBirtch-aot:
The library we're using for drawing has some quirks for sure, but I think deleting features having an extra "confirmation" step is maybe preferred?

usingtechnology (approved):
Looks fixed to me.
Two thumbs up!

abhilash-aot (approved):
Changes for delete bug and required field looks good

PR #1635, opened 2025-03-18
Refactor/1994 simplify form schedule component

revanth-banala:
# Description
Simplify the form schedule so that it is easier to use and maintain
The form schedule does three things:

- allows setting when a form opens (is available for submissions)

- allows setting when a form closes (no longer is available for submissions)

- allows setting a repetition schedule

The current user interface offers a very complicated way of setting the closing date, when a simpler approach would be better.

Background
The user interface looks like:
The Open submissions box sets the date on which the form starts to allow submission.

The user can choose the first radio button to Keep open until manually unpublished. Another option would be to Schedule a closing date and they then choose the date after which submissions are no longer accepted.

The above are a great and very simple user interface, and it is obvious when the form closes.

The problem is with the third option, Set up a submission period, which allows the user to choose the number of days, weeks, months, quarters, or years that the form is open. The logic needed to perform this calculation is complex and full of inconsistencies and bugs. For example, what does it mean to open on April 5th and close after one quarter? Is the close date June 30th (the end of the quarter) or is it July 5th which is three months after the start date? It would make more sense for the user to explicitly define the close date.

The important thing is that the third option doesn’t add any significant functionality - it is only a complicated way of setting the closing date.

Additionally, there is an option for the form to automatically re-open after a period of time. While the use case for this is unknown, it is seldom used and would be a good thing to remove.

Benefits of this change
The benefits include:

The code to determine if a form is closed is:

100 lines long - it would be much smaller if things were simplified

Has very little test coverage

Has many “code smells” and represents significant technical debt

The complicated schedule is only used by 5 forms, and they are now closed - this schedule can be converted into a close date

The repeat schedule is only used by 4 forms, and they are now closed - this repeated schedule would need to be removed

Acceptance Criteria
- The third option for selecting a closing date is removed from the frontend

- The backend code for handling the third option is removed

- The forms that use the third option are converted into a form with an option 2 closing date

- Full test coverage is completed for the new pared-down code

- Code smells identified by Sonar are fixed

## Type of Change

## Checklist


revanth-banala:
Fixed the form scheduling UI by removing the complex PERIOD option, simplifying to just MANUAL and CLOSINGDATE options. Added a comprehensive test that documents the now-legacy PERIOD functionality to ensure existing forms with this schedule type continue to work correctly. Users can no longer create forms with PERIOD scheduling, but all previously created forms will maintain their current behavior.  Code reduced by over 1000 lines reduces maintenance burden and improves overall performance

PR #1646, opened 2025-03-28
Allow form owners or team managers to restrict submission sharing to form team members only.

timisenco2015:
# Description

CHEFS Application - Sharing submission Potential Risk Assessment
Access to creating and submitting draft forms in CHEFS is controlled by role-based user access (‘Specific Team Members’ access).
image

Sharing submission function is combined with save and edit drafts functions and is used by Submitter to share the draft submissions with their team members for review and editing.
image

Potential Risk
A security risk has been identified in CHEFS where users maybe inadvertently share draft submissions with individuals who do not authorized access to the form (‘Specific Team Members’ access) but have a valid credential and logged in to CHEFS. This poses a risk of unauthorized users being able to view, edit, save, share, and submit draft submissions, potentially leading to data integrity and confidentiality violations.
Confidential
This risk occurs under the following conditions:
•The business process requires users to share draft submissions with their team members for collaboration.
•The user creating the draft submission inadvertently shares it with a user who is not on the specific users list for form but has accessed CHEFS platform (for any other forms outside of the specific business area).
image
•The recipient of the email notification opens the draft submission with the same credentials used to share the draft.
image
Scenarios and Analysis
1.Scenario 1 - Colleagues collaborating:
 Confidential
• Description: User A shares a submission with User B, who has access to the form as a submitter. User B can view, edit, save, share, and submit the draft submission. The audit log shows User A submitted the submission and User B modified it.
• Concerns: No concerns
2. Scenario 2 - Improper share to former colleague:
• Pre-condition: User A and User B both exist on the users list for the form. User B is removed from the users list after a form is submitted and their credentials have not been revoked yet.
• Description: User A shares a submission with User B, who has access to the form as a submitter. User B can view, edit, save, share, and submit the draft submission. The audit log shows User A submitted the submission and User B modified it. User B can still view the submitted form even after leaving the team if their credentials remain unchanged.
User B can view only the submission previously shared with them, no other system data.
• Concerns: Potential risk if User B's credentials are not revoked after leaving the team.
3. Scenario 3 - Improper share to someone outside the team:
• Description: User A inadvertently shares a submission with User B, who does not have access to the form (not in the users list). Two situations:
o User B was not added to the users list yet
o User B was selected by mistake
User B can view, edit, save, share, and submit the draft submission. The audit log shows User A submitted the submission and User B modified it
image
User B can view only the submission shared with them, no other system data.
• Concerns: Potential integrity and confidentiality risk as User B is unauthorized.
Confidential
Applicable to all scenarios: Applying Row Level Security (RLS) to the form minimizes the risk by preventing users the draft submissions are shared with (User B in the scenarios above) from submitting the draft if they are not assigned to the same organization/feature as the user.
Based on the sharing functionality analysis, the probability of selecting an incorrect user is higher with IDIR, which allows wildcard character searches. Business BCeID and Basic BCeID require exact ID or email searches, reducing the likelihood of errors and returning fewer results.
Risk Mitigation
1. Business:
a. Ensure users are trained and understand their responsibility to share drafts only with authorized users and double-check the recipients before sharing.
b. Ensure access to the form/sharing the form is based on either IDIR or Business BCeID.
c. Promptly revoke user credentials (IDIR or Business BCeID) when users leave the organization to prevent unauthorized access.
d. Use exact search by ID or email to minimize errors and ensure accurate recipient selection.
2. System (CHEFS Core Platform):
a. Include confidentiality statement in the email notification
b. Add controls to prevent unauthorized users from opening shared draft submissions (only users on the team management list can open the submission).
c. Implement a more controlled sharing mechanism that restricts the ability to share submissions with unauthorized users.

## Type of Change

feat (a new feature)

## Checklist


usingtechnology:
Taking a look.

app/src/forms/rbac/service.js
```diff
@@ -297,6 +297,22 @@ const service = {
   getIdentityProviders: (params) => {
     return idpService.getIdentityProviders(params.active);
   },
+
```
usingtechnology, line 300: I think you could use `getUserForms` or `getFormUsers` instead. It's bad enough we have 2 identical methods, we don't need to add a third. We don't like logic in controllers, but I think in this case it is OK, basically just checking results and returning a boolean. So re-use one of the rbac service methods and move the results-> boolean logic to the controller. Also, boolean logic should be negative by default and find if it is positive, just ensures if something weird happens that it should not return a false positive (ie if the Array.isArray called failed, we would return true which is bad).

app/src/forms/rbac/routes.js
```diff
@@ -52,4 +52,8 @@ routes.delete('/users', hasFormPermissions([P.TEAM_UPDATE]), hasFormRoles([R.OWN
   await controller.removeMultiUsers(req, res, next);
 });
 
+routes.get('/form/user', hasFormPermissions([P.SUBMISSION_CREATE, P.SUBMISSION_DELETE, P.SUBMISSION_READ, P.SUBMISSION_REVIEW, P.SUBMISSION_REVIEW]), async (req, res, next) => {
```
usingtechnology, line 55: Not sure you want all those permissions... A submitter who is not a reviewer can make this call.
Could go with `hasFormPermissions([P.FORM_READ])`, this function is read-only so I think it's ok just to check they have form access... makes this more easily re-used.

app/src/forms/rbac/controller.js
```diff
@@ -140,4 +140,12 @@ module.exports = {
       next(error);
     }
   },
+  isUserPartOfFormTeams: async (req, res, next) => {
+    try {
+      const response = await service.isUserPartOfFormTeams(req.query);
```
usingtechnology, line 145: see comment for service. I think it's ok if you want to return a boolean to put the logic here, get the results from the service, then do your check for returning true/false. easier to maintain than having such a specific call in the service class.

app/frontend/src/components/forms/RequestReceipt.vue
```diff
@@ -20,29 +20,49 @@ const properties = defineProps({
...
+      if (form.value.enableTeamMemberDraftShare) {
+        const formUsersResponse = await rbacService.isUserAssignedToFormTeams({
+          formId: properties.formId,
+          email: to.value,
+          roles: '*',
+        });
+        if (formUsersResponse && !formUsersResponse.data) {
+          notificationStore.addNotification({
+            ...NotificationTypes.ERROR,
+            text: `You can't share the draft with users who haven't been added to the form. Please contact the admin to add them.`,
```
usingtechnology, line 61: We need the internationalization for all the languages. We expended some effort to get all the languages up to date and don't want to add any tech debt, so we keep it up to date in each PR.
"Draft submissions may only be shared with team members. Contact the form owner to add them."

app/frontend/src/components/designer/settings/FormFunctionalitySettings.vue
```diff
@@ -318,5 +319,20 @@ defineExpose({
...
+      :disabled="!form.enableSubmitterDraft"
+      data-test="enableTeamMemberDraftShare"
+      hide-details="auto"
+      class="my-0"
+    >
+      <template #label>
+        <span
+          :class="{ 'mr-2': isRTL }"
+          :lang="locale"
+          v-html="'Share draft with form team members only'"
```
usingtechnology, line 333: See comment about i18n... need this added in for all languages.

usingtechnology (changes requested):
I like the form setting to enable the "user in team" check, shouldn't impact existing forms. And thanks for adding tests. I like the functionality and seems to work well. 

Just a little tweaking to re-use code where possible (easier for us to maintain) and the internationalization. It's a pain but more painful if we don't do it PR by PR now.

app/src/forms/rbac/routes.js
timisenco2015: fixed

app/src/forms/rbac/controller.js
timisenco2015: fixed

app/src/forms/rbac/service.js
timisenco2015: I am lost here. Please can you give more details?

usingtechnology: So, a couple of ways to do this... since you have already updated the controller to check the array and return a boolean, you no longer need this method in the service. I think you can just have the controller call `service.getFormUsers` and remove this method (`isUserPartOfFormTeams`) from the service. I don't know why we already have 2 identical `UserFormAccess` queries in the service but we shouldn't add a third. I'll add a backlog ticket to refactor out the duplicates (`service.getFormUsers` and `service.getUserForms`).

timisenco2015: Fixed

usingtechnology (approved):
This is great!
Thanks for contributing this.

PR #1649, opened 2025-03-31
restrict submission access for removed form users

timisenco2015:
# Description
When a user is removed from the form's assigned user list via team management, their access to all previously shared 
submissions must be revoked. This ensures that only authorized users retain access to sensitive or restricted submission data
Screenshot 2025-03-30 154815

A submission is shared with user AX by user AY, but AX is not in the team management list above
Screenshot 2025-03-30 155601

AX tries to access the submission, but because AX is not on or has been removed from  the form's users list, user AX gets an error
Screenshot 2025-03-30 155426

## Type of Change

fix (a bug fix)

## Checklist

- [] I have added necessary documentation (if appropriate)

### DIFF
--- a/app/frontend/src/components/forms/submission/ManageSubmissionUsers.vue
+++ b/app/frontend/src/components/forms/submission/ManageSubmissionUsers.vue
@@ -114,7 +114,7 @@ async function addUser() {
     if (formSubmissionUsers.value.some((u) => u.id === id)) {
       notificationStore.addNotification({
         ...NotificationTypes.WARNING,
-        text: t('trans.manageSubmissionUsers.remove', {
+        text: t('trans.manageSubmissionUsers.userExistInListMsg', {
           username: userSearchSelection.value.username,
         }),
       });
--- a/app/src/forms/auth/middleware/userAccess.js
+++ b/app/src/forms/auth/middleware/userAccess.js
@@ -422,7 +422,6 @@ const hasSubmissionPermissions = (permissions) => {
       // access to all submissions for the form.
       if (req.currentUser) {
         const formFromCurrentUser = await _getForm(req.currentUser, submissionForm.form.id, false);
-
         // Do they have the submission permissions requested on this form?
         if (_hasAllPermissions(formFromCurrentUser?.permissions, permissions)) {
           req.formIdWithDeletePermission = submissionForm.form.id;
@@ -430,7 +429,20 @@ const hasSubmissionPermissions = (permissions) => {
 
           return;
         }
+        const users = await rbacService.getFormUsers({ userId: req.currentUser.id, formId: submissionForm.form.id });
+        //Users who have been removed from the form's user list in team management should no longer have access to its submissions."
+        if (Array.isArray(users) && users.length === 1) {
+          const [{ roles }] = users;
+          if (roles && roles.length === 0) {
+            throw new Problem(401, {
+              detail: 'You do not have access to this submission.',
+            });
+          }
+        }
       }
+
+      // If the current user has elevated permissions on the form, they may have
+      // access to all submissions for the form.
 
       // Deleted submissions are only accessible to users with the form
       // permissions above.
--- a/app/src/forms/auth/middleware/userAccess.js
+++ b/app/src/forms/auth/middleware/userAccess.js
@@ -431,6 +431,7 @@ const hasSubmissionPermissions = (permissions) => {
         }
         const users = await rbacService.getFormUsers({ userId: req.currentUser.id, formId: submissionForm.form.id });
         //Users who have been removed from the form's user list in team management should no longer have access to its submissions."
+        //
         if (Array.isArray(users) && users.length === 1) {
           const [{ roles }] = users;
           if (roles && roles.length === 0) {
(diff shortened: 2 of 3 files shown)


npham49:
Hi reviewers, this PR is to address this bug within your Bug Tracker: https://bcdevex.atlassian.net/wiki/spaces/CCP/pages/1583742985/Manage+Team+Members+modal+shows+incorrect+information+after+copying+a+submission

usingtechnology:
taking a look.

usingtechnology (changes requested):
I think what you want to do is trigger a removal of `form_submission_user` records when you delete a `form_role_user` record. You shouldn't need to make changes to the middleware if the records no longer exist.

timisenco2015:
> I think what you want to do is trigger a removal of `form_submission_user` records when you delete a `form_role_user` record. You shouldn't need to make changes to the middleware if the records no longer exist.

After conducting thorough research, we deliberately implemented the necessary changes within the middleware. While an alternative approach—automatically removing form_submission_user records upon deleting a corresponding form_role_user record—was considered, it presents a significant drawback. Specifically, if a user is mistakenly removed from team management, automatically deleting their associated form_submission_user records would result in losing access to related submissions. Even if the user is later reinstated, their access would not be restored, leading to potential data access issues.

usingtechnology:
By not removing the records and relying on middleware, the data does not reflect reality. If someone queried the db it would appear the user does have access but they don't/shouldn't. 

You're adding more code to maintain and data access logic that all developers would need to know about and where to look when the db query shows they have access.

Leaving the data in an invalid state is a bad idea. Accessing that data (securely) is now completely dependent on that single point of entry through that particular piece of middleware.

If someone is removed then they get added back manually to restore their access, it may be a little painful to restore it but they will ask until they get what they need. Leaving the data to indicate they have access when they do not is a data security issue. We don't know how that data will be used or queried (or reported on) in the future.

You have identified a bug and data security risk and the fix is to remove the erroneous data. The middleware addresses a single vector fix but the actual data security risk remains.  

Hope that makes sense.

timisenco2015:
> By not removing the records and relying on middleware, the data does not reflect reality. If someone queried the db it would appear the user does have access but they don't/shouldn't.
> 
> You're adding more code to maintain and data access logic that all developers would need to know about and where to look when the db query shows they have access.
> 
> Leaving the data in an invalid state is a bad idea. Accessing that data (securely) is now completely dependent on that single point of entry through that particular piece of middleware.
> 
> If someone is removed then they get added back manually to restore their access, it may be a little painful to restore it but they will ask until they get what they need. Leaving the data to indicate they have access when they do not is a data security issue. We don't know how that data will be used or queried (or reported on) in the future.
> 
> You have identified a bug and data security risk and the fix is to remove the erroneous data. The middleware addresses a single vector fix but the actual data security risk remains.
> 
> Hope that makes sense.

I completely understand your point; however, it's important to note that this issue existed prior to the creation of this PR. At the moment, my team is aiming to limit the scope of changes and avoid diving too deeply into broader issues

usingtechnology:
For sure the bug existed and very grateful that you and your team have identified it. And I empathize with your immediate goals - restrict access and get the fix to prod ASAP.

However, I have to look at it from us maintaining and supporting it. From a support perspective, someone will raise an issue (probably to their team) that they used to have access to submission X. Someone will open the access screen for the submission and see that user as having permission. Realistically, that will get escalated to us as that would make no sense. Immediately a query would be done, again, it would appear the user does have access. If someone doesn't know about the additional filter in the middleware then it's a walk through the code until that is identified and work through it. It's a lot of work for us and it also looks bad - why does the user appear to have permission on screen but they don't? Whereas, if the initial pop-up of the share permissions doesn't have the user, it is obvious. The person with access may immediately try to add the user, again immediate response saying they are not a member of the team and cannot be added. All of that can happen immediately and obviously without escalating.

I hope this doesn't sound antagonistic, not trying to be. We have a large backlog of things to address and I don't want to add another item to the list (back out these changes, add record removal). It is greatly appreciated that you and your team care about CHEFS and are actively working to make it better, but in this case, it adds more tech debt. It's cool if you want to close the PR and I can add a ticket for this sprint (starts tomorrow) that will get the fix into production.

timisenco2015:
> For sure the bug existed and very grateful that you and your team have identified it. And I empathize with your immediate goals - restrict access and get the fix to prod ASAP.
> 
> However, I have to look at it from us maintaining and supporting it. From a support perspective, someone will raise an issue (probably to their team) that they used to have access to submission X. Someone will open the access screen for the submission and see that user as having permission. Realistically, that will get escalated to us as that would make no sense. Immediately a query would be done, again, it would appear the user does have access. If someone doesn't know about the additional filter in the middleware then it's a walk through the code until that is identified and work through it. It's a lot of work for us and it also looks bad - why does the user appear to have permission on screen but they don't? Whereas, if the initial pop-up of the share permissions doesn't have the user, it is obvious. The person with access may immediately try to add the user, again immediate response saying they are not a member of the team and cannot be added. All of that can happen immediately and obviously without escalating.
> 
> I hope this doesn't sound antagonistic, not trying to be. We have a large backlog of things to address and I don't want to add another item to the list (back out these changes, add record removal). It is greatly appreciated that you and your team care about CHEFS and are actively working to make it better, but in this case, it adds more tech debt. It's cool if you want to close the PR and I can add a ticket for this sprint (starts tomorrow) that will get the fix into production.

Thank you. I completely agree with your proposed solution; it’s the most reliable and future-proof approach. However, since we’re working with a client, we want to be mindful not to introduce any unnecessary complexity on their side. We aim to keep things as straightforward and non-disruptive as possible while still delivering a stable and maintainable solution. I will pass this on to the team.

usingtechnology:
I'm adding a ticket to our backlog with a team discussion, to discuss options, costs/benefits etc. Ideally, we could solve it without an extreme effort and keep everyone happy...

PR #1659, opened 2025-04-22
refactor: FORMS-2218-Extract schedule/reminder date logic to new scheduleservice.js module

revanth-banala:
# Description

Description of ticket:
Using FORMS-2216: Determine what date calculations are used to support the schedule
and FORMS-2217: Determine what data calculations are used to support the reminder service
we can create a service or class that supports all schedule related date handling. The aim is to reduce duplication of code and provide consistent calculations.

Created a dedicated scheduleservice.js to house all date calculation logic for schedules and reminders.
Moved date-related functions from utils.js and reminderservice.js into the new service.
Updated all references across the codebase to import from scheduleservice.js instead.
Added a dedicated unit test suite for the new module.
No frontend code changes.
Will update the techdocs in next ticket about the changes in Form schedule feature.

## Type of Change

## Checklist


app/src/forms/common/scheduleService.js
```diff
@@ -0,0 +1,597 @@
...
+
+/**
+ * Check if a date is in the future
+ * @param {String|Date} date The date to check
+ * @param {String} timezone The timezone to evaluate in
+ * @param {String|Date} referenceDate Optional reference date (defaults to now)
+ * @param {Boolean} ignoreTime Whether to ignore time component (defaults to true)
+ * @returns {Boolean} True if the date is in the future
+ */
+function isDateInFuture(date, timezone = 'America/Vancouver', referenceDate = null, ignoreTime = true) {
```
usingtechnology, line 17: Make 'America/Vancouver' and the date formats ('YYYY-MM-DD') into constants. Maybe in the future we can pull them from configuration... but for now, have them declared once so we can change them easily.

app/src/forms/common/scheduleService.js
```diff
@@ -0,0 +1,597 @@
...
+ * @param {String|Date} date The date to check
+ * @param {String|Date} startDate The start date of the range
+ * @param {String|Date} endDate The end date of the range
+ * @param {String} timezone The timezone to evaluate in
+ * @param {Boolean} ignoreTime Whether to ignore time component (defaults to true)
+ * @returns {Boolean} True if date is within the range (inclusive)
+ */
+function isDateInRange(date, startDate, endDate, timezone = 'America/Vancouver', ignoreTime = true) {
+  if (!date || !startDate) return false;
+  if (!endDate) return true; // No end date means infinite range
```
usingtechnology, line 44: You still need to check date is after startDate when there is no endDate.

app/src/forms/common/scheduleService.js
```diff
@@ -0,0 +1,597 @@
...
+  const defaults = {
+    openSubmissionTime: formSchedule.openSubmissionTime || '00:00',
+    closeSubmissionTime: formSchedule.closeSubmissionTime || (formSchedule.closeSubmissionDateTime ? '23:59' : null),
+    timezone: formSchedule.timezone || 'America/Vancouver',
+  };
+  const schedule = { ...formSchedule, ...defaults };
+
+  // Opening Time
+  const openDateTime = `${schedule.openSubmissionDateTime} ${schedule.openSubmissionTime}`;
+  const openingMoment = moment.tz(openDateTime, 'YYYY-MM-DD HH:mm', schedule.timezone).utc();
```
usingtechnology, line 203: why do we have utc in these calculations but not everywhere? do we need utc is we are setting the timezone?

app/src/forms/common/scheduleService.js
```diff
@@ -0,0 +1,597 @@
...
+}
+
+/**
+ * Determines what type of reminder email should be sent based on current date and schedule
+ * @param {Object} scheduleOrReport The form schedule or period report
+ * @param {Date|String} referenceDate Optional reference date (defaults to now)
+ * @param {Boolean} respectTimeComponent Whether to respect time components in dates (defaults to false)
+ * @returns {String|undefined} Email type constant or undefined if no reminder should be sent
+ */
+function getEmailReminderType(scheduleOrReport, referenceDate = null, respectTimeComponent = false) {
```
usingtechnology, line 444: This should be in the reminderService and use the scheduleService functions for getting period and date calculations. Schedule doesn't need to know about reminders but reminders do need information about schedules.

app/src/forms/common/scheduleService.js
```diff
@@ -0,0 +1,597 @@
...
+  // Get the current period if a schedule was provided
+  const report = scheduleOrReport.scheduleType !== undefined ? getCurrentPeriod(scheduleOrReport, referenceDate, respectTimeComponent) : scheduleOrReport;
+
+  if (!report || !report.dates) return undefined;
+
+  const now = referenceDate || new Date();
```
usingtechnology, line 450: make sure you are doing current or now the same everywhere. i see multiple different ways of setting now/current. should probably make a function you can export from scheduleService and use in reminderService.

app/src/forms/common/scheduleService.js
usingtechnology, line 444: is referenceDate ever passed in? And if so from where and where does that data originate? Would we need to pass in a timezone for it (could be defaulted to America/Vancouver)...?

app/src/forms/common/scheduleService.js
```diff
@@ -0,0 +1,597 @@
...
+
+  return undefined;
+}
+
+/**
+ * Validate a form schedule object
+ * @param {Object} schedule The schedule object to validate
+ * @returns {Object} Validation result {message, status}
+ */
+function validateScheduleObject(schedule = {}) {
```
usingtechnology, line 486: validateScheduleObject, isLateSubmissionConfigValid and isClosingMessageValid don't belong in this file. They can use functions from this file (formatting, date calculations etc). I think these belong in form.service - whereever we are validating the form data.

usingtechnology (changes requested):

usingtechnology (changes requested):
getting close. a few things to change for consistency. also, i don't think we have ever checked to see if the schedule is enabled before sending reminders. add that in there. if the schedule isn't enabled we shouldn't send a reminder.

usingtechnology (approved):
Great work.
Awesome coverage with the unit tests, and it's nice to have all that schedule logic together.

PR #1667, opened 2025-04-29
fix: FORMS-1992-empty-modified-column and modified date format

revanth-banala:
# Description
Observed Bug: 

The Modified column in the submissions list is empty unless the Modified By column is also visible. When the Modified column is made visible, it is in UTC. It should be displayed in the same format as the Submitted Date column.

Reproduction Steps:

In CHEFS (PROD) open the management console for an active form

Navigate to the submissions list.

Click the Select Columns button.

Select "Modified"

Click the Select Columns button.

Select "Modified By"

Observations:

1.) The Modified column in the submissions list is empty when it is first selected.
2.) You need to display the Modified By field at the same time.
3.) When both the Modified column and the Modified By column are displayed simultaneously, then the Modified column displays its values. 
4.) The Modified column displays values in UTC format rather the same format as Submitted Date.

The bug was caused by two issues:
1. A dependency in the data mapping where updatedAt was only populated if updatedBy existed
2. Missing date formatting for the updatedAt field in the template

This PR removes the dependency and adds proper date formatting consistent with the Submitted Date column.

## Type of Change

fix (a bug fix)

## Checklist


usingtechnology (approved):
Nice work @revanth-banala.

PR #1668, opened 2025-04-30
FORMS-2295 - fix translation generator

usingtechnology:
# Description
FORMS-2295. When there was an error loading a form (for example, invalid form id), the translation key was being returned in the error notification, not the resolved translation value.

permutils

Fix the translation resolution in permissionUtils.

Screenshot 2025-04-30 at 8 22 45 AM

permissionUtils does not contain a <script> block like Vue components, so importing and leveraging the i18n is different. Also, note that consoleError also acts differently than the base notification text. consoleErrors with options require the key but the base text requires the resolved translation.

## Type of Change

fix (a bug fix)

## Checklist

### DIFF
--- a/app/frontend/src/utils/permissionUtils.js
+++ b/app/frontend/src/utils/permissionUtils.js
@@ -9,6 +9,11 @@ import {
   NotificationTypes,
 } from '~/utils/constants';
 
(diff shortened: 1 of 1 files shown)


abhilash-aot (approved):
Translation update looks good

PR #1678, opened 2025-05-20
feat: FORMS-2335 backend implementation for "Assigned to Me" filter with tests

revanth-banala:
# Description
Backend Changes

Added filterAssignedToUserId filter methods to both submissionData and submissionMetadata model classes
Modified service.listFormSubmissions to include the new assigneeUserId column in results
Updated _initFormSubmissionsListQuery to support the new filter parameter
Enhanced the controller to extract the filter parameter and pass the current user object

Frontend Implementation

Added an "Assigned to me" checkbox in the submissions table UI

## Type of Change

feat (a new feature)

## Checklist


usingtechnology (changes requested):
This is looking good, nice work.
We do need a couple of enhancements - return columns we expect to be used in the frontend and have the unit tests test the new logic.

PR #1700, opened 2025-06-12
fix: FORMS-1451 Restrict form schedule late submission period to integers only

revanth-banala:
# Description
Form Submission Schedule allow decimal values for some fields
Description

The form submission schedule has a couple of places where it takes positive integers as values. However, the validation allows values with up to two decimal points.
For some reason if you enter three decimal places then the validation fails.
Expected behaviour: can only enter integer values greater than or equal to 1, and . cannot be entered.

Updated validation regex
Limits range to 1-999 previously it is 6 digits
Updated error message
Added translations
Decimal points can no longer be typed in the field

## Type of Change

fix (a bug fix)

## Checklist


app/frontend/src/components/designer/settings/FormScheduleSettings.vue
```diff
@@ -62,12 +62,25 @@ const scheduleCloseDate = ref([
...
   (v) =>
-    (v && new RegExp(/^[1-9]\d{0,5}(?:\.\d{1,2})?$/g).test(v)) ||
+    (v && new RegExp(/^[1-9]\d{0,2}$/g).test(v)) ||
     t('trans.formSettings.valueMustBeNumber'),
 ]);
 const scheduleTypedRules = ref([
   (v) => !!v || t('trans.formSettings.selectAnOptions'),
 ]);
+
+const handleIntegerInput = (event) => {
```
usingtechnology, line 72: Did you try using https://vuetifyjs.com/en/components/number-inputs/?
I think that could do what you need?

revanth-banala: I investigated this approach and found there are compatibility concerns v-number-input is currently an experimental component in Vuetify Labs so keeping the current implementation

abhilash-aot (approved):

PR #1711, opened 2025-06-24
FORMS-2435 - allow form to use multiple IDPs

usingtechnology:
# Description
Allow form owners/designers to specify multiple IDPs for Login protected forms.
The main requested use case is allowing BCeID Basic and Business users for the same form.

We have made no requirements, any set of IDPs is valid and up to the owner/designer to determine (ie IDIR + BC Services Card). 

This required a change to the Form Settings UX, no longer a set of cascading Radio Buttons. Now it is a dropdown/select for the 3 types: Public, Login, Team and upon selection, documentation and/or checkboxes (Login) are displayed.

Not only is the Form Access Setting component updated, but changes are made to the Login component as we now have a new use case where a form can be logged in with multiple IDPs. If a form has multiple IDPs and the user is not signed in with an approved IDP, then they will go to a Login selection screen that will list only the valid IDPs (not all 4 like the default Login screen).

Updates made in the i18n files to reflect changes in the functionality. And updates made in test classes on affected code.

## Type of Change

feat (a new feature)

## Checklist


jasonchung1871 (approved):
Code looks good, and it works in the deployment. I can't think off the top of my head of any of the ramifications for this. Are there any components that require specific IDP? Any CHEFS or external functionality that requires specific IDP? As far as I know, it seems to work.

PR #1715, opened 2025-07-02
fix: FORMS-2520 allow authenticated users to download files from public submissions

revanth-banala:
# Description
When a file is uploaded through a public form, the submitter cannot download the file post-submission, which is expected behavior. However, users with reviewer or higher roles also cannot download the file from the submission view, which is not expected.

Expected:

Submitters of public forms should not be able to download uploaded files after submission. However, reviewers and higher roles should be able to download these files from the submission view.

Actual:

Neither public users nor reviewers (or higher roles) can download uploaded files post-submission.

The previous fix for FORMS-2470 successfully disabled file downloads for public users after form submission, but it also unintentionally blocked authenticated reviewers and form managers from downloading files when viewing public form submissions.
Modified the shouldDisableFileDownloads computed property in FormViewer.vue to check user authentication status
This only blocks unauthenticated users on public forms

## Type of Change

fix (a bug fix)

## Checklist

### DIFF
--- a/app/frontend/src/components/designer/FormViewer.vue
+++ b/app/frontend/src/components/designer/FormViewer.vue
@@ -136,7 +136,9 @@ const shouldDisableFileDownloads = computed(() => {
     return false;
   }
 
-  return properties.readOnly && isFormPublic(form.value);
+  return (
+    properties.readOnly && isFormPublic(form.value) && !authenticated.value
+  );
 });
 
 const viewerOptions = computed(() => {


usingtechnology:
I'm not 100% sure that this wasn't an issue before but we should investigate and see if the fix is small...
I have created a public form, there are no members on my team, but if i log in as another IDIR user (not the form owner) I can see the submission and download the attachment. Some errors are thrown, but the submission is returned and download is accessible.

https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-1715/form/view?s=a5b64f73-0b13-4e58-a800-3c861215f29b

PR #1718, opened 2025-07-03
FORMS-2525: allow all but executable type files to be uploaded by def…

usingtechnology:
…ault

# Description
See FORMS-2525.
TLDR - Users lost the ability to upload various files (ex., *.msg) by default. This PR removes the default allowed values but retains restricted values (ie. executable files and files that can contain runnable scripts).

I update the file service tests to get greater coverage not just tests for this change.

## Type of Change

fix (a bug fix)

## Checklist

### DIFF
--- a/app/src/forms/file/service.js
+++ b/app/src/forms/file/service.js
@@ -33,36 +33,6 @@ const BLOCKED_EXTENSIONS = [
   '.aspx',
 ];
 
-const SECURE_EXTENSIONS = [
-  '.pdf',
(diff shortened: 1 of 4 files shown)


revanth-banala:
I have tested it out and now we can upload .msg file type too and still blocking the executables 
Looks good!

revanth-banala (approved):
I have tested it out and now we can upload .msg file type too and still blocking the executables
Looks good!

PR #1720, opened 2025-07-09
FORMS-2521: Fix form settings for public

usingtechnology:
# Description

If public form access is selected, "Allow multiple draft upload" is selectable. This option is blocked elsewhere in the code, so it is confusing if it is a selectable option on form settings.

Update UX to set false if access changed to Public (like with Submitter can save draft).
Add protections in the backend on create/update to ensure it can only be true if set to true AND is a protected form.

## Type of Change

fix (a bug fix)

## Checklist


usingtechnology:
> @usingtechnology "Allow multiple draft upload" checkbox deselects automatically after updating the form settings .So multiple draft button missing on submission page as well. Can you check that?

Good catch! The check was looking at the form identity providers and setting false if there was public. but when team, there are no identity providers to check! will address that...Thanks!

nimya-aot (approved):

PR #1723, opened 2025-07-10
fix:FORMS-2522 add permission check to prevent unauthorized access to submission review

revanth-banala:
# Description
Security Gap: Submission Reviewer screen

Description

If someone has a link to submission review and they are not part of the review team, the submission data is still retrieved and visible (for public submissions).
Prevent users who do not have SUBMISSION_REVIEW permissions on a submission from navigating to the Submission Review page.
A link is in the format of: 
[Common Hosted Forms Service](https://chefs-dev.apps.silver.devops.gov.bc.ca/pr-1715/form/view?s=a5b64f73-0b13-4e58-a800-3c861215f29b) 
This will load the frontend/src/views/form/View.vue that loads frontend/src/components/forms/FormSubmission.vue component. It is this component that successfully calls the read submission api (/api/v1/submissions/a5b64f73-0b13-4e58-a800-3c861215f29b).
Other related components (Status and Notes) used in the FormSubmission.vue will fail as they call api endpoints that look for SUBMISSION_REVIEW permission.

We should block navigation at the router level, if the current user doesn’t have SUBMISSION_REVIEW permissions then they should not load frontend/src/views/form/View.vue (FormSubmission.vue).

Update router.js to use a beforeEnter to call status or notes API and if it fails we know they don’t have the proper permissions, send them to NotFound, if they have permissions, carry on…

AC: Current User requires SUBMISSION_REVIEW permission to navigate to FormSubmission.

## Type of Change

fix (a bug fix)

## Checklist

## Further comments
- Users with only FORM_SUBMITTER role could view any submission if they had the direct link
- The submission data was loaded before permission checks occurred
- Only status and notes API calls were protected, but main submission data was exposed

### Solution
- Added `beforeEnter` route guard to check permissions before loading components
- Uses `formService.getSubmissionStatuses` to verify SUBMISSION_REVIEW permission
- Redirects unauthorized users to NotFound page
- Prevents View.vue and FormSubmission.vue from loading without proper permissions

### DIFF
--- a/app/frontend/src/router.js
+++ b/app/frontend/src/router.js
@@ -5,7 +5,7 @@ import { useAuthStore } from '~/store/auth';
 import { useFormStore } from '~/store/form';
 import { useIdpStore } from '~/store/identityProviders';
 import { preFlightAuth } from '~/utils/permissionUtils';
-
+import { formService } from '~/services';
 let isFirstTransition = true;
 let router = undefined;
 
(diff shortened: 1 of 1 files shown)


usingtechnology (approved):
Nice work!

PR #1724, opened 2025-07-10
fix: FORMS-1036 Clear reminder flag when form schedule is removed or disabled

revanth-banala:
# Description
Reminder enabled flag is not being cleared when schedule removed

Add
Description

Background: a form can have a submission schedule where it automatically opens, closes, or both. One of the features is that email reminders about these dates can be sent out to submitters.

There’s a bug where the form can have “Forms Submissions Schedule“ enabled, and a schedule set up with email reminders enabled. If the “Forms Submissions Schedule” is then disabled, the schedule is cleared but the email reminders remain enabled. This has uncovered a bug (FORMS-1037) in the backend that it is not handling this inconsistent database state.

Fix the backend in 1037 to handle this inconsistent state

Figure out how to prevent this inconsistent state. One way would be to change the frontend, but should the backend also enforce it (400 error) or silently fix it?

Clean up the databases where this inconsistent state happens.

## Type of Change

fix (a bug fix)

## Checklist

## Further comments
Solution
Implemented a two-layer fix:

1. **Frontend**: Added watchers to automatically clear the reminder flag when:
   - Schedule type changes to MANUAL or null
   - Schedule is disabled (enabled = false)

2. **Backend**: Added validation to ensure data integrity:
   - Validates reminder settings against schedule configuration
   - Silently corrects any inconsistent state
   - Prevents reminders from being enabled without a valid schedule

### DIFF
--- a/app/src/forms/form/service.js
+++ b/app/src/forms/form/service.js
@@ -120,6 +120,40 @@ function isClosingMessageValid(schedule) {
 }
 
 const service = {
+  /**
+   * Validates reminder settings against schedule configuration
+   * Ensures reminders are only enabled when a valid schedule exists
+   * @param {Object} data Form data containing schedule and reminder_enabled
+   * @returns {boolean} Validated reminder_enabled value
+   */
+  _validateReminderSettings: (data) => {
+    // If reminders are not enabled, no validation needed
+    if (!data.reminder_enabled) {
+      return false;
+    }
+
+    // Check if schedule exists and is properly configured
+    const hasValidSchedule =
+      data.schedule &&
+      data.schedule.enabled === true &&
+      data.schedule.scheduleType !== null &&
+      data.schedule.scheduleType !== undefined &&
+      data.schedule.scheduleType !== ScheduleType.MANUAL;
+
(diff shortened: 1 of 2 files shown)


app/src/forms/form/service.js
```diff
@@ -120,6 +120,40 @@ function isClosingMessageValid(schedule) {
...
+      data.schedule.scheduleType !== null &&
+      data.schedule.scheduleType !== undefined &&
+      data.schedule.scheduleType !== ScheduleType.MANUAL;
+
+    if (!hasValidSchedule) {
+      return false;
+    }
+
+    // Additional check: ensure open date exists and is in the future
+    if (data.schedule.openSubmissionDateTime) {
```
usingtechnology, line 148: We should be leveraging logic in scheduleService. You spent a lot of time working with schedules and dates to make sure the timezones were respected and there are functions for doing all the date checks. Use that functionality.

Also, add tests for this function in the unit test class.

usingtechnology (changes requested):
Unit tests and leverage existing code for schedule date checks.

PR #1725, opened 2025-07-11
FORMS-2537: move submission notification outside transaction

usingtechnology:
# Description
We had a case where a user submitted their draft, the browser appeared to spin and never return. The staff got a Submission Received email notification with a link to that submission. However, that transaction failed (error not returned to browser? that is unclear), leaving the submission in draft state. Which meant that when staff clicked the link to view the submission they got an error that the submission could not be loaded. The submission was also not in their table of submissions to review (as it was still in draft).

It appears that the transaction was rolled back, but we had the notification logic inside the transaction; can't rollback the email notification.

So this PR is to refactor the update logic (it is gnarly) and move notifications outside the transaction. If the update is successful, then we want to send out notifcations (event stream, event webhooks, emails).

The logic for the update method was intense, I refactored it to its own file for clarity and testing.

## Type of Change

fix (a bug fix)

## Checklist

### DIFF
--- a/app/src/forms/submission/service.js
+++ b/app/src/forms/submission/service.js
@@ -1,14 +1,12 @@
 const uuid = require('uuid');
-const log = require('../../components/log')(module.filename);
 
 const { Statuses } = require('../common/constants');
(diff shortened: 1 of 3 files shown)


jasonchung1871 (approved):
Looks good, way clearer to read. Not sure if there any edge cases specifically with updating submissions but it otherwise works as it used to.

PR #1729, opened 2025-07-16
FORMS-2540: Fix Email send error on RequestReceipt component

usingtechnology:
# Description
RequestReceipt formId value was not valid. Fix how it gets populated (the stored form).
Remove unused parameters on Success component and its router.

When a form has enabled Only Share Draft with Team, there is a check to see if the current user is a member of the team. The call to check was failing because the value provided for the form Id was never set. There were a few parts of the way it should have been set that were completely incorrect too. However, the reality is we didn't need to fix those two bugs (query string not getting set, incorrect property name) because we already have the form.id. Remove the broken an unnecessary query param and property. 

## Type of Change

fix (a bug fix)

## Checklist

### DIFF
--- a/tests/functional/cypress/e2e/form-submission-public-no-status-assign.cy.js
+++ b/tests/functional/cypress/e2e/form-submission-public-no-status-assign.cy.js
@@ -22,14 +22,9 @@ describe('Form Designer', () => {
     });
   });
   it('Visits the form settings page', () => {
-    
-    
     cy.viewport(1000, 1100);
     cy.waitForLoad();
-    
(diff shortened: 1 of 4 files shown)


nimya-aot:
It is working good

usingtechnology:
> It is working good

Can you approve it?

nimya-aot:
> > It is working good
> 
> Can you approve it?

I dont have permission to approve. Initially it was. May be somebody changed permissions on GitHub.

nimya-aot (approved):

PR #1733, opened 2025-07-21
FORMS-2589: Admin panel fixes

RyanBirtch-aot:
The Admin Panel has strange behaviour surrounding search, namely the search results bounce around as the various requests come in and update the table. I implemented debouncing on the search input so there are not seperate requests for each keypress in the search bar. It is these requests arriving out of order which would cause the results table to jump around.  Also the filter for only viewing deleted forms is fixed. 

# Description

## Checklist

### DIFF
--- a/app/frontend/src/components/admin/AdminAPIsTable.vue
+++ b/app/frontend/src/components/admin/AdminAPIsTable.vue
@@ -31,7 +31,7 @@ const editDialog = ref({
 const firstDataLoad = ref(true);
 const forceTableRefresh = ref(0);
 const debounceInput = ref(null);
-const debounceTime = ref(300);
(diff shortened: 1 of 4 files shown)


RyanBirtch-aot:
We could set it to that, but I'd worry about people trying to search for rather short words or names? I could see extending the timeout for the search so that it triggers after a second or so after typing commences?

usingtechnology:
> We could set it to that, but I'd worry about people trying to search for rather short words or names? I could see extending the timeout for the search so that it triggers after a second or so after typing commences?

Yeah, a slight time lag would work too... it just seems too jumpy as you type.

usingtechnology (approved):
i like it better, time to find out if the users care.

PR #1737, opened 2025-07-24
Devcontainer Performance Improvement

usingtechnology:
Rebuilding and opening the devcontainer is slow, optimizing the Dockerfile for layering, adding checks to post-install, making cypress optional, adding tasks for cypress.

When changing branches and after a merge to dev, I Rebuild the Container and it takes far too long. This is to speed it up and give me back more time for development work.

The biggest change is moving cypress to an optional part of the install, once the container is built and open: Run Tasks -> Install Cypress or Install Cypress (Interactive).

The .devconainer/.dockerignore allows for a smaller image (faster to build, smaller layers).
Added a JRE to the image for SonarQube, so that should load faster after rebuilds and removes the prompt asking if it should download a JRE.

# Description

## Type of Change

## Checklist

### DIFF
--- a/.devcontainer/install-cypress.sh
+++ b/.devcontainer/install-cypress.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+# Interactive Cypress installation script
+echo "🤔 Checking for Cypress setup..."
+
+if [ -f "package.json" ]; then
+  echo "📦 Found package.json in $(pwd)"
+  read -p "Install Cypress? (y/n): " -n 1 -r
+  echo ""
+  
+  if [[ $REPLY =~ ^[Yy]$ ]]; then
+    echo "📦 Installing Cypress..."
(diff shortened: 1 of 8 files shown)


.devcontainer/install-cypress.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+# Interactive Cypress installation script
+echo "🤔 Checking for Cypress setup..."
+
+if [ -f "package.json" ]; then
```
nimya-aot, line 6: @usingtechnology  I think we need to specify package.json file for Cypress. The path would be tests/functional/cypress/package.json

usingtechnology: @nimya-aot - made the shell script work from any directory.

nimya-aot (approved):

PR #1748, opened 2025-08-06
Map Required Setting

RyanBirtch-aot:
# Description

I tested the map required bug in production and noticed a bug when I used one of my pre-existing forms. This comes up when the map does not allow for the base layer to be changed.

## Checklist


nimya-aot:
@RyanBirtch-aot Can you please deploy this PR to PR environment?

nimya-aot:
@RyanBirtch-aot So If this checkbox is not checked, then its read only map and we need to block the "Required" option? is that is the fix?
image
image

nimya-aot:
> We chose to just ignore the required flag in case of a read-only component because of some issues with form settings during form creation and also to prevent any issues if a form designer forgot to turn off the required flag. I'm not 100% sure if we can turn off settings in other form tabs based on another tab with how the form settings are rendered with formio? The bug came about because of the base map being added to the saved value of the map component. Previously the check was for only an empty value, but once the base map was added, we needed to change that check.

Problem is Users can still submit the form without placing a point on the map even though its a required one if they accidently uncheck the checkbox to make it a read only map
image

PR #1785, opened 2025-11-20
fix: Make file move and DB update atomic for submissions

jasonchung1871:
# Description

Refactored moveSubmissionFile to use a transaction, ensuring the file's DB update and move are atomic. The original file is deleted only after a successful commit, and errors during deletion are logged but do not fail the operation. Updated related service and tests to reflect new error handling and atomicity.

https://citz-do.atlassian.net/browse/CCP-2917

## Checklist


jasonchung1871:
This PR https://github.com/bcgov/common-hosted-form-service/pull/1791 was recently pushed through which allows for files that are not in the correct location to be checked against the fallback location. This doesn't resolve the issue yet as we don't know what's causing them to be placed in the wrong location. This is still a necessary PR towards determining the cause of the issue.

nimya-aot:
@jasonchung1871 The file download fails after submission. Could you please check?
image

nimya-aot:
@jasonchung1871 This PR fails while downloading the submitted file
image

RyanBirtch-aot (approved):

PR #1833, opened 2026-03-11
Endpoint Limit Bug

RyanBirtch-aot:
# Description
A client was unable to save an event subscription callback URL (endpoint URL) in CHEFS. After debugging, the root cause appears to be the database column length limit for the subscription URL in the form subscription table. The url column is currently limited to 255 characters. The client’s callback URL includes many query parameters and exceeds 255 characters, so the UI/API rejects or fails to save the URL. I have added migration to increase the length of these URLs to 512 characters.

## Type of Change

fix (a bug fix) 

## Checklist


jasonchung1871 (approved):

usingtechnology (approved):
LGTM!

PR #1857, opened 2026-04-29
Revert "fix: Make file move and DB update atomic for submissions"

RyanBirtch-aot:
Reverts bcgov/common-hosted-form-service#1785


nimya-aot (approved):
Reverting needed as there are failures on push

PR #1906, opened 2026-06-18
fix: file access for api, transaction race conditions

usingtechnology:
fix: restore file API access and fix file storage transaction races

# Description

A user reported they could no longer download files using a Basic Auth API key. This turned out to be fallout from the earlier `requireApiKeyBasic` gateway work: the file routes had been switched to the system `checkApiKey` middleware, which is meant to guard internal cron-only endpoints, not file access.

While tracing that, I found two file-storage bugs in the submission flow that were (potentially) causing uploads to hang and deletes to leave files behind.

This PR has three parts:

1. **File API access.** The file download, clone, and delete routes now use the Basic Auth middleware (`basicApiAccess`) instead of the system `checkApiKey`. This restores the three access paths these routes are supposed to support: public users on public-form drafts, API users with Basic Auth on protected forms, and signed-in users with Bearer tokens. Comment added on the routes describing what each piece of the chain allows and blocks, so the next person doesn't have to reverse-engineer it.

2. **Hardened `checkApiKey`.** Removed the `Authorization` header bypass so the middleware only honours the `apikey` header. It now fails closed when `APITOKEN` isn't configured and compares keys with a constant-time comparison. This is the only thing guarding the reminder and records-management deletion cron routes, so the bypass was an auth hole. The OpenShift cron jobs send the `apikey` header, so they keep working.

3. **File transaction race conditions.**
   - Uploading a submission with new files could deadlock. `moveSubmissionFile` opened its own transaction to update a `FileStorage` row that the submission's outer transaction had already locked, so the two waited on each other. It now runs inside the caller's transaction and deletes the original object after that transaction commits.
   - Deleting a submission called `deleteFiles` without awaiting it, in a separate transaction that deleted the same rows the outer transaction was deleting. That was a fire-and-forget call that could orphan files in storage. The deletion now happens in one transaction and the stored objects are purged (best effort) after commit.

## Type of Change

fix (a bug fix)

## Checklist

## Further comments

The storage cleanup after a move or delete is intentionally best effort: it runs after the database transaction commits and a storage failure is logged rather than thrown. The alternative is failing a submission or delete that has already succeeded in the database, which is worse than leaving an object to be cleaned up later. The bulk file-delete route still uses `deleteFiles` as before, since it manages its own transaction correctly.

Tests added/updated cover the hardened `checkApiKey` (including the bypass regression and the unset-`APITOKEN` case), the file routes' middleware chain, `moveSubmissionFile` reusing a passed-in transaction, the new `deleteStorageObject` helper, and the submission-delete path (commit-then-purge, no fire-and-forget, rollback leaves storage untouched).

### DIFF
--- a/.devcontainer/devcontainer-lock.json
+++ b/.devcontainer/devcontainer-lock.json
@@ -1,0 +1,14 @@
+{
+  "features": {
+    "ghcr.io/devcontainers/features/docker-in-docker:2": {
+      "version": "2.17.0",
+      "resolved": "ghcr.io/devcontainers/features/docker-in-docker@sha256:25b9f05705ffba7dbe503230ac76081419306f8c8bc88e0ce78c4ecd99a0c78c",
+      "integrity": "sha256:25b9f05705ffba7dbe503230ac76081419306f8c8bc88e0ce78c4ecd99a0c78c"
+    },
+    "ghcr.io/devcontainers/features/git:1": {
+      "version": "1.3.5",
+      "resolved": "ghcr.io/devcontainers/features/git@sha256:27905dc196c01f77d6ba8709cb82eeaf330b3b108772e2f02d1cd0d826de1251",
+      "integrity": "sha256:27905dc196c01f77d6ba8709cb82eeaf330b3b108772e2f02d1cd0d826de1251"
+    }
+  }
+}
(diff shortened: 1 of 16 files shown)


jasonchung1871 (approved):
It's working for me!

PR #1913, opened 2026-06-24
fix: CCP-4948 Submission Fields access

usingtechnology:
fix: CCP-4948 show submission columns for Reviewers

# Description

The "Manage columns" picker on the Submissions page only showed the form's field columns to Owners and Form Designers. A user with just the Reviewer role saw none of them — so searching for a field like `simpletextfield` returned nothing, even though Reviewers can see the submissions themselves.

The cause: the column picker pulled the form field list off `form.versions`, but `readForm` intentionally strips `versions` from anyone without design access (a data-integrity decision we want to keep). Reviewers never got a version id, so the field list came back empty.

Rather than loosen `readForm`, this adds a small dedicated endpoint, `GET /forms/:formId/fields`, gated on `FORM_READ` + `SUBMISSION_READ` — the same permissions already required to view the submissions table. It returns the schema-free version list plus the field keys for the current version (the published one, falling back to the latest if none is published). No schema or draft data is exposed.

The same `form.versions` dependency existed in two other places reachable by non-designers, so they're fixed in the same pass:
- the submissions column picker (`SubmissionsTable.vue`) — the reported bug
- "My Submissions" column picker (`MySubmissionsTable.vue`)
- the Export Submissions version selector (`ExportSubmissions.vue`)

All three now read from the new endpoint instead of `form.versions`.

## Type of Change

fix (a bug fix)

## Checklist

### DIFF
--- a/app/frontend/src/components/forms/ExportSubmissions.vue
+++ b/app/frontend/src/components/forms/ExportSubmissions.vue
@@ -37,6 +37,9 @@ const endDateRules = ref([
 const exportFormat = ref('json');
 const formFieldsSearchFilter = ref('');
 const formVersions = ref([]);
+// Schema-free version metadata ({ id, version, published }) from the form fields
+// endpoint, so this works for reviewers who can't read versions via the form.
+const formVersionList = ref([]);
 // This has the form fields in the table
 // Table selection does not work if we use a computed value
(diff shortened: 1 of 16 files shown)


nimya-aot (approved):
working good

PR #1919, opened 2026-06-26
feat: CCP-3947 cdogs v3 carbone enterprise

usingtechnology:
# Description

Adds support for a second  CDOGS (v3 / Carbone) instance alongside the existing CDOGS v2 service, and routes all document generation through a single `docGenService` facade that decides which generator a form should use.

Routing is driven by the existing feature flags: `documentGenerationV2` (universal) and `documentGenerationV3` (allowlisted per form/tenant). The facade resolves V3 first and falls back to V2; if neither is active for a form it fails with a 403. Every render attempt — success, failure, or denial — is recorded in a new `form_document_generation` metrics table (form, generator, outcome, duration, user). CHEFS Users with Admin role will have to add the form to the V3 Allowlist.

A few supporting changes came along with this:
- `AxiosService` now supports a credential-less mode, since the v3 service currently sits behind a private URL with no token/credentials (auth can be added later without code changes).
- The draft/preview render endpoint moved off `/utils/template/render` (auth was just `currentUser`) to `POST /forms/:formId/template/render`, guarded by `hasFormPermissions`. The `/utils` module is retired. The frontend draft call moved from `utilsService` to `formService` accordingly.
- `errorToProblem` now sanitizes what it logs — CDOGS errors were dumping huge binary request/response bodies into the logs.

Existing forms are unaffected: they continue to use CDOGS v2 transparently. V3 stays off until the flag is enabled and a form is allowlisted.

## Checklist

## Further comments

The facade lives in `components/docGenService.js` and is the only thing that talks to the two CDOGS clients, so gating and metrics stay in one place. We deliberately kept the existing v1 endpoints and route internally rather than exposing a parallel `/api/v2` — external API consumers upgrade to v3 automatically once their form is allowlisted, with no client change.

Rollout is config-driven: set `features.documentGenerationV3.enabled` and allowlist forms via the Feature Flags admin page. The v3 endpoint is configured per environment via `SC_CS_CDOGS_V3_ENDPOINT` (currently in`config/default.json`).


app/src/components/docGenService.js
```diff
@@ -0,0 +1,156 @@
...
+  }
+};
+
+/**
+ * Derive a meaningful, storable error detail string from a thrown error.
+ * Prefers `error.detail`: a non-empty, non-`[object Object]` string is used as-is;
+ * an object is JSON-stringified (so structured CDOGS/Problem details survive);
+ * otherwise falls back to `error.message`. Returns null when nothing useful exists.
+ */
+const normalizeErrorDetail = (error) => {
```
jasonchung1871, line 38: not an issue at all, but it would be nice if we could reuse this in other error handlers at least until chefs 1.0 gets sunset, e.g., CHES, CSS, other external services

usingtechnology: ahhh, this might be left behind? later I updated errorToProblem to do this in a more global way....

jasonchung1871 (approved):

PR #1920, opened 2026-06-28
Feature/ccp 5048 fix email submission link

bhuvan-aot:
# Description
ccp-5048
When an unauthenticated user clicks the "View this submission" link from a submission confirmation email, the app navigates to /form/view?s=<submissionId>.

What was broken:
The FormView route's beforeEnter guard calls getSubmissionStatuses() to check permissions. For unauthenticated users, this returns a 401. The guard was treating 401 the same as 404 — redirecting to the NotFound page. So users saw a 404 instead of being asked to log in.

The fix (router.js):
Split the 401 and 404 handling:
- 401 → call next() with no redirect. Since FormView has requiresAuth: true, the global guard calls authStore.login(), which redirects the user to the Keycloak login page. After login, they land back on the submission view.
- 404 → still redirects to NotFound (submission genuinely doesn't exist).

## Checklist


usingtechnology (approved):
LGTM

PR #1924, opened 2026-07-02
fix(ui): Review/fix propogate advanced checkbox

usingtechnology:
See #1922


RyanBirtch-aot (approved):
Works well! Good work

